PR #8, opened 2026-02-10
feat: Kademlia DHT lookup for closest-peer chunk routing

mickvandijke:
## Summary
- Replace naive first-connected-peer selection with iterative Kademlia network queries (`find_closest_nodes`) so chunks are routed to the peer whose BLAKE3-derived DHT key is closest to the chunk address by XOR distance
- Exclude the local node from results by comparing against its own BLAKE3 DHT key hex

**This is still a WIP** — further testing and refinement needed before merging.

## Test plan
- [ ] Verify chunk routing targets the closest peer by XOR distance
- [ ] Confirm local node is excluded from routing candidates
- [ ] E2E test with multi-node setup
- [ ] Run full test suite (`cargo test`)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ path = "src/bin/saorsa-client/main.rs"
 
 [dependencies]
 # Core (provides EVERYTHING: networking, DHT, security, trust, storage)
-saorsa-core = "0.10.4"
+saorsa-core = "0.11.0"
 saorsa-pqc = "0.4.0"
 
 # Payment verification - autonomi network lookup + EVM payment
--- a/src/client/data_types.rs
+++ b/src/client/data_types.rs
@@ -18,6 +18,32 @@ pub fn compute_address(content: &[u8]) -> XorName {
     address
 }
 
+/// Compute the XOR distance between two 32-byte addresses.
+///
+/// Lexicographic comparison of the result gives correct Kademlia distance ordering.
+#[must_use]
+pub fn xor_distance(a: &XorName, b: &XorName) -> XorName {
+    let mut result = [0u8; 32];
+    for i in 0..32 {
+        result[i] = a[i] ^ b[i];
+    }
+    result
+}
+
+/// Convert a hex-encoded peer ID string to an `XorName`.
+///
+/// Returns `None` if the string is not valid hex or is not exactly 32 bytes (64 hex chars).
+#[must_use]
+pub fn peer_id_to_xor_name(peer_id: &str) -> Option<XorName> {
+    let bytes = hex::decode(peer_id).ok()?;
+    if bytes.len() != 32 {
+        return None;
+    }
+    let mut name = [0u8; 32];
+    name.copy_from_slice(&bytes);
+    Some(name)
+}
+
 /// A content-addressed identifier (32 bytes).
 ///
 /// The address is computed as SHA256(content) for chunks,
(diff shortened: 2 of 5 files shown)


dirvine (changes requested):
## Review Summary

Thanks @mickvandijke for this contribution! The approach of using Kademlia DHT lookup for closest-peer chunk routing is exactly right. However, there are a few issues that need to be addressed before merging:

### 🔴 Blocking: Compilation Error

The CI is failing because `node.dht()` returns `Option<&Arc<DhtNetworkManager>>`, not `&Arc<DhtNetworkManager>` directly. You need to handle the `Option`:

```rust
// Current (broken):
let closest_nodes = node
    .dht()
    .find_closest_nodes(target, CLOSE_GROUP_SIZE)
    .await
    .map_err(...)?;

// Fixed:
let dht = node.dht().ok_or_else(|| {
    Error::Network("DHT not available".into())
})?;
let closest_nodes = dht
    .find_closest_nodes(target, CLOSE_GROUP_SIZE)
    .await
    .map_err(|e| Error::Network(format!("Kademlia closest-nodes lookup failed: {e}")))?;
```

### 🟢 Good News: XOR Sorting Concern is Invalid

Both Copilot and Greptile raised concerns about not explicitly sorting by XOR distance. I verified that `find_closest_nodes` in saorsa-core **does return results already sorted by XOR distance** (see `src/dht/core_engine.rs:224`: `candidates.sort_by(|a, b| a.1.cmp(&b.1))`). So your `.find()` approach is correct - the first non-self peer IS the closest.

### 📝 Suggestions (non-blocking)

1. **Consider adding a unit test** with mocked DHT results to verify the peer selection logic
2. **Consider making `CLOSE_GROUP_SIZE` configurable** via `QuantumConfig` for deployment flexibility

Once the compilation error is fixed, I'll be happy to approve this. Great work! 🎉

dirvine (commented):
## Update: Simpler Fix Available! 🎉

I just realized the root cause: the PR uses `saorsa-core = "0.10.4"` but **v0.11.0** (just published) changed the `dht()` API from returning `Option` to returning `&Arc<DhtNetworkManager>` directly.

### The Fix

Simply update `Cargo.toml`:

```diff
-saorsa-core = "0.10.4"
+saorsa-core = "0.11.0"
```

This should make the code compile as-is without any changes to the `pick_target_peer` logic. The current implementation is correct for v0.11.0! 

v0.11.0 was just published from PR #25 (feat/dht-transport-refactor) which cleaned up the DHT API.

cc @davidirvine

dirvine (approved):
## ✅ Approved!

I've pushed a commit that bumps `saorsa-core` to **v0.11.0**, which fixes the compilation error. The new version has a non-optional `dht()` API that returns `&Arc<DhtNetworkManager>` directly.

### Changes Made
- Updated `Cargo.toml`: `saorsa-core = "0.10.4"` → `"0.11.0"`

### Verification
- ✅ `cargo check` passes
- ✅ `cargo clippy -- -D warnings` passes  
- ✅ All 138 unit tests pass

### Summary

This is a great contribution @mickvandijke! 🎉

- **Correct approach**: Using Kademlia DHT lookup for closest-peer chunk routing
- **Correct implementation**: `find_closest_nodes` in saorsa-core already returns results sorted by XOR distance, so `.find()` correctly selects the closest non-self peer
- **Good tests**: The `xor_distance` and `peer_id_to_xor_name` helpers have solid unit test coverage

Thanks for contributing to saorsa-node! 🙌

dirvine:
Thanks so much for this contribution @mickvandijke! 🙌

The Kademlia-based closest-peer selection is exactly the right approach for chunk routing. Great work on:

- Adding the XOR distance utilities with comprehensive tests
- Properly filtering out the local node from routing candidates
- Moving address computation before peer selection to enable routing by content address

Looking forward to more contributions from you! 🚀

PR #9, opened 2026-02-12
Replace shutdown channels with CancellationToken

mickvandijke:
## Summary
- Replace `tokio::sync::broadcast` and `tokio::sync::watch` shutdown channels with `tokio_util::CancellationToken` in `Devnet` and `RunningNode`
- Shut down devnet nodes concurrently via `join_all` instead of sequentially
- Simplify shutdown select branches — `CancellationToken::cancelled()` needs no extra borrow-check, removing nested `if` guards

## Test plan
- [ ] `cargo build --release` compiles cleanly
- [ ] `cargo test` passes all existing tests
- [ ] Devnet startup/shutdown cycle completes without hanging

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,6 +51,7 @@ sha2 = "0.10"
 
 # Async runtime
 tokio = { version = "1.35", features = ["full", "signal"] }
+tokio-util = { version = "0.7", features = ["rt"] }
 futures = "0.3"
 
 # CLI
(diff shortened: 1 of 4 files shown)


dirvine (approved):
Clean refactor: broadcast/watch shutdown channels → CancellationToken. Concurrent shutdown via join_all is a solid improvement. All CI gates pass. LGTM.

PR #11, opened 2026-02-19
feat: LMDB storage + persistent node identity

mickvandijke:
## Summary

- Replace file-per-chunk `DiskStorage` with `LmdbStorage` backed by LMDB (heed 0.22), providing ACID transactions, zero-copy mmap reads, and O(1) key lookups
- Same public API shape (`put`/`get`/`exists`/`delete`/`stats`); `exists()` now returns `Result<bool>` and `current_chunks()` returns `Result<u64>` to propagate LMDB errors instead of silently swallowing them
- LMDB environment at `{root_dir}/chunks.mdb/` with 32 GiB default map size (configurable via `storage.db_size_gb` in `config.toml`)
- Persist ML-DSA-65 node identity keypair across restarts, giving each node a stable `peer_id`
- Scope each node's data directory by peer ID hex prefix (`{base_dir}/{node_id_prefix}/`), preventing LMDB collisions when running multiple nodes on one machine
- Auto-scan base directory for existing identities on startup; error if multiple found without an explicit root dir. Custom `root_dir` from `config.toml` is respected (not just `--root-dir`)
- `unsafe_code` lint relaxed from `forbid` to `deny` with a single scoped `#[allow(unsafe_code)]` on `LmdbStorage::new()` for LMDB's `open()` call
- Fix pre-existing `is_running().await` compile error in devnet/testnet health monitors

## Test plan

- [x] `cargo build --release` compiles cleanly
- [x] `cargo test` -- all tests pass (0 failures)
- [x] `cargo clippy --all-features -- -D clippy::panic -D clippy::unwrap_used -D clippy::expect_used` -- no warnings
- [x] `cargo fmt --all -- --check` -- clean
- [x] `test_persistence_across_reopen` verifies data survives LMDB close/reopen cycle
- [x] `test_identity_persisted_across_restarts` verifies identity survives node restart (implicit scan path)
- [x] `test_explicit_root_dir_persists_across_restarts` verifies identity survives restart with `--root-dir`
- [x] `test_multiple_identities_errors` verifies error when multiple identities found without explicit root
- [x] `test_chunk_persistence_across_node_restarts` verifies chunks survive full node restart (e2e)

### DIFF
--- a/src/node.rs
+++ b/src/node.rs
@@ -299,6 +299,7 @@ impl NodeBuilder {
             root_dir: config.root_dir.clone(),
             verify_on_read: config.storage.verify_on_read,
             max_chunks: config.storage.max_chunks,
+            max_map_size: config.storage.db_size_gb.saturating_mul(1_073_741_824),
         };
         let storage = LmdbStorage::new(storage_config)
             .await
--- a/src/node.rs
+++ b/src/node.rs
@@ -2,7 +2,8 @@
 
 use crate::ant_protocol::CHUNK_PROTOCOL_ID;
 use crate::config::{
-    default_root_dir, EvmNetworkConfig, IpVersion, NetworkMode, NodeConfig, NODE_IDENTITY_FILENAME,
+    default_nodes_dir, default_root_dir, EvmNetworkConfig, IpVersion, NetworkMode, NodeConfig,
+    NODE_IDENTITY_FILENAME,
 };
 use crate::error::{Error, Result};
 use crate::event::{create_event_channel, NodeEvent, NodeEventsChannel, NodeEventsSender};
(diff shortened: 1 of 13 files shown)


src/node.rs
```diff
@@ -169,6 +179,114 @@ impl NodeBuilder {
...
+    ///   3. **None found** — first run: generate identity, derive `{base_dir}/{hex_prefix}/`,
+    ///      save identity there, and update `config.root_dir`.
+    ///   4. **Exactly one found** — load it and update `config.root_dir`.
+    ///   5. **Multiple found** — return an error asking the user to specify `--root-dir`.
+    async fn resolve_identity(config: &mut NodeConfig) -> Result<NodeIdentity> {
+        if config.root_dir_explicit {
+            return Self::load_or_generate_identity(&config.root_dir).await;
+        }
+
+        let base_dir = default_root_dir();
```
dirvine: Blocking: this ignores `config.root_dir` whenever `root_dir_explicit == false` and always falls back to `default_root_dir()`. Because `root_dir_explicit` is only set for CLI/env `--root-dir`, a `root_dir` loaded from `config.toml` is silently ignored. That means existing deployments with a custom `root_dir` in config file will start reading/writing identity and storage under the platform default directory instead.\n\nCan we treat non-default/root-dir-from-config as explicit (or propagate an explicit-source flag from config loading) before this branch?

dirvine (changes requested):
Superseded by my next review; please ignore this malformed body.

dirvine (changes requested):
I found one blocking regression in the identity/root-dir resolution path: root_dir values loaded from config.toml are ignored unless --root-dir or SAORSA_ROOT_DIR is also set.

Details are in the inline comment on src/node.rs.

mickvandijke:
@greptile-apps @claude please re-review

mickvandijke:
@greptile-apps @claude please do a final review

dirvine (approved):
Great work Mick! 🎉

Thorough review complete — this is solid engineering across the board:

- **LMDB migration is clean** — same API shape, ACID transactions, zero-copy reads, no more filesystem scanning at startup
- **TOCTOU fix in `put()` is textbook correct** — existence check + capacity enforcement + write all atomic inside a single write txn
- **Removing the `AtomicU64` counter** in favour of querying LMDB's B-tree metadata directly was the right call — single source of truth, no drift
- **Identity persistence** with peer-ID-scoped directories is well designed — handles first-run, single-identity, and multi-identity cases cleanly
- **Test coverage is comprehensive** — persistence across reopen, capacity enforcement, identity restart, e2e chunk persistence
- **All prior review feedback addressed** — race conditions, stale comments, error propagation, tempdir fixes, JoinHandle await

The `unsafe` scope is minimal and well-justified with a clear SAFETY comment. CI green on all three platforms.

Ship it! 🚀

dirvine:
Hey @mickvandijke — PR approved and merged! 🎉

Really clean work on the LMDB migration and identity persistence. The iterative refinement across the 12 commits was great to see — each round of feedback made the code tighter. The TOCTOU fix, removing the atomic counter in favour of LMDB metadata, and the peer-ID-scoped directory layout are all solid architectural decisions.

Thanks for the thorough test coverage too — the e2e chunk persistence test especially gives good confidence in the restart path.

Nice one! 🚀

PR #14, opened 2026-02-26
feat: EVM payment integration for client chunk storage

grumbach:
## What

End-to-end EVM payment flow for storing chunks on saorsa-node. Clients request quotes from network nodes, pay on-chain (Arbitrum / local Anvil), and submit payment proofs with chunk storage requests. Nodes verify proofs before accepting data.

## Why

Storage on the saorsa network must be paid. This PR implements the full cycle: quote, pay, store, verify — mirroring the autonomi MerklePaymentVault contract pricing model. Without this, nodes accept data for free with no economic incentive to run a node.

## Architecture

```
Client                          Network Nodes                    EVM (Arbitrum/Anvil)
  |                                  |                                |
  |-- find 5 closest peers (DHT) -->|                                |
  |-- request quotes -------------->| generate ML-DSA-65 signed quote|
  |<-------------- 5 quotes -------|                                |
  |-- select median, pay 3x -------|----- on-chain payment -------->|
  |<-------------------------------|------- tx confirmation --------|
  |-- PUT chunk + payment proof -->| verify signatures (blocking)   |
  |                                | verify on-chain payment ------>|
  |<-------------- stored ---------|                                |
```

### Payment flow (`src/payment/`)
- **Pricing** — Logarithmic capacity-based pricing: empty nodes charge minimum, nearly-full nodes charge exponentially more, creating natural load balancing
- **Quotes** — ML-DSA-65 signed quotes with signer probe at startup to fail fast on bad keys
- **Verification** — Proof size validation, exactly 5 quotes required, duplicate peer rejection, xorname binding checks, signature verification offloaded to blocking threads
- **Caching** — LRU cache for verified addresses with atomic stats counters
- **Metrics** — Quoting metrics with persistence debounce (every 10 ops) and Drop flush

### Client (`src/client/`)
- `put_chunk_with_payment()` — full quote > pay > store flow via DHT peer discovery
- `put_chunk_with_proof()` — store with pre-built proof (skip payment cycle)
- File chunking: 4MB chunks + manifest for multi-chunk file upload/download

### Unified CLI (`src/bin/saorsa-cli/`)
- `file upload/download` — multi-chunk file operations with EVM payment
- `chunk put/get` — single-chunk operations (stdin/stdout support)
- Wallet via `SECRET_KEY` env var (no CLI flags for secrets)
- `saorsa-client` binary removed — all functionality merged into `saorsa-cli`

### Node enforcement (`src/node.rs`, `src/storage/handler.rs`)
- Production nodes block startup if payment is disabled
- Storage handler verifies payment proof before accepting chunks
- `max_message_size` configured for 4MB chunk transport

### E2E test suite (`tests/e2e/`)
- `complete_payment_e2e` — 10-node network with full quote > pay > store > retrieve
- `payment_flow` — payment workflow tests (quotes, median pricing, caching)
- `security_attacks` — forged signatures, wrong amounts, replay attacks, oversized proofs
- `scripts/test_e2e.sh` — shell-based E2E: upload, download, SHA256 verify, on-chain TX verification, payment rejection

## Stats

- 44 files changed, +7017 / -1160
- 245 unit tests, 13 security E2E tests
- code clean (zero warnings, zero errors)


mickvandijke (changes requested):
## Review: Two critical correctness issues

### 1. Double payment in `test_complete_payment_flow_live_nodes` — Steps 3-5 are dead code

In `tests/e2e/complete_payment_e2e.rs`, the test manually performs quote collection (Step 3), creates a `SingleNodePayment` (Step 4), and pays on-chain (Step 5). Then Step 6 calls `client.put_chunk()`.

Because the client has a wallet configured (via `set_wallet` in setup), `put_chunk()` delegates to `put_chunk_with_payment()`, which performs the **entire** quote→pay→store cycle again internally. This means:

- The on-chain payment from Step 5 is made but **never attached to any chunk store** — it's thrown away
- `put_chunk()` in Step 6 makes a **second, independent on-chain payment**
- The assertions in Steps 3-5 validate the first payment, but the chunk is actually stored using a completely different second payment

**Steps 3-5 are effectively dead code** that don't influence the actual chunk storage. The test passes but doesn't validate what it claims.

**Fix:** Step 6 should use `client.put_chunk_with_proof()` with the proof built from Steps 3-5, rather than `client.put_chunk()` which ignores all the manual work above it.

### 2. `TestAnvil::create_funded_wallet()` spawns a separate Anvil per call — wallets operate on wrong blockchain

In `tests/e2e/anvil.rs`, both `create_funded_wallet()` and `create_empty_wallet()` call `Testnet::new().await` internally, which spawns a **new Anvil process** with its own contract deployments each time. Wallets created this way have a `network` field pointing to a completely different set of contract addresses and chain state than the shared test Anvil.

This means any test using `TestAnvil::create_funded_wallet()` (e.g., `test_chunk_store_with_payment` in `chunk.rs`) is paying on one Anvil instance and the nodes are verifying on another. It works today only because EVM verification is disabled in these tests — if it were enabled, the payments would silently fail because the contract addresses don't match.

**Fix:** `create_funded_wallet()` should accept the shared `Testnet`/`Network` as a parameter rather than creating its own, or the `TestAnvil` struct should hold a `Testnet` instance that all wallets share.

grumbach:
@greptile-apps @claude please review

grumbach:
@greptile-apps @claude please review, assuming a fix is coming for the pricing

grumbach:
@greptile-apps @claude please review,

grumbach:
@greptile-apps @claude please review again

grumbach:
@claude @greptile-apps please review

grumbach:
@claude @greptile-apps one more review

grumbach:
@claude @greptile-apps one hopefully last review

dirvine:
## Multi-Model Code Review — PR #14: EVM Payment Integration

**Reviewed by:** Claude Sonnet 4.6 · MiniMax M2.1 · Z.AI GLM-5

Thanks Anselme — this is a really solid piece of work. The pricing architecture, proof structure, production enforcement, and 13-test security suite are all well thought out. Good luck with the fixes below! 🙌

---

## Verdict: Do Not Merge (2 Critical blockers)

The foundational design is sound — logarithmic capacity pricing, `SingleNode` median payment strategy, ML-DSA-65 post-quantum quote signatures, and hard production enforcement are all in good shape. Two critical issues block merge.

---

## 🔴 Critical (Blocking)

### C1. `pub_key` not bound to `encoded_peer_id`

The code itself acknowledges this:

```rust
// TODO: Verify that quote.pub_key belongs to encoded_peer_id.
// Currently we verify the signature is valid for the pub_key IN the quote,
// but don't verify that pub_key actually belongs to the claimed peer.
```

`encoded_peer_id` is a libp2p Ed25519 `PeerId`. `quote.pub_key` is an ML-DSA-65 key. The verifier confirms the signature is self-consistent but **never proves the ML-DSA key belongs to that peer**.

**The attack:** An attacker can construct a `ProofOfPayment` using their own 5 ML-DSA keypairs, mapping them to any 5 legitimate peer IDs. The verifier accepts (signatures valid ✅, content address matches ✅, no duplicate peer IDs ✅). The attacker still has to pay on-chain, but they now control the payment identity, can manipulate median prices, and can impersonate nodes.

**Root cause:** Dual identity system (libp2p Ed25519 for DHT + ML-DSA-65 for payments) with no cryptographic bridge.

**Best fix:** Adopt the unified ML-DSA-65 `PeerId = BLAKE3(ML-DSA pubkey)` from saorsa-core (PR #32, merged today) as the payment identity too. This eliminates the dual-identity problem entirely — nodes use the same key for DHT routing and quote signing.

---

### C2. `NodeId` → `PeerId` compilation conflict

```rust
// node.rs — this PR
use saorsa_core::identity::{NodeId, NodeIdentity};
```

PR #32 (merged today) renamed `NodeId → PeerId` throughout saorsa-core. **This PR will not compile against current `main`.** Needs a rebase + rename pass.

---

## 🟠 High (Must Fix)

### H1. No quote timestamp expiry

`PaymentQuote.timestamp` exists but is never validated server-side. An attacker can obtain a cheap quote when the network is empty, then replay it against a nearly-full node (where prices should be high). The on-chain payment only confirms *a* payment was made at the quoted amount — not that the quote is still valid.

**Fix:**
```rust
const QUOTE_MAX_AGE_SECS: u64 = 300; // 5 minutes
for (_, quote) in &payment.peer_quotes {
    let age = SystemTime::now().duration_since(quote.timestamp)
        .map_err(|_| Error::Payment("Quote timestamp in future"))?;
    if age.as_secs() > QUOTE_MAX_AGE_SECS {
        return Err(Error::Payment("Quote expired"));
    }
}
```

### H2. Median manipulation via DHT positioning

The client asks the 5 DHT-closest nodes for quotes. Since saorsa-core has no PoW on DHT positioning (see PR #32 review), an attacker with enough compute can grind keypairs to control 3+ of the 5 closest slots for any target `XorName`, then set median = 0 (free storage) or MAX_PRICE (censorship). This is inherently linked to C1 — fixing the identity binding reduces the attack surface significantly.

### H3. Pricing formula edge case near 100% capacity

The formula uses `(r - 1.0).abs().ln()`. Near capacity, `abs()` means the formula increases on both sides of 1.0, potentially making prices *lower* than expected at 99.9% fill rather than spiking to infinity. The epsilon guard prevents a crash but the economic curve may not match the Solidity `MerklePaymentVault` contract.

**Recommended:** Add unit tests comparing `calculate_price()` output against the contract at 95%, 99%, 99.9%, and 100% fill.

---

## 🟡 Medium (Should Fix)

- **M1. LRU cache not persisted** — node restart flushes all 100k entries, forcing full on-chain re-verification on every subsequent PUT. Consider optional RocksDB/LMDB persistence.
- **M2. 4/5 quoting nodes get `Amount::ZERO`** — expected value per quote is 0.6× quoted price. The 3× multiplier compensates, but monitor for nodes de-prioritising quote responses in production.
- **M3. EVM disabled misconfiguration** — if EVM chain ID matches Arbitrum One (42161), consider forcing `enabled = true` regardless of config to prevent accidental free-storage on mainnet.

---

## ✅ What's Good

- `spawn_blocking` for ML-DSA-65 signature verification — correct pattern
- Production startup guard (hard error if payment disabled in production mode)
- Size bounds on proof bytes (`MIN = 32`, `MAX = 100KB`)
- XorName binding check prevents proof reuse across different content addresses
- 13 security E2E tests covering replay, forgery, oversized proofs, duplicate peer IDs
- `probe_signer()` at startup to fail fast on bad keys
- Pricing formula generally sound (logarithmic load balancing intent is correct)

---

## Blocking Action Items

1. **[Critical]** Implement `pub_key ↔ encoded_peer_id` cryptographic binding — best path is unifying identities with PR #32's ML-DSA-65 `PeerId`.
2. **[Critical]** Rebase onto main, rename `NodeId → PeerId` throughout.
3. **[High]** Add `QUOTE_MAX_AGE_SECS = 300` expiry check in `verify_evm_payment()`.
4. **[High]** Add pricing formula tests at capacity boundaries vs. the Solidity contract.

Good work on the overall design Anselme — once those four items are addressed this is close to ready. 💪

dirvine (dismissed):
Thanks for the substantial payment integration work — this is very close, but I’m requesting changes for a few important issues before merge.

## Blocking findings

1) **Missing signer-to-peer binding in quote verification**
- File: `src/payment/verifier.rs` (see TODO around `verify_evm_payment`)
- Current logic verifies quote content + ML-DSA signature validity, but does not verify that `quote.pub_key` belongs to the claimed `encoded_peer_id`.
- This leaves a trust gap in peer attribution.
- **Requested change:** cryptographically bind signer identity to claimed peer identity (or include claimed peer identity in signed payload), and add an attack test for "valid signature, wrong claimed peer".

2) **Node acceptance does not prove this node is among paid recipients**
- Files: `src/storage/handler.rs`, `src/payment/verifier.rs`
- `handle_put()` calls `verify_payment()` without local-node payment ownership validation.
- Verification confirms payment validity generally, but not that the receiving node itself is one of the intended paid peers.
- **Requested change:** include local node identity/rewards address in verifier context and reject proofs that don’t include this node in the paid set.

## Non-blocking but should be addressed promptly

3) **Pricing/capacity mismatch can underprice after nominal fullness**
- Files: `src/payment/pricing.rs`, `src/node.rs`, `src/config.rs`
- Pricing returns min price at/after fullness boundary, while default storage max chunks is unlimited.
- **Requested change:** align enforceable capacity and pricing behavior (or clamp fullness near 1.0 instead of dropping to min).

4) **`config/production.toml` schema mismatch with `NodeConfig`**
- The template appears to use keys/sections that do not match runtime config struct fields.
- **Requested change:** update template so operators can apply it directly without silent misconfiguration.

Once the two blocking items are addressed, I’m happy to re-review quickly.

grumbach:
@claude review

grumbach:
@claude review

mickvandijke (approved):
LGTM!

PR #19, opened 2026-03-09
fix: disable geo IP diversity checks in devnet

mickvandijke:
## Summary
- Devnet nodes were using production `IPDiversityConfig` defaults (geo checks, ASN limits, subnet limits), causing unnecessary restrictions during local testing
- Set `IPDiversityConfig::permissive()` on devnet node configs to disable all diversity enforcement

## Test plan
- [ ] Start a local devnet and verify nodes connect without diversity-related rejections
- [ ] Confirm production node startup is unaffected (does not use devnet code path)


dirvine (approved):
Thanks Mick. I checked this change and the devnet-only diversity override is scoped correctly.

PR #20, opened 2026-03-09
chore: bump saorsa-core to 0.14.1 and saorsa-pqc to 0.5

mickvandijke:
## Summary
- Bump `saorsa-core` from 0.14.0 to 0.14.1
- Bump `saorsa-pqc` from 0.4.0 to 0.5

## Test plan
- [ ] Verify `cargo build --release` succeeds
- [ ] Verify `cargo test` passes
- [ ] Verify node starts and connects to bootstrap peers


dirvine (approved):
Thanks Mick. I checked the dependency bump and it looks good; the earlier CI failure was dependency publication timing rather than a source regression.

PR #23, opened 2026-03-11
feat: integrate self-encryption for streaming file encrypt/decrypt

grumbach:
Replace plaintext file chunking with convergent encryption via the self_encryption crate (quantum-safe: ChaCha20-Poly1305 + BLAKE3).

- Add src/client/self_encrypt.rs with streaming encrypt/decrypt, DataMap serialization, and public/private data mode support
- Update CLI with --public upload flag and --datamap download option
- Set plaintext chunk size to 4MB-4KB via .cargo/config.toml to leave headroom for AEAD tag and compression overhead
- Point self_encryption dep at grumbach/post_quatum branch


grumbach:
@claude @greptile-apps please review

grumbach:
@claude @greptile-apps please review again. Also, in private mode it's normal that the datamap is printed to the user.

.cargo/config.toml
```diff
@@ -0,0 +1,8 @@
+# Set the plaintext chunk size for self_encryption to leave headroom
+# for encryption overhead (ChaCha20-Poly1305 AEAD tag: 16 bytes,
+# Brotli framing: ~50 bytes). This ensures encrypted chunks always
+# fit within saorsa-node's MAX_CHUNK_SIZE (4 MiB = 4,194,304 bytes).
+#
+# Value: 4 MiB - 4 KiB = 4,190,208 bytes (4 KiB headroom).
+[env]
+MAX_CHUNK_SIZE = "4190208"
```
mickvandijke: **Does this env var actually do anything?**

This sets `MAX_CHUNK_SIZE` as a Cargo environment variable, but `saorsa_node::ant_protocol::MAX_CHUNK_SIZE` is a Rust `const` — it's not read via `env!()` or `std::env::var()`. This config only has an effect if the `self_encryption` crate itself reads `MAX_CHUNK_SIZE` from the environment at build time or runtime.

Could you verify that the `self_encryption` crate actually consumes this? If not, this file is dead config giving a false sense of safety — the plaintext chunk size wouldn't actually be constrained.

src/client/self_encrypt.rs
```diff
@@ -0,0 +1,833 @@
...
+            // Await the next completed upload
+            let (num, hash, result) = in_flight
+                .next()
+                .await
+                .ok_or_else(|| Error::Crypto("Upload stream unexpectedly empty".into()))?;
+            match result {
+                Ok((address, tx_hashes)) => {
+                    if address != hash.0 {
+                        // Drain remaining in-flight futures before returning
+                        let mut succeeded = all_tx_hashes.len();
```
mickvandijke: **`succeeded` counts tx hashes, not chunks**

`all_tx_hashes.len()` gives the number of *transaction hashes* accumulated so far, which can be multiple per chunk. But the warning message says `"{succeeded} chunk(s) already uploaded"`. This will overcount and produce misleading diagnostics.

Same issue on line 150.

Consider tracking a separate `uploaded_chunks` counter that increments once per successful upload result, independent of how many tx hashes each upload returns.

src/client/self_encrypt.rs
```diff
@@ -0,0 +1,833 @@
...
+        "Encrypting file: {} ({file_size} bytes)",
+        file_path.display()
+    );
+
+    let file = std::fs::File::open(file_path).map_err(Error::Io)?;
+    let mut reader = BufReader::new(file);
+    let read_error: Arc<Mutex<Option<std::io::Error>>> = Arc::new(Mutex::new(None));
+    let read_error_writer = Arc::clone(&read_error);
+    let data_iter = std::iter::from_fn(move || {
+        let mut buf = vec![0u8; 65536];
```
mickvandijke: **Magic number + per-call allocation**

Two small things here:
1. `65536` is a magic number — consider extracting it to a named constant like `const READ_BUFFER_SIZE: usize = 64 * 1024;` to make the intent clear.
2. This allocates a fresh 64 KiB `Vec` on every `next()` call. For a large file that's thousands of allocations and a lot of heap churn. You could reuse a buffer across calls (e.g., keep a `Vec<u8>` outside the closure and `Bytes::copy_from_slice(&buf[..n])` each iteration), or use `BytesMut` with a fixed capacity.

src/client/self_encrypt.rs
```diff
@@ -0,0 +1,833 @@
...
+        .into_datamap()
+        .ok_or_else(|| Error::Crypto("DataMap not available after encryption".into()))?;
+
+    info!("All {chunk_num} encrypted chunks uploaded");
+    Ok((data_map, all_tx_hashes))
+}
+
+/// Encrypt a file from disk using `stream_encrypt`, returning the `DataMap`
+/// and a list of `(XorName, Bytes)` encrypted chunks.
+fn encrypt_file_to_chunks(
```
mickvandijke, line 257: **Near-identical code to `encrypt_and_upload_file` (lines 60–82)**

The file-open, `BufReader`, `Arc<Mutex<Option<io::Error>>>`, `data_iter` construction, `stream_encrypt` call, and post-iteration error check here are ~30 lines copy-pasted from `encrypt_and_upload_file`. If the error-handling pattern changes (e.g., the I/O error propagation gets refined), it would need updating in both places.

Consider extracting a shared helper like:
```rust
fn open_encrypt_stream(file_path: &Path, file_size: usize)
    -> Result<(EncryptStream, Arc<Mutex<Option<std::io::Error>>>)>
```
that both callers use.

mickvandijke (commented):
Really nice work on this — the streaming self-encryption integration is well-structured, the bounded-concurrency upload loop is a solid design, and the public/private data mode split is clean. The error-handling discipline (draining in-flight futures to preserve tx hashes on failure, atomic temp-file writes) shows good attention to production edge cases. The test suite is thorough too.

A few things I noticed below, mostly around deduplication and minor polish:

src/client/self_encrypt.rs
```diff
@@ -0,0 +1,833 @@
...
+pub fn decrypt_file_local<S: BuildHasher>(
+    data_map: &DataMap,
+    chunk_store: &HashMap<XorName, Bytes, S>,
+    output_path: &Path,
+) -> Result<()> {
+    decrypt_from_store(data_map, chunk_store, output_path)
+}
+
+/// Shared helper: decrypt a `DataMap` using a chunk store `HashMap`.
+fn decrypt_from_store<S: BuildHasher>(
```
mickvandijke, line 451: **Temp-file-write-and-rename logic duplicated from `download_and_decrypt_file` (lines 306–338)**

This entire block — create tmp path, write chunks in a closure, cleanup on error, rename on success — is a verbatim copy of the same pattern ~130 lines above. Both even use the same `.saorsa_decrypt_` prefix.

A shared helper like:
```rust
fn write_stream_to_file(
    stream: impl Iterator<Item = Result<Bytes, self_encryption::Error>>,
    output_path: &Path,
) -> Result<()>
```
would eliminate the duplication and ensure any future fixes (e.g., Windows rename handling) only need to be made once.

src/client/self_encrypt.rs
```diff
@@ -0,0 +1,833 @@
...
+        if let Err(cleanup_err) = std::fs::remove_file(&tmp_path) {
+            warn!(
+                "Failed to remove temp file {}: {cleanup_err}",
+                tmp_path.display()
+            );
+        }
+        return Err(e);
+    }
+
+    std::fs::rename(&tmp_path, output_path).map_err(Error::Io)?;
```
mickvandijke: **`std::fs::rename` fails on Windows if destination exists**

On Windows, `std::fs::rename` returns an error if `output_path` already exists (unlike Unix where it atomically replaces). Since this is a cross-platform library (CI builds and tests on Windows), consider either removing the destination first or using `tempfile::NamedTempFile::persist()` which handles this portably. Same applies to the identical rename on line 480.

src/client/self_encrypt.rs
```diff
@@ -0,0 +1,833 @@
...
+                Ok(results)
+            })
+        })
+    })
+    .map_err(|e| Error::Crypto(format!("Decryption failed: {e}")))?;
+
+    // Write to a temp file first, then rename atomically on success
+    // to prevent leaving a corrupt partial file on failure.
+    let parent = output_path.parent().unwrap_or_else(|| Path::new("."));
+    let unique: u64 = rand::random();
```
mickvandijke: **Consider using `tempfile` instead of `rand::random` for tmp files**

`tempfile` is already a dev-dependency (used in tests). Promoting it to a regular dependency and using `tempfile::Builder::new().prefix(".saorsa_decrypt_").tempfile_in(parent)?` would give you guaranteed-unique temp paths with atomic creation, and `NamedTempFile::persist()` handles the cross-platform rename issue too. This would replace the manual `rand::random()` uniqueness, the manual cleanup on error, and the manual rename — all in one.

Same on line 454.

src/lib.rs
```diff
@@ -54,6 +54,10 @@ pub use ant_protocol::{
     ChunkGetRequest, ChunkGetResponse, ChunkMessage, ChunkMessageBody, ChunkPutRequest,
     ChunkPutResponse, ChunkQuoteRequest, ChunkQuoteResponse, CHUNK_PROTOCOL_ID, MAX_CHUNK_SIZE,
 };
+pub use client::self_encrypt::{
+    deserialize_data_map, download_and_decrypt_file, encrypt_and_upload_file,
+    fetch_data_map_public, serialize_data_map, store_data_map_public,
+};
 pub use client::{
     compute_address, create_manifest, deserialize_manifest, peer_id_to_xor_name, reassemble_file,
```
mickvandijke: **Are the old `file_ops` exports still needed?**

`create_manifest`, `deserialize_manifest`, `reassemble_file`, `serialize_manifest`, `split_file` — these are the plaintext chunking functions that self-encryption replaces. If they're still needed for backward compatibility with data already stored on the network, a comment noting that would be helpful. If not, consider deprecating or removing them to avoid confusion about which path to use.

src/lib.rs
```diff
@@ -54,6 +54,10 @@ pub use ant_protocol::{
     ChunkGetRequest, ChunkGetResponse, ChunkMessage, ChunkMessageBody, ChunkPutRequest,
     ChunkPutResponse, ChunkQuoteRequest, ChunkQuoteResponse, CHUNK_PROTOCOL_ID, MAX_CHUNK_SIZE,
 };
+pub use client::self_encrypt::{
```
mickvandijke, line 57: **High: Encrypted-chunk headroom fix doesn't apply to downstream library consumers**

The reduced plaintext chunk size (4 MiB - 4 KiB) lives only in `.cargo/config.toml`, which sets a Cargo env var scoped to *this* workspace. But this PR publicly re-exports the self-encryption API right here.

I verified by building `saorsa-node` as a dependency from a clean temp crate outside the repo root — `self_encryption::MAX_CHUNK_SIZE` resolved to `4194304` (full 4 MiB), not `4190208`. So downstream library consumers calling `encrypt_and_upload_file` can still produce encrypted chunks at the full 4 MiB plaintext size and hit oversize uploads against Saorsa's 4 MiB chunk limit.

The headroom constraint needs to be enforced at the library level (e.g., a runtime assertion, a wrapper constant, or configuring `self_encryption` programmatically) rather than relying on a workspace-local env var that disappears when the crate is consumed externally.

grumbach: 100% let's move this to self_encryption as a default

grumbach:
@claude @greptile last review

src/client/self_encrypt.rs
```diff
@@ -0,0 +1,819 @@
...
+}
+
+/// Check whether the read-error capture from `open_encrypt_stream` recorded
+/// an I/O error during iteration.
+fn check_read_error(read_error: &ReadErrorCapture) -> Result<()> {
+    if let Ok(guard) = read_error.lock() {
+        if let Some(ref e) = *guard {
+            return Err(Error::Io(std::io::Error::new(e.kind(), format!("{e}"))));
+        }
+    }
```
mickvandijke, line 82: **Poisoned mutex silently drops I/O error**

If the read iterator panics (poisoning the mutex), `lock()` returns `Err` and this function returns `Ok(())` — silently swallowing whatever I/O error was captured. This means a panicking read path produces a successful result with corrupt/truncated data.

```suggestion
    let guard = read_error.lock().unwrap_or_else(|e| e.into_inner());
    if let Some(ref e) = *guard {
        return Err(Error::Io(std::io::Error::new(e.kind(), format!("{e}"))));
    }
    Ok(())
```

`unwrap_or_else(|e| e.into_inner())` recovers the guard even when poisoned, so the captured error is always checked.

src/client/self_encrypt.rs
```diff
@@ -0,0 +1,819 @@
...
+            );
+        }
+        return Err(e);
+    }
+
+    // On Windows, rename fails if destination exists. Remove it first.
+    if output_path.exists() {
+        std::fs::remove_file(output_path).map_err(Error::Io)?;
+    }
+    std::fs::rename(&tmp_path, output_path).map_err(Error::Io)?;
```
mickvandijke, line 125: **TOCTOU race in Windows rename workaround**

`exists()` + `remove_file()` is not atomic — another process (or concurrent decrypt) could create a file at `output_path` between the two calls, causing the subsequent `rename` to fail on Windows anyway.

`tempfile` is already a regular dependency (used for `rand` in the tmp name). Consider using `tempfile::NamedTempFile` with `.persist()` which handles the cross-platform atomic-replace correctly, and would also eliminate the manual cleanup logic above.

src/client/self_encrypt.rs
```diff
@@ -0,0 +1,819 @@
...
+    info!("All {chunk_num} encrypted chunks uploaded");
+    Ok((data_map, all_tx_hashes))
+}
+
+/// Encrypt a file from disk using `stream_encrypt`, returning the `DataMap`
+/// and a list of `(XorName, Bytes)` encrypted chunks.
+fn encrypt_file_to_chunks(
+    file_path: &Path,
+    file_size: usize,
+) -> Result<(DataMap, Vec<(XorName, Bytes)>)> {
```
mickvandijke, line 260: **All encrypted chunks collected in memory**

`encrypt_and_upload_file` now streams with bounded concurrency, but this function (used by the `pub` `encrypt_file_local`) still collects every chunk into a `Vec`. For a large file this means the full encrypted size in RAM.

Since `encrypt_file_local` is public API, library consumers could hit this unknowingly. At minimum, add a doc note on `encrypt_file_local` warning about peak memory being proportional to file size. Longer term, consider a callback/visitor pattern so callers can process chunks incrementally.

src/client/self_encrypt.rs
```diff
@@ -0,0 +1,819 @@
...
+
+                let mut results = Vec::with_capacity(futs.len());
+                while let Some((idx, hash, result)) = futs.next().await {
+                    let addr_hex = hex::encode(hash.0);
+                    let chunk = result
+                        .map_err(|e| {
+                            self_encryption::Error::Generic(format!(
+                                "Network fetch failed for {addr_hex}: {e}"
+                            ))
+                        })?
```
mickvandijke, line 336: **Early `?` abandons remaining in-flight fetch futures**

If one chunk fetch fails in the `while let Some(...)` loop, the `?` on the result drops the `FuturesUnordered` with remaining futures still running. Those futures continue executing in the background, consuming network resources and Tokio worker time for chunks that will never be used.

Not a correctness issue, but for large batches it wastes resources. Consider collecting all results first and then checking for errors, similar to how `encrypt_and_upload_file` drains in-flight uploads on failure.

src/bin/saorsa-cli/main.rs
```diff
@@ -99,126 +112,98 @@ async fn main() -> color_eyre::Result<()> {
...
-    println!("FILE_ADDRESS={manifest_hex}");
-    println!("CHUNKS={chunk_count}");
-    println!("TOTAL_SIZE={file_size}");
-    println!("PAYMENTS={total_tx_count}");
-    println!("TX_HASHES={tx_hashes_str}");
+        info!("Upload complete (public): address={address_hex}, chunks={chunk_count}");
+    } else {
+        // Private mode: save DataMap locally, never upload it
+        let data_map_bytes = serialize_data_map(&data_map)?;
+        let datamap_path = path.with_extension("datamap");
```
mickvandijke, line 148: **DataMap (decryption key) written to predictable path without safeguards**

The datamap is the sole capability needed to decrypt the file. Two concerns:

1. **Silent overwrite**: If `secret.datamap` already exists (e.g., from a previous upload of the same filename), it is silently overwritten. The user loses their previous decryption key with no warning.

2. **Predictable location**: Anyone who knows the original filename knows where the datamap is. For a feature explicitly called "private mode", the key sitting next to the original file in a guessable path weakens the security posture.

Suggestion: warn (or error) if the datamap file already exists, and consider printing a reminder that this file must be kept secret.

mickvandijke (commented):
Review of final state (commit e916936). Five items — two correctness bugs, three worth discussing.

mickvandijke (approved):

PR #52, opened 2026-03-31
Remove --ip-version CLI flag, always use dual-stack

grumbach:
The --ip-version flag allowed users to restrict the node to IPv4-only or IPv6-only mode. This was flagged for removal as it enables misconfiguration ("we cannot go back to tweakers"). The node now always runs in dual-stack mode (both IPv4 and IPv6), which was already the default.

Changes:
- Remove --ip-version CLI arg and ANT_IP_VERSION env var
- Remove IpVersion enum from config
- Remove ip_version field from NodeConfig
- Remove CliIpVersion enum and From impl
- Hardcode ipv6=true in core config builder (dual-stack)
- Update production.toml, testnet scripts, and README

https://claude.ai/code/session_01ApfhBFUhWMnYoZj7P1VkvU

### DIFF
--- a/README.md
+++ b/README.md
@@ -907,7 +907,6 @@ You can also specify bootstrap peers explicitly:
 ./target/release/ant-node \
     --root-dir ~/.ant-node \
     --port 12000 \
-    --ip-version dual \
     --upgrade-channel stable \
     --migrate-ant-data auto \
     --log-level info
@@ -929,9 +928,6 @@ Options:
         Listening port (0 for automatic selection)
         [default: 0]
 
-    --ip-version <VERSION>
-        IP version to use: ipv4, ipv6, or dual
-        [default: dual]
 
     --bootstrap <ADDR>
         Bootstrap peer socket addresses (can be specified multiple times)
@@ -1016,7 +1012,6 @@ export ANT_UPGRADE_CHANNEL=stable
 ```toml
 root_dir = "~/.ant-node"
 port = 0  # Auto-select
-ip_version = "dual"
 bootstrap = [
     "10.0.0.1:10000",
     "10.0.0.2:10000",
--- a/README.md
+++ b/README.md
@@ -928,6 +928,9 @@ Options:
         Listening port (0 for automatic selection)
         [default: 0]
 
+    --ipv4-only
+        Force IPv4-only mode (disable dual-stack).
+        Use on hosts without working IPv6.
 
     --bootstrap <ADDR>
         Bootstrap peer socket addresses (can be specified multiple times)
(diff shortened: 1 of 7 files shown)


mickvandijke (commented):
## Deep Review: Address Reachability Risk

### What the PR does

Clean removal — `IpVersion` enum, `CliIpVersion`, `--ip-version` CLI arg, `ANT_IP_VERSION` env var, `ip_version` config field all gone. `ipv6` is now hardcoded to `true` (dual-stack) in `build_core_config`. The code change itself is correct and well-scoped.

### Problem: two failure modes on hosts without working IPv6

The old `--ip-version` flag existed for a reason. With it removed and dual-stack hardcoded, there are two scenarios where this breaks:

**1. IPv4-only kernel (no IPv6 support at all)**

saorsa-core's `DualStackNetworkNode::new_with_options()` tries to bind `[::]:<port>`. On an IPv4-only kernel this fails with `EAFNOSUPPORT` or `EADDRNOTAVAIL`. The error propagates via `?` with **no fallback** — the node crashes on startup.

**2. IPv6-capable kernel but no routable IPv6 address (the worse case)**

The bind to `[::]` *succeeds* (kernel supports it), but the machine has no global IPv6 address — only link-local `fe80::` or nothing routable. The node starts fine but:
- Advertises an unreachable IPv6 address to the DHT
- saorsa-core does **not** validate routability before advertising — it only filters unspecified (`::`, `0.0.0.0`), multicast, and loopback addresses. There's no `is_global()` check, no STUN/AutoNAT probe.
- Peers accept the address into their routing tables (it passes all filters)
- Peers waste time trying to dial it — hole-punching/relay fallback chain is 5s → 15s → 30s before giving up
- This silently degrades network health with no obvious error on the node itself

### Why this matters

- The testnet scripts (`spawn-nodes.sh`, `start-genesis.sh`) previously hardcoded `--ip-version ipv4` — likely because those VPS nodes have unreliable or absent IPv6 routing
- Many cloud instances (older providers, some DigitalOcean/Hetzner configs) and corporate networks are IPv4-only or have non-routable IPv6
- There's now no workaround for operators who know their network is IPv4-only

### Existing config file compatibility

No issue here — `NodeConfig` doesn't use `deny_unknown_fields`, so old TOML files with `ip_version = "dual"` are silently ignored. Good.

### Missed cleanup

`README.md:1002` still lists `export ANT_IP_VERSION=dual` in the environment variables section.

### Suggestion

The goal of preventing misconfiguration ("we cannot go back to tweakers") is valid, but removing the escape hatch entirely is unsafe without routability detection in saorsa-core. Options:

1. **Keep dual-stack as default, add `--ipv4-only` opt-out** — simpler than the old 3-way enum, still prevents casual tweaking (it's an explicit opt-out rather than a menu), and gives operators the escape hatch they need
2. **Add a fallback in `build_core_config`** — try dual-stack, if bind fails retry with `ipv6=false`. Only catches failure mode #1, not #2.
3. **Fix in saorsa-core** — add routability detection before advertising addresses. The right long-term fix but a larger scope change.

I'd recommend option 1 as the pragmatic fix for this PR, with option 3 tracked as follow-up work.

mickvandijke (approved):

PR #68, opened 2026-04-09
fix: clock skew tolerance 30s -> 5min (stacks on #67 -> #66)

grumbach:
**Stacked on #67 (Mick's NAT fixes) -> #66 (workspace consolidation).**

One constant change: `MAX_FUTURE_SECS: 30 -> 300`.

Without this: 0/3 uploads from Mac (31s clock skew caused all messages to be rejected as future-dated). With this: 3/3 uploads succeeded.

A decentralized network cannot partition when a consumer device's clock drifts by 31 seconds. The past window was already 5 minutes; the future window was asymmetrically tight at 30 seconds.

492 ant-node tests pass.


grumbach:
outdated

PR #77, opened 2026-04-22
fix(payment): block merkle pay-yourself via DHT closeness check

grumbach:
## Summary

- Block the merkle "pay-yourself" attack by adding a DHT-closeness check to `verify_merkle_payment`.
- Attach the `P2PNode` to `PaymentVerifier` at startup so the check can query the live routing table.
- Add an e2e test that fails on `main` (verifier accepts 16 attacker-generated keys all rewarding the payer) and passes with the fix.

## The attack

`verify_merkle_payment` checked candidate ML-DSA signatures, timestamps, the cryptographic merkle proof, and on-chain paid-node bindings — but nothing tied the 16 candidate `pub_keys` to the pool midpoint's close group on the network. An attacker could generate 16 ML-DSA keypairs locally, point all 16 `reward_address` fields at a self-owned wallet, fund the on-chain merkle payment themselves, and receive their own rewards back.

## The fix

In `verify_merkle_payment`, between the cheap signature check and the on-chain RPC:

1. Pull the pool midpoint address from `winner_pool.midpoint_proof.address()`.
2. Call `dht_manager().find_closest_nodes_network(&pool_address.0, 16)` with a 60s timeout.
3. Derive `PeerId = BLAKE3(pub_key)` for each of the 16 candidates.
4. Require at least 13/16 candidate `PeerId`s to appear in the DHT-returned set.

A `MerklePaymentProof` carries exactly one `winner_pool`, so every storing node that receives the proof runs the same check against that single pool — a forged pool is rejected everywhere it lands.

The check runs *before* the on-chain lookup so a forged proof never pays for an RPC round-trip.

## Concurrency & DoS hardening

After two rounds of adversarial review:

- **Pass cache** (`closeness_pass_cache`): within a 256-chunk batch sharing one pool, only the first chunk pays the Kademlia lookup; subsequent chunks hit the cache.
- **Single-flight** (`inflight_closeness`): `N` concurrent PUTs for the same `pool_hash` coalesce onto one lookup. The leader publishes its `Result<(), String>` through an `Arc<ClosenessSlot>` that holds a `Notify` + `OnceLock`; waiters read the published result directly instead of re-running `_inner`. Bounds DoS amplification to one lookup per unique `pool_hash` for BOTH success and failure cases.
- **Lost-wakeup safety**: waiters use `Arc<Notify>::notified_owned()` so the `notify_waiters` counter is snapshotted while the inflight lock is still held; a race where the leader fires notify between the lock drop and the poll is handled by the counter check on first poll.
- **LRU-eviction safety**: the leader's `InflightGuard` owns its own `Arc<ClosenessSlot>` independent of the cache. `notify_waiters` fires on the owned slot regardless of eviction; the cache pop is gated on an `Arc::ptr_eq` identity check so an eviction + reinsertion doesn't make the original leader pop someone else's entry.
- **Cancellation-cascade bound**: waiters retry up to `MAX_LEADER_RETRIES = 4` times if the leader gets cancelled before publishing, then return a visible error rather than spinning forever.

## Fail-closed in production

`PaymentVerifier::attach_p2p_node` is called at node startup in `node.rs`, `devnet.rs`, and the e2e testnet. If the call is ever missed:
- **Release builds**: closeness check returns `Err` with a visible error log — any regression surfaces as a rejected PUT, not a silent accept.
- **Test / `test-utils` builds**: falls back to fail-open so unit tests that don't exercise merkle verification don't need a real DHT. `PaymentVerifier::new` emits a loud startup-time error log when built with `test-utils`, so a production binary accidentally built with the feature is visible in logs.

## Known limitation — Sybil-grinding

`MidpointProof::address()` is BLAKE3 of attacker-controllable inputs (leaf bytes, tree root, timestamp). An attacker who *also* runs 13+ Sybil DHT nodes can grind the midpoint until those Sybils are the true network-closest peers, at which point closeness passes for the attacker. Closing that gap requires either a Sybil-resistant identity layer or binding the midpoint to an attacker-uncontrolled value (on-chain VRF or block hash at payment time). This PR raises the attack cost from "free" to "run 13+ Sybil nodes AND grind" — a meaningful but not complete improvement. Documented in the code.

## Test plan

- [x] `test_attack_merkle_pay_yourself_fabricated_pool` fails on `main` with "Pay-yourself attack SUCCEEDED"
- [x] Same test passes with the fix (closeness rejection)
- [x] 3 new unit tests: pass-cache short-circuit, concurrent-readers agreement, waiter-reads-leader's-published-failure
- [x] All 43 verifier unit tests pass: `cargo test --lib payment::verifier`
- [x] All 8 merkle e2e tests pass: `cargo test --features test-utils --test e2e merkle -- --test-threads=1`
- [x] `cargo clippy --all-targets --all-features -- -D warnings` clean
- [x] `cargo fmt --all` clean
- [x] `RUSTDOCFLAGS="--deny=warnings" cargo doc --no-deps` clean
- [x] Full CI matrix (see checks)

## Follow-up (not in this PR)

- Bind `MidpointProof` derivation to an on-chain VRF or block hash to close the Sybil-grinding gap.
- Add a global rate limit on iterative Kademlia lookups to further bound cross-pool DoS.

### DIFF
--- a/src/devnet.rs
+++ b/src/devnet.rs
@@ -635,6 +635,11 @@ impl Devnet {
         *node.state.write().await = NodeState::Running;
 
         if let (Some(ref p2p), Some(ref protocol)) = (&node.p2p_node, &node.ant_protocol) {
+            // Wire the P2PNode into the payment verifier for merkle-closeness checks.
+            protocol
+                .payment_verifier_arc()
+                .attach_p2p_node(Arc::clone(p2p));
+
             let mut events = p2p.subscribe_events();
             let p2p_clone = Arc::clone(p2p);
             let protocol_clone = Arc::clone(protocol);
--- a/src/node.rs
+++ b/src/node.rs
@@ -130,6 +130,15 @@ impl NodeBuilder {
 
         let p2p_arc = Arc::new(p2p_node);
 
+        // Wire the P2PNode handle into the payment verifier so merkle-payment
+        // checks can query the live DHT for peers actually closest to a pool
+        // midpoint (pay-yourself defence).
+        if let Some(ref protocol) = ant_protocol {
+            protocol
+                .payment_verifier_arc()
+                .attach_p2p_node(Arc::clone(&p2p_arc));
+        }
+
         // Initialize replication engine (if storage is enabled)
         let replication_engine =
             if let (Some(ref protocol), Some(fresh_rx)) = (&ant_protocol, fresh_write_rx) {
--- a/src/payment/verifier.rs
+++ b/src/payment/verifier.rs
@@ -19,9 +19,12 @@ use evmlib::Network as EvmNetwork;
 use evmlib::ProofOfPayment;
 use evmlib::RewardsAddress;
 use lru::LruCache;
-use parking_lot::Mutex;
+use parking_lot::{Mutex, RwLock};
 use saorsa_core::identity::node_identity::peer_id_from_public_key_bytes;
+use saorsa_core::identity::PeerId;
+use saorsa_core::P2PNode;
 use std::num::NonZeroUsize;
+use std::sync::Arc;
 use std::time::SystemTime;
 
 /// Minimum allowed size for a payment proof in bytes.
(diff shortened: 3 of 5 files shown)


grumbach:
@greptile-apps @claude please review

jacderida (approved):

PR #98, opened 2026-05-18
fix(replication): guard prune confirmation safety

mickvandijke:
## Summary
- require remote prune-confirmation audit proofs before deleting out-of-range records
- increase prune hysteresis from 6 hours to 3 days before records or paid-list entries become prune-eligible
- preserve one bootstrap-claim grace window across sync, audit, and prune-confirmation paths
- keep expired active bootstrap claims audit-eligible so continued claims can be observed and penalized after the grace window
- avoid clearing active bootstrap claims on audit or prune-confirmation transport failures
- add focused unit coverage and a live e2e test for prune confirmation gating and missing peer proof deferral

## Tests
- cargo test --lib replication::audit
- cargo test --lib replication::pruning
- cargo test --lib replication::tests
- cargo test --lib bootstrap_claim
- cargo clippy --lib
- git diff --check
- cargo test --features test-utils --test e2e test_prune_pass_requires_remote_confirmation_before_delete

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -170,6 +170,8 @@ cobertura.xml
 test_data/
 tmp/
 temp/
+/test-*.bin
+/test-*.datamap
 
 # Benchmark results
 criterion/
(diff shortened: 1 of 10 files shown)


dirvine (approved):
Reviewed locally. The prune-confirmation flow is logically consistent with the replication/network design, and local fmt/check/clippy/targeted tests pass. No blocking issues found.

PR #99, opened 2026-05-19
fix(replication): bound replication queues with per-source fairness (D1 DoS)

grumbach:
## Summary

Security fix for **D1**: `ReplicationQueues::pending_verify` (`HashMap`) and `fetch_queue` (`BinaryHeap`) had **no capacity bound** — the source carried the project's own `TODO`. `handle_neighbor_sync_request` documents *"No per-request hint count limit"* and the only gate is `sender_in_rt`, so a single routing-table peer can flood `NeighborSyncRequest` hints (each message capped only by `MAX_REPLICATION_MESSAGE_SIZE` ≈ 10 MiB → ~320k 32-byte hints) and grow these structures unboundedly:

- node memory exhaustion (OOM), and
- a self-amplifying storm of outbound verification/fetch requests onto honest peers (reflective amplification).

## Fix (two layers)

1. **Global memory backstop** — `MAX_PENDING_VERIFY` / `MAX_FETCH_QUEUE` (131_072 each). `add_pending_verify` / `enqueue_fetch` reject once full; callers already treat non-admission as "not added".

2. **Per-source fairness — the actual D1 defence.** Each pending entry is accounted to its `hint_sender` via a `pending_per_sender` counter maintained in lockstep with `pending_verify` (insert / `remove_pending` / `evict_stale`). A single source may hold at most `MAX_PENDING_VERIFY_PER_PEER` (8_192) entries, so a flooding peer exhausts only its own quota and **can never starve honest peers**, whose quotas are independent. `hint_sender` is the cryptographically authenticated connection identity (not a forgeable payload field), so the accounting cannot be evaded.

A global cap *alone* would merely convert the memory DoS into a worse **silent honest-replication starvation DoS** (one ~4 MB message every <30 min permanently rejecting all honest hints). The per-source quota is what actually closes D1. This was caught by adversarial review and is the reason for layer 2.

Hardening: counter decrement uses `saturating_sub` + `debug_assert!`; `get_pending_mut` documents the no-reattribution invariant. A residual (~16 simultaneously-in-routing-table Sybil `PeerId`s to reach the global backstop) is documented as an accepted, tracked follow-up — it degrades only to the bounded memory backstop, never to silent starvation of non-Sybil peers.

## Proof

`tests/poc_d1_bounded_queues.rs` (6 tests) demonstrates the attack is defeated:

- `poc_d1_flooding_peer_cannot_starve_honest_peer` — **the key test**: a 508k-entry single-peer flood is clamped to 8 192 slots and honest hints are still admitted. Verified to **FAIL against a global-cap-only fix** (the weak fix), proving it validates the real defence, not just the memory bound.
- Global-bound, fetch-queue-bound, per-source counter consistency (remove + stale eviction, no underflow/leak), `get_pending_mut` no-desync, and legitimate-traffic / dedup non-regression.

The original unbounded behaviour was also reproduced (pre-fix the queues grew 1:1 with attacker input); the flip was verified by neutering each enforcement layer and observing the corresponding test fail.

## Validation

- 6/6 D1 PoC tests pass.
- All 460 library tests pass (incl. 190 replication tests) — no regression.
- `cfd` clean (fmt + clippy, no warnings/errors).
- Two rounds of independent multi-agent adversarial review; all blockers resolved.

### DIFF
--- a/src/replication/bootstrap.rs
+++ b/src/replication/bootstrap.rs
@@ -122,12 +122,52 @@ pub async fn check_bootstrap_drained(
         return false;
     }
 
+    // Hints capacity-rejected at the pending_verify bounds during bootstrap
+    // must be re-delivered by the source before drain can be claimed;
+    // otherwise we'd silently mark ourselves complete with outstanding work
+    // the source still owes us (codex round-2 BLOCKER).
+    if state.capacity_rejected_outstanding > 0 {
+        debug!(
+            "Bootstrap NOT drained: {} hints capacity-rejected and awaiting re-admission",
+            state.capacity_rejected_outstanding
+        );
+        return false;
+    }
+
     if queues.is_bootstrap_work_empty(&state.pending_keys) {
         state.drained = true;
         info!("Bootstrap drained: all peer requests completed and work queues empty");
         true
     } else {
         false
+    }
+}
+
+/// Record `n` capacity-rejected bootstrap hints.
+///
+/// Hints rejected at the `pending_verify` capacity bounds during bootstrap.
+/// Bootstrap cannot drain while this counter is non-zero; the source's next
+/// periodic hint will replay them once the pending queue has space.
+pub async fn note_capacity_rejected(bootstrap_state: &Arc<RwLock<BootstrapState>>, n: usize) {
+    let mut state = bootstrap_state.write().await;
+    state.capacity_rejected_outstanding = state.capacity_rejected_outstanding.saturating_add(n);
+    debug!(
+        "Bootstrap: +{n} capacity-rejected hints (now {})",
+        state.capacity_rejected_outstanding
+    );
+}
+
+/// Clear the capacity-rejected counter. Called by an admission cycle that
+/// observed zero rejections (the source successfully re-delivered all
+/// previously-overflowed hints).
+pub async fn clear_capacity_rejected(bootstrap_state: &Arc<RwLock<BootstrapState>>) {
+    let mut state = bootstrap_state.write().await;
+    if state.capacity_rejected_outstanding > 0 {
+        debug!(
+            "Bootstrap: clearing {} previously-outstanding capacity-rejected hints",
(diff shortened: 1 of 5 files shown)


dirvine (commented):
## Review: PR #99 — D1 DoS fix

### Architecture & Soundness

The two-layer defence is exactly right:

1. Global memory backstop (MAX_PENDING_VERIFY / MAX_FETCH_QUEUE) prevents unbounded memory growth.
2. Per-source fairness (MAX_PENDING_VERIFY_PER_PEER) prevents a single peer from starving honest peers — the key architectural insight that a global cap alone would convert OOM into worse honest-replication starvation.

The test poc_d1_flooding_peer_cannot_starve_honest_peer is critical: it verifies the actual D1 defence (per-source fairness), not just the memory bound. Well designed.

### Implementation details

- AdmissionResult enum (Admitted / AlreadyPresent / CapacityRejected) is a clean improvement over the previous bool — distinguishes dedup from genuine capacity rejection, essential for bootstrap drain correctness.
- promote_pending_to_fetch (remove+enqueue under the write lock) prevents the silent-drop regression. The RwLock serialization means there is no race.
- set_pending_state replacing get_pending_mut is a good API narrowing — prevents accidental hint_sender re-attribution that would desync the per-source counter.
- release_sender_slot with saturating_sub + debug_assert is defensive and test-friendly.
- Bootstrap drain with capacity_rejected_sources per-peer is thorough — the round-3 regression test validates the clear mechanism explicitly.

### Residual risk

The PR documentation notes the approx 16 Sybil residual. This is acceptable because:
- The global memory backstop still bounds worst-case memory to ~131K entries (~128 MB).
- 16 simultaneous Sybils in the routing table is a substantial operational bar.
- The degradation is to bounded memory, not to silent starvation.
- This is correctly tracked as a follow-up, not left unexamined.

**Verdict:** LGTM — well-structured, well-tested, merge-ready.

jacderida (approved):

PR #101, opened 2026-05-19
fix(payment): bind single-node payment to on-chain recipient + price floor (F2/F5 free storage)

grumbach:
## Summary

Security fix for **F2/F5** (Critical — free storage). The single-node payment path had no DHT/identity binding (unlike the merkle path) and delegated the decision to `SingleNodePayment::verify`, which accepts if *any quote tied at the median price* was paid 3× on-chain. Combined with `validate_local_recipient` only checking `.any()`, and with `PaymentVaultV2.payForQuotes` being **unauthenticated** (the caller sets `{quoteHash, rewardsAddress, amount}` independently and the ANT is sent to the caller-chosen address), an attacker could:

- **underprice**: submit 7 self-signed 1-atto quotes (one carrying our address) and pay a negligible amount; and/or
- **pay-yourself**: register an on-chain `completedPayments` record for a *victim-addressed* quote's hash while sending the ANT to their **own** wallet.

Either way the node stored data while earning nothing — free storage.

## Fix — sound invariant

`verify_evm_payment` no longer calls `SingleNodePayment::verify`. It accepts only if there exists a quote `Q` with **all** of the following, on the **same** quote:

- **(a)** `Q.rewards_address == this node's local_rewards_address`
- **(b)** `Q.price >= price_floor` — `calculate_price(records_stored)/TOL`, wired live in production via a shared `Arc<QuotingMetricsTracker>` (the same tracker the quote generator prices from)
- **(c)** on-chain `completedPayments(Q.hash).amount >= 3 * Q.price`
- **(d)** on-chain `completedPayments(Q.hash).rewardsAddress == first16(local_rewards_address)` — the **trusted on-chain record**, not the attacker-controlled quote object

**(d)** is the decisive check that closes pay-yourself: the contract stores the recipient it actually paid (`bytes16`, matching `getFirst16Bytes`); the verifier now requires that to be **this node**. The `bytes16` truncation leaves 128-bit preimage resistance — an EVM address is `keccak256(pubkey)[12..]`, not an attacker-structured `high16 ++ low4`, so forging a colliding address is ~2^128 work (cryptographically infeasible), not 2^32.

`PaymentVerifierConfig` gains an optional `PriceFloorProvider`; production (`node.rs`, `devnet.rs`) wires it to the metrics tracker. `validate_quote_structure` still enforces exactly `CLOSE_GROUP_SIZE` quotes; the merkle path is untouched; the on-chain check is fail-closed on RPC error; the candidate loop is bounded to ≤7.

## Proof

- `tests/poc_f2_f5_price_floor.rs` — 4 offline deterministic tests: underpriced and pay-yourself-decoy proofs rejected at the pre-RPC recipient+floor gate; flip (without the binding the proof passes the gate and reaches the on-chain step); fair payment not pre-rejected.
- `tests/e2e/f2_f5_pay_yourself.rs` — **live Anvil** with the real `PaymentVaultV2`: the **real pay-yourself `payForQuotes` transaction** (attacker pays 3× to their own wallet for a victim-addressed quote hash) is **rejected**; an honest 3× payment to the node's address is **accepted** (positive control).

The flip was verified by neutering the recipient binding: the e2e pay-yourself attack then returns `PaymentVerified` — i.e. the pre-fix free-storage acceptance, confirming the vulnerability was real and that the on-chain recipient binding closes it.

## Validation

- 4/4 offline PoC tests pass; 2/2 live-Anvil e2e tests pass.
- All 460 library tests pass; e2e target compiles (`--features test-utils`).
- `cfd` clean (fmt + clippy, no warnings/errors).
- **Three** rounds of independent multi-agent adversarial review. Round 1 found the price-floor-only approach insufficient (empty-node + pay-yourself); round 2 confirmed the pay-yourself on-chain gap was still open; the on-chain recipient binding was added; round 3 verified all blockers resolved and that the `bytes16` truncation does not reopen the attack.

### Notes / out of scope

- The `bytes16` recipient truncation is the contract's design (`PaymentVaultV2`); the node-side check matches it exactly and 128-bit resistance is sufficient.
- F1 (empty-node baseline pricing) is subsumed: with the recipient binding closed, paying 3× baseline *to the node* on a fresh node is correct economics, not free storage.

### DIFF
--- a/src/devnet.rs
+++ b/src/devnet.rs
@@ -550,14 +550,21 @@ impl Devnet {
         };
 
         let rewards_address = RewardsAddress::new(DEVNET_REWARDS_ADDRESS);
+        // Shared quoting metrics: generator prices from it, verifier reads the
+        // same live count for its F2/F5 price-floor defence.
+        let metrics_tracker = Arc::new(QuotingMetricsTracker::new(DEVNET_INITIAL_RECORDS));
+        let floor_metrics = Arc::clone(&metrics_tracker);
         let payment_config = PaymentVerifierConfig {
             evm: evm_config,
             cache_capacity: DEVNET_PAYMENT_CACHE_CAPACITY,
             local_rewards_address: rewards_address,
+            price_floor: Some(crate::payment::PriceFloorProvider::new(Arc::new(
+                move || crate::payment::calculate_price(floor_metrics.records_stored()),
+            ))),
         };
         let payment_verifier = PaymentVerifier::new(payment_config);
-        let metrics_tracker = QuotingMetricsTracker::new(DEVNET_INITIAL_RECORDS);
-        let mut quote_generator = QuoteGenerator::new(rewards_address, metrics_tracker);
+        let mut quote_generator =
+            QuoteGenerator::new(rewards_address, Arc::clone(&metrics_tracker));
 
         // Wire ML-DSA-65 signing from the devnet node's identity
         crate::payment::wire_ml_dsa_signer(&mut quote_generator, identity)
--- a/src/node.rs
+++ b/src/node.rs
@@ -381,18 +381,29 @@ impl NodeBuilder {
             }
         };
 
-        // Create payment verifier
+        // Shared quoting metrics: the quote generator prices quotes from it,
+        // and the payment verifier reads the SAME live `records_stored` to
+        // enforce its price floor (F2/F5 defence).
+        let metrics_tracker = Arc::new(QuotingMetricsTracker::new(0));
+
+        // Create payment verifier with a live price floor wired to the same
+        // metrics tracker, so an attacker cannot get this node to accept a
+        // self-signed, far-underpriced single-node proof.
         let evm_network = config.payment.evm_network.clone().into_evm_network();
+        let floor_metrics = Arc::clone(&metrics_tracker);
         let payment_config = PaymentVerifierConfig {
             evm: EvmVerifierConfig {
                 network: evm_network,
             },
             cache_capacity: config.payment.cache_capacity,
             local_rewards_address: rewards_address,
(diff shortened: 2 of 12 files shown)


dirvine (commented):
## Review: PR #101 — F2/F5 free storage fix

### Architecture & Soundness

The fix is structurally rigorous. The four-condition invariant (a-d) on the **same quote** correctly closes both attacker primitives:

- **Underpricing** → (b) price floor kills 1-atto self-signed quotes.
- **Pay-yourself** → (d) on-chain recipient binding kills redirected payments; condition (a) already ensures the quote pays this node, so the attacker cannot use a victim-addressed quote while paying themselves.

Removing the SingleNodePayment::verify delegation (which trusted the attacker-supplied median) and replacing with per-candidate enforcement is the right architectural change.

### Security analysis of bytes16 truncation

I verified the preimage argument: EVM addresses are keccak256(pubkey)[12..]. An attacker cannot freely structure the 16 bytes; they must find a pubkey whose hash suffix matches a specific target. This is ~2^128 preimage work, not ~2^64 birthday-collision work. The analysis in the PR description is correct.

### Implementation

- PriceFloorProvider sharing via Arc is clean; the Into<Arc<...>> bound on QuoteGenerator::new is a nice API touch for test compatibility.
- Fail-closed on RPC error is correct.
- Candidate loop bounded to <=7.
- Startup hydration from storage.current_chunks() — a nice hardening detail.
- Tests: offline PoCs + live Anvil e2e with flip verification = thorough.

### Residual notes

- The price floor tolerance divisor of 4 is a judgement call. Not blocking.

**Verdict:** LGTM — sound, well-tested, merge-ready.

jacderida (approved):

PR #111, opened 2026-05-24
fix: answer Merkle closeness check from local routing table

jacderida:
## Summary

The Merkle pay-yourself defence verified candidate closeness with an iterative Kademlia **network** lookup (`find_closest_nodes_network`) on the PUT-handling hot path. That lookup runs up to `MAX_ITERATIONS` rounds bounded by `CLOSENESS_LOOKUP_TIMEOUT` (240s) and is the dominant term in slow per-chunk store times; its instability (fresh transient peers pulled in on every call) also contributes to the closeness disagreements that cause outright rejections.

- Answer the closeness check from the **local routing table** (`find_closest_nodes_local` — a pure in-memory k-bucket read, no network I/O), matching the precedent already used for the close-group responsibility check (`find_closest_nodes_local_with_self`).
- Fall back to the network lookup **only** when the local table is genuinely too sparse to be authoritative (fewer than `CLOSENESS_LOOKUP_WIDTH` peers near the midpoint).
- The fallback is gated on **local table size, not match outcome**, so a forged pool cannot force the expensive 240s path — an attacker cannot make a victim's local routing table sparse (DoS-safe).
- `check_closeness_match` and the single-flight pass-cache wrapper are **unchanged**.
- Node-side only, **no wire/protocol change** → backwards compatible across a mixed-version fleet.
- The fallback decision is extracted into a pure `const fn` (`closeness_should_fall_back_to_network`) so its `CLOSENESS_LOOKUP_WIDTH` boundary is unit tested without standing up a `P2PNode`.

## Test plan

- [x] `cargo fmt -- --check` — clean
- [x] `cargo clippy --lib --all-features -- -D clippy::panic -D clippy::unwrap_used -D clippy::expect_used` — no warnings
- [x] `cargo test --lib payment::verifier` — 67 passed, 0 failed (incl. new boundary test `closeness_falls_back_to_network_only_below_lookup_width`)
- [x] e2e test target (`--test e2e --features test-utils`) compiles
- [ ] Follow-up (not in this PR): `prod uploaders` run comparing per-chunk `store_durations_ms` p50/p99 against baseline

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4846,7 +4846,7 @@ dependencies = [
 [[package]]
 name = "saorsa-core"
 version = "0.24.5-rc.1"
-source = "git+https://github.com/saorsa-labs/saorsa-core?branch=rc-2026.5.4#7da46fe36b6124ac4cde434bcd18f70d4a8338ab"
+source = "git+https://github.com/saorsa-labs/saorsa-core?branch=rc-2026.5.4#1be73520bb8939d651f6d55ef1b2308810995052"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/src/payment/verifier.rs
+++ b/src/payment/verifier.rs
@@ -1363,18 +1363,23 @@ impl PaymentVerifier {
         let lookup_count = Self::closeness_lookup_count(pool.candidate_nodes.len());
 
         // Fast path: answer from the local routing table. This is a pure
-        // in-memory k-bucket read (`find_closest_nodes_local` returns
-        // `Vec<DHTNode>` with no network I/O and no `Result`), so it is safe to
-        // call from this PUT-handling request handler — unlike
+        // in-memory k-bucket read (`find_closest_nodes_local_by_distance`
+        // returns `Vec<DHTNode>` with no network I/O and no `Result`), so it is
+        // safe to call from this PUT-handling request handler — unlike
         // `find_closest_nodes_network`, which runs an iterative Kademlia lookup
         // (up to MAX_ITERATIONS rounds, bounded by CLOSENESS_LOOKUP_TIMEOUT) and
-        // is the dominant term in slow per-chunk store times. The local table is
-        // already the view this node trusts for the close-group responsibility
-        // check (`find_closest_nodes_local_with_self` above), so using it here
-        // brings the Merkle closeness check in line with that precedent.
-        let mut network_peers = p2p_node
+        // is the dominant term in slow per-chunk store times.
+        //
+        // We use the XOR-only `_by_distance` variant deliberately, NOT the
+        // reachability-reranked `find_closest_nodes_local`: this is a closeness
+        // *verification*, so it must mirror the uploader's pure XOR-distance
+        // view. The reachability re-rank (which the close-group *selection* path
+        // uses) could demote an XOR-close relay-only peer out of the compared
+        // window and falsely reject an honest candidate pool that legitimately
+        // contains that peer. See saorsa-labs/saorsa-core#121.
+        let mut closeness_peers = p2p_node
             .dht_manager()
-            .find_closest_nodes_local(&pool_address.0, lookup_count)
+            .find_closest_nodes_local_by_distance(&pool_address.0, lookup_count)
             .await;
 
         // Sparse-table fallback: only when the local table genuinely cannot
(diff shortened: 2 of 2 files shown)


dirvine (commented):
## Review

**Overall:** The local-table-first approach is a sound performance optimization (avoids 240s network lookup on the hot path). The DoS-safe argument (gate on table size, not match outcome) is correct. No visible CI checks.

### Coordination concern with saorsa-core PR #121

`saoresa-core` PR #121 (``fix/closegroup-reachability``) modifies `find_closest_nodes_local` to re-rank by `(reachability_tier, xor_distance)` before truncation. This PR (#111) calls `find_closest_nodes_local` for the Merkle closeness check.

`check_closeness_match` does a **set-membership check** (`HashSet<PeerId>`) — it verifies that a quorum of the candidate pool peer IDs appear in the returned network peers set. Because `find_closest_nodes_local` now over-fetches 2×, re-ranks by reachability, then truncates, the **set of peers returned** can differ from a pure XOR-distance set:

- The top 32 XOR-closest peers from the routing table might include some relay-only nodes
- Reachability re-ranking promotes direct peers (even if XOR-farther) into the top 32
- A valid candidate pool containing a relay-only peer could be **rejected** because that peer is no longer in the truncated set

The uploader still constructs its candidate pool via network lookup (XOR-only), not via the reachability-aware local path — so the uploader and storer could disagree on which peers belong in the close-group set.

**Recommendation:** Before merging, either:
1. Have this PR use XOR-only ordering for the closeness check (e.g. sort network_peers by XOR distance before the set-membership check), or
2. Coordinate with PR #121 to ensure the ordering semantics are compatible

The ideal ordering for the **closeness check** should be XOR-only, while the **close-group selection for storage** benefits from the reachability re-rank. These are different concerns.

jacderida:
Thanks @dirvine — good catch on the interaction with saorsa-core#121. We're addressing it at the source.

**The concern.** Once saorsa-core#121 re-ranks `find_closest_nodes_local` by `(reachability_tier, xor_distance)` + truncate, the set of peers it returns can differ from the XOR-closest set. Since `check_closeness_match` is a set-membership test, a reachability re-rank could push an XOR-close relay-only peer out of the compared window and falsely reject an honest pool whose candidates the uploader (XOR-only) legitimately picked. (Worth noting the lever is *which peers survive truncation*, not their order — a `HashSet` membership check is order-insensitive — so the fix has to change the selection, not sort after the fact.)

**How we're addressing it.** Rather than special-casing in ant-node, we've added a distance-pure variant in saorsa-core#121: [`find_closest_nodes_local_by_distance`](https://github.com/saorsa-labs/saorsa-core/pull/121) (commit `a04f428`). It returns the routing table's pure XOR-distance order (no over-fetch, no re-rank) for the verification path, while all selection callers keep the reachability re-rank. This matches the "closeness check should be XOR-only, storage selection benefits from the re-rank — different concerns" framing exactly, and keeps the two concerns cleanly separated.

**Why this PR is safe to merge as-is today.** ant-node currently pins published `saorsa-core 0.24.4`, where `find_closest_nodes_local` is already XOR-only (the re-rank is unreleased, on the #121 branch). So this PR's closeness check compares against an XOR set today, exactly like the uploader.

**Follow-up.** When saorsa-core#121 merges, we'll update this PR to reference the saorsa-core RC branch and switch the closeness check (`src/payment/verifier.rs`) from `find_closest_nodes_local` to `find_closest_nodes_local_by_distance` in the same change — so the bump and the call-site switch land together and the hazard never arms.

grumbach (approved):
LGTM. Fast-path correctness preserved, DoS gate is on local-table size (not match outcome), and `check_closeness_match` is byte-for-byte unchanged. Verified against `find_closest_nodes_local` in saorsa-core (`src/dht_network_manager.rs:1865`), which is a pure k-bucket read returning only `is_dht_participant`-validated peers — so the local set is a strict-or-equal-security subset of what the network lookup would have surfaced.

- `src/payment/verifier.rs:1342-1361` — **security**: primary path now calls `find_closest_nodes_local` and feeds the result into the unchanged `check_closeness_match`. The 13/16 threshold and `>= CANDIDATE_CLOSENESS_REQUIRED` sparse-network guard still execute. Passing a forged pool still requires landing 13 Sybil PeerIds in the storer's bucket-disciplined view near the midpoint, which is no easier than against the prior network lookup.
- `src/payment/verifier.rs:1369-1374` — **DoS**: fallback gate is `local < WIDTH`, evaluated before `check_closeness_match`. A forged pool with bad candidates cannot push verification onto the 240s path — it will simply fail closeness against the dense local set. Sparse-table fallback is bounded by the existing per-`pool_hash` single-flight cache and is no worse than the pre-PR baseline where every PUT paid the network lookup.
- `src/payment/verifier.rs:1041` — **correctness**: boundary cases match intent — 0 and `WIDTH-1` fall back; exactly `WIDTH` (32) does not. 32 comfortably exceeds the `>= 13` sparse-network short-circuit in `check_closeness_match`, so the at-boundary case cannot be rejected for the wrong reason.

Minor (non-blocking):

- `src/payment/verifier.rs:1369` — fallback gate is fixed at `CLOSENESS_LOOKUP_WIDTH`, not `lookup_count`. Today `candidate_nodes` is `[_; CANDIDATES_PER_POOL]` so they're equal, but the in-file comment around 1333-1341 already anticipates a future `Vec`-typed pool. Worth scaling the gate with `lookup_count` (e.g. `local < lookup_count.min(WIDTH)`) when that change lands, otherwise the K-too-small failure mode re-opens silently.
- `src/payment/verifier.rs:1376` — `network_peers` now usually holds local-table results. Rename to `closeness_peers` for readability; the `let mut … if fallback { … }` pattern would also read more cleanly as `let peers = if … { network… } else { local… };`.
- `src/payment/verifier.rs:3409-3439` — boundary unit test for the const fn is solid (0 / W-1 / W / W+1). No integration coverage exercises the local-success vs sparse-local-then-network-fallback split — not blocking for this scoped fix but worth a follow-up before the production rollout note in the PR body.

PR #112, opened 2026-05-24
feat: build Linux releases against musl with mimalloc allocator

jacderida:
## Summary

- **Switch both Linux release targets from glibc to musl** (`x86_64-unknown-linux-musl`,
  `aarch64-unknown-linux-musl`, built via `cross`). Published binaries are now musl-static, so they
  run on any Linux distribution — including Alpine and other musl-based systems — while continuing
  to run on glibc hosts (a static binary has no dynamic-linker dependency). Asset filenames are
  **unchanged** (`ant-node-cli-linux-{arm64,x64}.tar.gz`), so the existing auto-upgrade asset
  matcher and `ant node add` keep working with no client-side change.
- **Use `mimalloc` as the global allocator** for `ant-node` and `ant-devnet`. musl's default
  allocator is significantly slower than glibc's under the concurrent allocation churn of a
  DHT/relay node; mimalloc neutralises that, and tends to beat glibc's allocator too.
- **Carries the upgrade-cache hardening from #100** (per-cache-hit ML-DSA re-verification,
  FIFO/pipe rejection at the cache entry path, narrowed copy TOCTOU). These were validated together
  with the musl swap on the same testnets — see evidence below. #100 can be closed in favour of
  this PR.

## Why

On musl distros (e.g. Alpine), `ant node start` previously failed with a misleading
`No such file or directory (os error 2)`: the downloaded glibc `ant-node` requires a glibc dynamic
linker (`/lib/ld-linux-*.so`) that doesn't exist on musl, so `execve` returns `ENOENT` for the
missing loader rather than the binary. Shipping musl-static Linux binaries fixes this for all
Linux distributions at once, with no separate "musl variant" to maintain.

### DEV-03 — auto-upgrade glibc → musl (PASS)

A staged-rollout test confirming existing glibc nodes upgrade cleanly to the musl build.

```
DEV-03 auto-upgrade musl-swap test — PASS

Setup verified before triggering: pre-upgrade binary was glibc-dynamic (interpreter
/lib64/ld-linux-x86-64.so.2), ant-node 0.11.3, identical across the sampled VMs — premise held. Channel was
correctly set to Beta (the gotcha from the DEV-02 run), and the v0.11.14-rc.1 pre-release was already
published, so the staged rollout fired automatically. 157 services (10 node VMs × 15 + 7 bootstraps) rolled
over the 2-hour window with zero errors.

┌───────────────────────────┬─────────────────────────────────────────────────────────────────────────────┐
│ Binary downloaded once    │ ✅ Downloading ant-node binary ×1 on every host (422 MB build)              │
│ per host                  │                                                                             │
├───────────────────────────┼─────────────────────────────────────────────────────────────────────────────┤
│ GitHub release info       │ ✅ 1 detection fetch/host; 36 cache-hits vs 2 fetches confirms lock+cache   │
│ fetched once per host     │ dedup (2nd fetch = TTL refresh or post-upgrade routine re-check, never      │
│                           │ per-service)                                                                │
├───────────────────────────┼─────────────────────────────────────────────────────────────────────────────┤
│ No upgrade errors         │ ✅ 0 ERROR/WARN, 0 crashes, 0 failed units                                  │
├───────────────────────────┼─────────────────────────────────────────────────────────────────────────────┤
│ Upgraded binary is        │ ✅ 157/157 static-pie linked / ldd: statically linked; 0 glibc              │
│ musl-static everywhere    │                                                                             │
├───────────────────────────┼─────────────────────────────────────────────────────────────────────────────┤
│ Cache-hit signature       │ ✅ Reused signature-verified cached archive ×14 per node VM                 │
│ re-verification ran       │                                                                             │
├───────────────────────────┼─────────────────────────────────────────────────────────────────────────────┤
│ Peer ID retained          │ ✅ 157/157 identical pre/post                                               │
├───────────────────────────┼─────────────────────────────────────────────────────────────────────────────┤
│ Restart times evenly      │ ✅ scheduled & actual both uniform, 12 bins of 10–18, not 2–4 bursts        │
│ distributed               │                                                                             │
├───────────────────────────┼─────────────────────────────────────────────────────────────────────────────┤
│ Real process restart each │ ✅ pre-exit msg + NRestarts≥1 + new PID + journald stop/start, all 157      │
│  node                     │                                                                             │
├───────────────────────────┼─────────────────────────────────────────────────────────────────────────────┤
│ No (deleted) inode (hard  │ ✅ 0/157                                                                    │
│ failure)                  │                                                                             │
└───────────────────────────┴─────────────────────────────────────────────────────────────────────────────┘

The key result for the (deleted) concern: all 157 services share /usr/local/bin/ant-node. The coordination
logged exactly Replacing binary ×1 per host, with every other service hitting Binary already upgraded …
skipping replacement after reusing the signature-verified cached archive — so the file is replaced once per
host, no process runs from a stale inode, and /proc/<pid>/exe is clean on every service.
```

### DEV-01 vs DEV-02 — 12 h musl + mimalloc vs glibc reference A/B soak (PASS)

Two identical 450-node fleets under identical load; only the binary differs.

```
DEV-01 (musl + mimalloc) vs DEV-02 (glibc reference) — 12h A/B Report

Setup: two identical 450-node fleets (30 VMs × 15 nodes, DO+Vultr, 9 NAT-simulated), identical client load,
both ant-node 0.11.4. Only the binary differs — DEV-01 musl-static + mimalloc (BuildID 884f505), DEV-02
official glibc release (4e07325). Soak window ~12.2 h (17:05 → 05:17 UTC).

Crash / stability — PASS, both arms

- 906 services (453/arm) active/running, NRestarts=0. Zero crashes, restarts, or failed units on either arm.
- Zero OOM-killer events, zero SIGSEGV/SIGABRT/panic/stack-overflow lines across all 66 VMs.
- musl-specific risks (128 KiB thread stack, stripped resolver) did not materialise — 0 DNS errors fleet-wide.

Memory — the headline

Total RSS tracks within 1–3% between arms the entire window:

┌───────────────────┬──────┬──────┬────────┐
│                   │ t+1h │ t+6h │ t+12h  │
├───────────────────┼──────┼──────┼────────┤
│ DEV-01 musl mean  │ 230  │ 567  │ 979 MB │
├───────────────────┼──────┼──────┼────────┤
│ DEV-02 glibc mean │ 202  │ 565  │ 982 MB │
└───────────────────┴──────┴──────┴────────┘

Both grow ~5×, but that growth is memory-mapped LMDB chunk store, not heap. Per-node breakdown at 12h:

┌─────────────────┬────────────────────────────────────┬──────────────────────────────────┐
│       Arm       │ RssAnon (heap, allocator-governed) │ RssFile (LMDB mmap, reclaimable) │
├─────────────────┼────────────────────────────────────┼──────────────────────────────────┤
│ musl + mimalloc │ 250 MB/node                        │ 726 MB/node                      │
├─────────────────┼────────────────────────────────────┼──────────────────────────────────┤
│ glibc + system  │ 231 MB/node                        │ 750 MB/node                      │
└─────────────────┴────────────────────────────────────┴──────────────────────────────────┘

chunks.mdb is 4.6 GB (mmap'd); free shows ~10.8 GB available per VM — no memory pressure. The RSS climb is
chunk accumulation under continuous upload load (file-backed page cache, reclaimable), allocator-independent,
and identical on both arms.

- Allocator verdict: mimalloc's heap (RssAnon) is ~250 vs glibc's ~231 MB/node — a modest +8% (~19 MB/node),
stable, no leak. Consistent with mimalloc's known slight RSS-for-speed tradeoff.
- Worst-node total-RSS tail ran a bit higher on musl (peak ~3.8 vs ~2.8 GB), but that's dominated by random
chunk-distribution variance in the mmap component, not the allocator.

CPU — neutral

Mean per-node CPU: musl 25.7% vs glibc 25.1%.

P2P health — PASS, both arms

Uploads completing ok (10/25/500 MB) on both; 1300–3200 accepted connections/node; 0 DNS errors; negligible
ERROR rate (0–3 lines/node/12h). Comparable across arms.

Overall verdict: PASS — musl + mimalloc is a safe swap

Over 12 h under identical load, the musl + mimalloc build is stability-, memory-, and CPU-neutral versus the
glibc reference release: zero crashes, ~8% (~19 MB/node) more anonymous heap, identical total-RSS trajectory,
identical CPU. No leak, no musl regression. The large RSS growth seen on both arms is the LMDB chunk store and
is not allocator-related (it affects the current reference release equally).
```

## Test plan

- [x] Auto-upgrade glibc → musl on a 157-service testnet (DEV-03): single download per host, musl-static
  binary on every node, peer IDs retained, even rollout, real process restarts, no stale-inode execution.
- [x] 12 h A/B soak musl+mimalloc vs glibc reference (DEV-01/DEV-02): no crashes, memory/CPU neutral, no
  musl-specific failures (DNS, thread-stack), no allocator leak.
- [ ] Confirm release workflow produces the expected musl assets (filenames unchanged) on the next tagged
  build.

Closes #100

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -82,13 +82,16 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          # Use ubuntu-22.04 for GLIBC 2.35 compatibility with server deployments
-          - target: x86_64-unknown-linux-gnu
+          # Linux builds use musl for portability across glibc and musl distros
+          # (e.g. Alpine). Built via `cross` so the musl toolchain is provided
+          # by the cross-rs container image.
+          - target: x86_64-unknown-linux-musl
             os: ubuntu-22.04
             binary: ant-node
             archive: tar.gz
+            cross: true
             friendly_name: linux-x64
-          - target: aarch64-unknown-linux-gnu
+          - target: aarch64-unknown-linux-musl
             os: ubuntu-22.04
             binary: ant-node
             archive: tar.gz
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -825,7 +825,9 @@ dependencies = [
  "futures",
  "heed",
  "hex",
+ "libc",
  "lru",
+ "mimalloc",
  "objc2",
  "objc2-foundation",
  "page_size",
@@ -3467,6 +3469,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b6d2cec3eae94f9f509c767b45932f1ada8350c4bdb85af2fcab4a3c14807981"
 
 [[package]]
+name = "libmimalloc-sys"
+version = "0.1.47"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2d1eacfa31c33ec25e873c136ba5669f00f9866d0688bea7be4d3f7e43067df6"
+dependencies = [
+ "cc",
+]
+
+[[package]]
 name = "libredox"
 version = "0.1.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3582,6 +3593,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "488016bfae457b036d996092f6cb448677611ce4449e970ceaf42695203f218a"
 dependencies = [
  "autocfg",
+]
+
+[[package]]
+name = "mimalloc"
+version = "0.1.50"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b3627c4272df786b9260cabaa46aec1d59c93ede723d4c3ef646c503816b0640"
+dependencies = [
+ "libmimalloc-sys",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -106,6 +106,9 @@ page_size = "0.6"
 # Protocol serialization
 postcard = { version = "1.1.3", features = ["use-std"] }
 
+[target.'cfg(unix)'.dependencies]
+libc = "0.2"
+
 [target.'cfg(windows)'.dependencies]
 self-replace = "1"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,12 @@ name = "ant-devnet"
 path = "src/bin/ant-devnet/main.rs"
 
 [dependencies]
+# Global allocator. musl's default malloc is significantly slower than
+# glibc's under concurrent allocation churn, which matches the node's
+# steady-state workload. mimalloc neutralises that regression for the
+# musl Linux builds (and tends to beat glibc's allocator too).
+mimalloc = "0.1"
+
 # Wire protocol — the single version-pin shared with ant-client.
 # Bumping ant-protocol's `evmlib`/`saorsa-core`/`saorsa-pqc` pins ripples
 # through here automatically; we keep a direct saorsa-core dep for
--- a/src/bin/ant-devnet/main.rs
+++ b/src/bin/ant-devnet/main.rs
@@ -2,6 +2,9 @@
 
 #![cfg_attr(not(feature = "logging"), allow(unused_variables))]
 
+#[global_allocator]
+static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
+
 mod cli;
 
 use ant_node::devnet::{Devnet, DevnetConfig, DevnetEvmInfo, DevnetManifest};
--- a/src/bin/ant-node/main.rs
+++ b/src/bin/ant-node/main.rs
@@ -2,6 +2,9 @@
 
 #![cfg_attr(not(feature = "logging"), allow(unused_variables))]
 
+#[global_allocator]
+static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
+
 mod cli;
 mod platform;
 
--- a/src/upgrade/apply.rs
+++ b/src/upgrade/apply.rs
@@ -20,7 +20,7 @@ use std::path::{Path, PathBuf};
 use tar::Archive;
 
 /// Maximum allowed upgrade archive size (200 MiB).
-const MAX_ARCHIVE_SIZE_BYTES: usize = 200 * 1024 * 1024;
+pub(super) const MAX_ARCHIVE_SIZE_BYTES: usize = 200 * 1024 * 1024;
 
 /// Exit code that signals the service manager to restart the process.
 ///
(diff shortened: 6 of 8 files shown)


dirvine (approved):
## Review: PR #112 — musl builds + mimalloc + upgrade cache hardening

This is a high-quality PR. The changes are well-scoped, thoroughly documented in the code, and validated with real-world testnet evidence (12h A/B soak on 2×450-node fleets + 157-node auto-upgrade test). I have no blocking concerns. Below is a structured review.

### 1. musl build switch (CI) ✅
Clean change. Switching both `x86_64-unknown-linux-gnu` and `aarch64-unknown-linux-gnu` to their `-musl` counterparts, using `cross: true` for the musl toolchain container. Asset filenames are unchanged so the auto-upgrade asset matcher works without client-side changes. The DEV-03 test confirmed glibc→musl auto-upgrade on 157 services with zero errors, peer IDs retained, and no stale-inode execution.

### 2. mimalloc global allocator ✅
Correct choice — musl's default malloc under concurrent allocation churn is the known weak point, and mimalloc neutralises it (tending to beat glibc too). The 12h A/B soak confirmed: zero crashes, ~8% more RssAnon (250 vs 231 MB/node, ~19 MB difference), identical total-RSS trajectory. No leak, no CPU regression (25.7% vs 25.1%).

**Minor note (non-blocking):** The `Cargo.toml` spec `mimalloc = "0.1"` is a semver range (`^0.1`). The `Cargo.lock` pins 0.1.50, so deterministic builds are fine. If you want to guard against `cargo update` pulling an untested minor bump, consider adding a precise pin comment in `Cargo.toml` referencing the soak-tested version.

### 3. Upgrade cache hardening — security rewrite ✅✅
This is the most important change. The previous design cached the *extracted binary* and only SHA-256-checked it on cache hits — meaning a cache-dir write attacker could replace the binary with a matching SHA-256 hash and get it executed without any signature verification (persistent RCE). The new design:

- **Caches signed archives** (not extracted binaries), with detached ML-DSA-65 signatures
- **Re-verifies the ML-DSA signature on every cache hit** — the SHA-256 metadata is a corruption pre-check only
- **Closes the verify-vs-extract TOCTOU** by copying the archive into a process-private 0700 staging directory *before* verification, so extraction reads exactly the bytes that were verified
- **Defends against FIFO/pipe/symlink attacks** via `open_regular_capped()`: pre-check + O_NONBLOCK (via `libc::O_NONBLOCK` for correct per-arch constant) + post-open `is_file()` check
- **Sets 0700 on the shared cache directory** (defence in depth)
- **Creates upgrade temp dirs with 0700** via `tempfile::Builder::permissions`

The test coverage for the cache is excellent: store-and-retrieve, tampered archive rejection, private-copy immunity to post-verify swap, missing signature, missing meta, oversize rejection, wrong-size signature, symlink rejection, FIFO non-hang, oversized meta rejection, meta-symlink-to-special rejection. The FIFO test correctly uses `libc::mkfifo` (not a shell-out), so CI can't silently skip it.

### 4. Residual concerns (documented, acceptable)

**Cache version binding:** The PR's own docs acknowledge that `SIGNING_CONTEXT = "ant-node-release-v1"` is constant across versions, so an attacker with cache-dir write could swap an old signed archive under a new version's key (forced downgrade/wrong-arch crash loop, not RCE). This is out of scope and is acceptable given:
- The 0700 cache dir permissions shrink the attack surface significantly
- An attacker with same-UID write can already replace the running binary directly

**open_regular_capped TOCTOU:** Between the `fs::metadata()` pre-check and `opts.open()`, a cache-dir writer could swap the file to something that passes the size/file-type check but contains different bytes. The security model handles this correctly: the open fd is used to stream into the private staging dir, and the *private copy* is what gets ML-DSA-verified. Even with a swapped-open file, the signature won't match the new bytes → cache hit rejected → fresh verified download.

### Summary

**Verdict:** Approve. This is a well-structured, well-tested, properly documented PR. The musl+mimalloc swap is validated by production-grade soak testing, and the upgrade cache hardening closes a genuine RCE vulnerability (the old SHA-256-only cache-hit gate). The code quality is high, with thorough comments explaining the threat model, design choices, and residual risks throughout.

PR #113, opened 2026-05-26
feat(replication): v12 storage-bound audit — design-complete

grumbach:
## Summary

v12 storage-bound audit (`notes/security-findings-2026-05-22/proposal-gossip-audit-v12.md`), **end-to-end and design-complete**. Closes Finding 1 (lazy node defeats audit via on-demand fetch) and Finding 2 (bootstrap-claim shield).

This is a **breaking wire change** (StorageCommitment now embeds the sender's ML-DSA-65 public key; old peers will fail to decode). Auto-upgrade handles the version break.

## Design checklist (v10 §§1-13 + v12 §5)

| § | Item | Status |
|---|---|---|
| §1 | Merkle tree + StorageCommitment + piggyback on NeighborSync | ✅ |
| §2 step 1 | sender_peer_id == authenticated transport peer | ✅ |
| §2 step 2 | sender in routing table | ✅ |
| §2 step 3 | rate limit: 1 sig verify per peer per 60s (atomic, race-free) | ✅ |
| §2 step 4 | ML-DSA-65 signature verify (against embedded pubkey) | ✅ |
| §2 step 5 | sticky `commitment_capable` flag | ✅ |
| §3 | bootstrap-claim shield: capable peer with no commitment → no audit, no credit | ✅ |
| §4 | Responder: retained-slot rotation (4 slots × 1h = 4h coverage) | ✅ |
| §5 (v12) | Conditional `UnknownCommitmentHash` invalidation | ✅ |
| §6 | RecentProvers with 40min TTL + hash-bound credit + per-key cap 16 | ✅ |
| §6 | `is_credited_holder` threaded into quorum (`evaluate_key_evidence_with_holder_check`) | ✅ |
| §7-8 | Lazy-node + replay attack closure | ✅ (PoC tests) |
| §10 | Domain separation constants | ✅ |
| §11 | DoS mitigations (rate limit, RT gate, cap, structural bounds) | ✅ |
| §12 | Backwards compat | ⚠ Wire-break accepted (auto-upgrade handles) |
| §13 | All implementation checklist items | ✅ |

## What this PR ships

- **Foundation** (4 modules + struct): `commitment.rs`, `commitment_state.rs` (incl. new `PeerCommitmentRecord`), `commitment_audit.rs`, `recent_provers.rs`.
- **Responder side** fully wired: 1h rotation, immediate startup build + gossip trigger, 4-slot retention, streaming per-key response build.
- **Auditor side** fully wired: pinned challenges, streaming verifier, conditional invalidation, holder-credit recording.
- **Gossip ingest**: 5-gate verification (RT, peer-id binding, pubkey-binding, signature, cache cap) with atomic per-peer 60s sig-verify rate limit. Auto-cleanup on `PeerRemoved`.
- **Quorum integration**: `evaluate_key_evidence_with_holder_check` downgrades uncredited Present claims to Unresolved. Wired into `run_verification_cycle`.
- **Bandwidth**: ~5.3 KiB per commitment, ~85 KiB/h per node at default cadence. Documented.

## Security review

15 rounds of codex review + David's PR review. All findings addressed. Selected highlights:

| Round | Severity | Issue | Fix |
|---|---|---|---|
| 5 | BLOCKER | Auditor preloaded all chunks (multi-GB DoS) | Streaming verifier (one chunk peak) |
| 5 | BLOCKER | Sender pubkey not bound to peer-id | Gate 2c: `BLAKE3(pubkey) == sender_peer_id` |
| 6 | BLOCKER | Substring match on rejection allowed audit bypass | Strict exact-match + pinned-challenge gate |
| 7 | MAJOR | Off-RT sybils could churn the cache | RT-membership gate at ingest |
| 8 | MAJOR | Dropping pin on "unknown" reopened on-demand-fetch attack | Initially kept pin; revisited in round-13 with §3 shield in place |
| 9 | MAJOR | Plain Digests accepted on pinned challenges | Reject as MalformedResponse when pinned |
| 9 | MAJOR | Responder still preloaded chunks | Streaming responder via precheck + per-key helpers |
| 10 | MAJOR | Rotation cadence (10 min) vs gossip cooldown (1 h) | Rotation → 1 h |
| 11 | MAJOR | 2-slot retention insufficient for gossip lag | Retention → 4 slots |
| 11 | MAJOR | "key not in commitment" wrongly penalised | Treat as benign staleness |
| 12 | MAJOR | "missing bytes for committed key" indistinguishable | Differentiated rejection strings |
| 12 | MAJOR | Pre-restart pins unrecoverable | Immediate post-startup gossip trigger |
| David | MAJOR | RecentProvers lacked TTL eviction | TTL + periodic sweep |
| 13 | MEDIUM | Rate limit only fired for peers with cached commitments | Separate `sig_verify_attempts` map stamped on every attempt |
| 13 | MEDIUM | §6 TTL was 4h vs design's 40min | Corrected to 40m |
| 14 | MEDIUM | Rate-limit check + stamp under separate locks (race) | Atomic single write-lock critical section |
| Final (15) | — | No findings | — |

## Threat coverage (20 PoC tests in `tests/poc_commitment_audit_attacks.rs`)

- Lazy node passes audit with on-demand fetch under original pin (admitted v12 economic-not-cryptographic limit)
- Fresh-commitment substitution rejected by pin
- Cross-peer commitment substitution rejected by sender-id binding (gate 2a)
- Throwaway-key substitution rejected by peer-id ↔ pubkey binding (gate 2c)
- All structural failure modes rejected with correct `AuditVerifyError` variant
- Audit response replay blocked by per-challenge nonce
- Responder drops old commitments past retention window
- Silent peer earns no credit
- Rotated commitment drops holder credit
- Overclaim via partial commitment yields no holder credit (lemma + end-to-end)
- `commitment_capable` flag is sticky across eviction
- `capable_but_no_commitment` constructor preserves capability

Plus 60 unit tests across the four new modules (including 3 new quorum tests for the holder-credit predicate).

## Tests + CI

- **557 lib tests** pass.
- **20 PoC tests** pass.
- `cargo fmt --check` clean.
- `cargo clippy --all-targets --all-features` warning-only; the deny gates (`-D clippy::panic -D clippy::unwrap_used -D clippy::expect_used`) pass.

## Test plan

- [x] `cargo test --lib`: 557 pass
- [x] `cargo test --test poc_commitment_audit_attacks`: 20 pass
- [x] cfd clean
- [x] codex review (15 rounds; no findings on round 15)
- [x] David's PR architectural review
- [ ] Single-node devnet smoke test (stage 0 of testnet plan)
- [ ] 4-9 node testnet (stage 1: gossip + responder + auditor + holder-credit enabled)

## Out-of-scope (future PRs)

- **Disk persistence of commitments**: ML-DSA signatures are randomized, so the commitment hash changes across restart. Mitigated by immediate post-startup gossip trigger (recovery window measured in seconds). Disk persistence would let pre-restart pins survive.
- **Event-driven rotation on fresh-write**: a just-replicated key is currently auditable only after the next 1h rotation. The auditor treats this as benign staleness (no trust penalty). Event-driven rebuild would close the gap but adds wiring complexity for marginal gain.
- **Concurrent-ingest regression test** for the sig-verify rate limit: only inspected for race-correctness; no dedicated test.

---

## Update — large-testnet hardening (folded in, rebased onto latest main)

Validated on a 400-node DigitalOcean testnet (5 regions + local Anvil EVM). Two product fixes are folded into this PR on top of the v12 audit:

### 1. Audit-timeout strike grace (no honest-node false positives)

The audit-response timeout is an economic deterrent calibrated for residential bandwidth, not a hard bound — so a single slow response is routine for an honest node under transient load. Penalizing on the first timeout false-positived honest nodes on the testnet.

A `Timeout` now records a per-peer strike and only emits `ApplicationFailure` after `AUDIT_TIMEOUT_STRIKE_THRESHOLD` (3) **consecutive** timeouts; any successful audit resets the counter. This removes honest false positives while preserving the deterrent against a peer that does not store the data and must fetch it at audit time (it times out on *every* audit, so its strikes never reset and it crosses the threshold). The per-challenge window is not widened.

Confirmed storage-integrity failures (`DigestMismatch` / `KeyAbsent` / `Rejected` / `MalformedResponse`) still penalize **immediately** and revoke holder credit — unchanged. The decision is a pure function (`plan_failed_audit` / `decide_audit_failure_action`), so the strike-selection glue is unit-testable without a live `P2PNode`; tests are written to fail if the feature is reverted (mutation-verified).

### 2. Deletion-aware quoting

The quote price is resynced from the authoritative LMDB record count (`current_chunks()`) at quote time, on both quote entry points, so a node that deletes data it held prices on what it **actually holds** instead of a monotonic store counter. (This is complementary to the verifier-side record count on `main`, which governs payment verification rather than the advertised quote price.) Peer-verifiable quote-quantity audit — binding the signed quote to the audited commitment `key_count` — is a tracked follow-up, out of scope here; this PR's scope remains: a node must prove it holds what it currently claims to hold.

### Testnet result (summary)

- **Relay attackers** (don't store; fetch at audit time): detected — they fail every audit and accumulate strikes to penalty/eviction. Zero passes.
- **chunk-deleter / lazy** (drop committed data): caught via the storage-integrity gate when challenged on a dropped-but-still-committed key.
- **Honest nodes**: no storage-integrity false accusations. Residual honest timeout penalties were an artifact of the accelerated test cadence (and were cut sharply by the strike grace); they shrink further at production cadence.
- Workload: 100% upload / 100% download under attack.

The testnet harness, the adversary/bad-node modes, the local-EVM host, the structured event log, and the accelerated-cadence env knobs used to produce these results live on a separate `grumbach/testnet-relay-10pct` branch (not for merge) so this PR stays product-only — none of that instrumentation is in the default/release build.


dirvine:
## Architectural Review — PR #113: Storage-Bound Audit (v12)

Anselme, this is a substantial and well-structured piece of work. I've reviewed the full diff (~7k lines), the design document (v12), the PoC tests, and the CI results. Here's my assessment.

---

### Design: Correct

The v12 design correctly closes **Finding 1** (lazy-node audit defeat) and **Finding 2** (bootstrap-claim shield). The key mechanism — pinning audit challenges to a gossiped-to-you commitment hash — means a lazy node must have had the bytes at gossip time to build a valid Merkle inclusion path. The conditional invalidation on `UnknownCommitmentHash` (the v10→v11→v12 fix) correctly handles the honest-rotation race.

The staged rollout (responder first, auditor follow-up) is safe because `expected_commitment_hash: Option<[u8; 32]>` defaults to `None`, keeping the legacy plain-digest path intact.

---

### CI: Must Fix Before Merge

**The e2e tests don't compile.** This is a blocker:

| File | Error | Location |
|---|---|---|
| `tests/e2e/replication.rs` | `AuditChallenge` missing field `expected_commitment_hash` | Lines 392, 442, 864 |
| `tests/e2e/replication.rs` | `NeighborSyncRequest` missing field `commitment` | Lines 804, 1253, 1400 |
| `tests/e2e/replication.rs` | `ReplicationEngine::new` missing `NodeIdentity` argument | (line ~?) |
| `tests/e2e/replication.rs` | `NeighborSyncResponse` missing field `commitment` | Line ? |

These all need `None` for the optional fields and the new identity arg. Also:

- **`cargo fmt --check` fails** in `audit.rs` (lines 570, 612) and `mod.rs` (lines 1002, 1150).
- **Clippy**: `first doc comment paragraph is too long` at `src/replication/audit.rs:557` — the docstring on `handle_audit_challenge_with_commitment` needs a blank line after the first sentence.
- **Documentation build** also fails (likely the same doc comment issue or a similar one).

---

### Security

**1. `ingest_peer_commitment` stores without signature verification (phase-3.5 TODO)**

The comment explains the rationale — the audit-time verifier catches forged commitments — but this creates a DoS surface: an attacker can inject fake commitments for honest peers into `last_commitment_by_peer`. Once the auditor is wired, an honest node whose commitment hash was poisoned will fail commitment-bound audits (signature won't match). Consider adding signature verification on ingest *now* behind a feature flag, or at least rate-limiting per-source peer.

**2. `last_commitment_by_peer: HashMap<PeerId, StorageCommitment>` has no eviction policy**

If a peer leaves the routing table, their commitment lingers forever. This could grow unbounded over churn. Consider either:
- A TTL on entries (e.g. 2 × `COMMITMENT_ROTATION_INTERVAL`)
- Hook into the RT eviction path to `remove()` the peer

**3. RecentProvers: no TTL eviction**

The PR mentions `TTL-evicted` in the v10 design doc but the implementation only has LRU-by-cap eviction. Without a time-based expiry, entries linger until the per-key cap (16) is hit. If a key is rarely audited, stale entries may persist for the auditor's lifetime. Add a TTL check in `is_credited_holder` or a background sweep.

---

### Code Quality

Generally excellent. Particular strengths:

- **Clean module boundaries**: `commitment.rs` (wire + Merkle + sig), `commitment_state.rs` (rotation + lookup), `commitment_audit.rs` (pure verifier), `recent_provers.rs` (cache). Easy to review individually.
- **Cheapest-first gates** in `verify_commitment_bound_response`: structural checks before any hashing, hash pin before sig verify, sig verify before per-key loop.
- **Domain separation constants** cover every hash context (commitment sig, commitment hash, leaf, internal node).
- **Backward-compat wire tests** (`old_decoder_tolerates_new_*`) pin the postcard-trailing-bytes invariant.
- **17 PoC threat-model tests** document each attack path and its closure — excellent for future maintainers.

Minor style: `commitment_hash` returns `Option<[u8; 32]>` when it could return a result. The comment says "callers may safely treat None as malformed" — a `Result<_, Infallible>` or a panic-safe `expect` in practice is fine but the `Option` forces a `.ok_or_else` on every construction site.

---

### Recommendations

1. **Fix the CI failures** (e2e compile, fmt, clippy, docs) before merge — these are trivial.
2. **Add signature verification on ingest** now (phase 3, not phase 3.5). The public key is derivable from the peer ID in `PeerInfo` in the routing table; the code already has the path to look it up.
3. **Add eviction to `last_commitment_by_peer`** — at minimum hook into RT removal.
4. **Add TTL eviction to `RecentProvers`** to complement the LRU-by-cap.
5. **Document the bandwidth impact** of the ~3 KB ML-DSA-65 piggyback on `NeighborSync` (every 10 min × |RT| peers). It's negligible, but worth calling out.

---

**Summary:** Strong design, clean implementation, well-tested foundation. The CI issues need fixing before merge. With the ingest-verification and eviction gaps closed, this is ready for review and stage-0 deployment.

grumbach:
Replaced by #128

PR #120, opened 2026-05-28
fix(payment): replace clock-skew tolerance with storage-delta freshness

dirvine:
## Summary

Replaces payment quote freshness checks based on wall-clock timestamps with storage-delta validation derived from the signed quote price. This removes the dependency on client/node clock synchronisation that was causing upload failures for users who only succeeded after setting macOS/Windows to UTC.

## Problem

- `QUOTE_FUTURE_SKEW_TOLERANCE_SECS = 300` could reject otherwise valid upload quotes when local clocks/timezones disagreed.
- Discord reports that setting machines to UTC makes the issue disappear, strongly pointing at wall-clock quote validation.
- Testnets are likely mostly UTC, masking the issue in test coverage.

## Solution

Instead of comparing quote timestamps, the verifier now:
1. Derives the quoting node's record count from the signed `price` using the existing quadratic pricing formula.
2. Compares that inferred count with this node's current stored-record count.
3. Accepts the quote if the delta is within tolerance: `max(5 records, 5% of quoted records)`.

Quote timestamps are deliberately not used for single-node quote freshness in this path. There is no timestamp fallback.

## Changes

- `src/payment/pricing.rs`: adds `derive_records_stored_from_price(price)` plus round-trip tests.
- `src/payment/verifier.rs`: replaces timestamp freshness validation with storage-delta validation.
- `src/storage/handler.rs`: updates the verifier's `records_stored` counter whenever a new chunk is stored.
- Tests updated so old/future timestamps are no longer rejected by timestamp age/skew.

## Verification

- `cargo test -p ant-node payment::pricing --lib --no-fail-fast` — 15/15 passed.
- `cargo test -p ant-node payment::verifier --lib --no-fail-fast` — 57/57 passed.
- `cargo test -p ant-node --lib --no-fail-fast` — 495/495 passed.
- GitHub CI previously passed across Ubuntu, macOS, and Windows for this head.

## Checklist

- [x] No protocol/wire-format changes.
- [x] No `records_stored` field added to `PaymentQuote`.
- [x] No timestamp fallback restored.
- [x] Verifier derives freshness from signed `price` only.

### DIFF
--- a/src/payment/pricing.rs
+++ b/src/payment/pricing.rs
@@ -42,6 +42,27 @@ const PRICE_BASELINE_WEI: u128 = 3_906_250_000_000_000;
 /// `0.03515625 ANT × 10¹⁸ wei/ANT = 35_156_250_000_000_000 wei`.
 const PRICE_COEFFICIENT_WEI: u128 = 35_156_250_000_000_000;
 
+/// Price increment per squared record after simplifying `PRICE_COEFFICIENT_WEI / DIVISOR_SQUARED`.
+const PRICE_PER_RECORD_SQUARED_WEI: u128 = PRICE_COEFFICIENT_WEI / DIVISOR_SQUARED;
+
+/// Derive the quoted record count from a quote price.
+///
+/// This is the inverse of [`calculate_price`] and is used to validate quote
+/// freshness without relying on wall-clock timestamps. It intentionally floors
+/// to the nearest integer record count, matching the existing storage-delta
+/// tolerance behaviour.
+#[must_use]
+pub fn derive_records_stored_from_price(price: Amount) -> u64 {
+    let baseline = Amount::from(PRICE_BASELINE_WEI);
+    if price <= baseline {
+        return 0;
+    }
+
+    let excess = price - baseline;
+    let n_squared = excess / Amount::from(PRICE_PER_RECORD_SQUARED_WEI);
+    n_squared.root(2).to::<u64>()
+}
+
 /// Calculate storage price in wei from the number of close records stored.
 ///
 /// Formula: `price_wei = BASELINE + n² × K / D²`
@@ -185,5 +206,22 @@ mod tests {
         assert_eq!(price, expected_price(100));
         assert!(price < Amount::from(WEI_PER_TOKEN)); // well below 1 ANT
         assert!(price > Amount::from(PRICE_BASELINE_WEI)); // strictly above baseline
+    }
+
+    #[test]
+    fn test_derive_records_stored_from_price_round_trips() {
+        for records in [0usize, 1, 5, 100, 6_000, 12_000, 60_000] {
+            let price = calculate_price(records);
+            assert_eq!(derive_records_stored_from_price(price), records as u64);
+        }
+    }
+
+    #[test]
+    fn test_derive_records_stored_from_baseline_or_lower_is_zero() {
+        assert_eq!(derive_records_stored_from_price(Amount::ZERO), 0);
+        assert_eq!(
+            derive_records_stored_from_price(Amount::from(PRICE_BASELINE_WEI)),
+            0
+        );
     }
 }
--- a/src/payment/verifier.rs
+++ b/src/payment/verifier.rs
@@ -7,6 +7,7 @@ use crate::ant_protocol::CLOSE_GROUP_SIZE;
 use crate::error::{Error, Result};
 use crate::logging::{debug, info};
 use crate::payment::cache::{CacheStats, VerifiedCache, XorName};
+use crate::payment::pricing::derive_records_stored_from_price;
 use crate::payment::proof::{
     deserialize_merkle_proof, deserialize_proof, detect_proof_type, ProofType,
 };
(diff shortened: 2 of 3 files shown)


dirvine:
Updated PR #120 with option 1: storage-delta-only quote freshness, no timestamp fallback, and no canonical-price hardening in this PR. The branch is rebased onto current main and the PR is now mergeable.\n\nLocal verification: `cargo test -p ant-node --lib --no-fail-fast` passes: 495/495.

dirvine:
Addressed review comments on current head:

- Removed the old `records_stored == 0` compatibility/timestamp-fallback path; verifier now derives quoted records from signed `price` and never uses quote timestamp for single-node freshness.
- `record_payment()` / `record_store(data_type)` comments are stale from the earlier version; current `QuoteGenerator::record_store()` keeps the original no-arg API and compiles.
- Wired verifier `records_stored` updates in production storage path: `src/storage/handler.rs` updates it immediately after `quote_generator.record_store()`.
- Updated PR body to match current implementation: no timestamp fallback, no wire-format change, no `records_stored` field in `PaymentQuote`.
- Existing single-node security validations are still present; this PR only swaps timestamp freshness for storage-delta freshness.

Retested locally:
- `cargo test -p ant-node payment::pricing --lib --no-fail-fast` — 15/15 passed
- `cargo test -p ant-node payment::verifier --lib --no-fail-fast` — 57/57 passed
- `cargo test -p ant-node --lib --no-fail-fast` — 495/495 passed

mickvandijke (approved):

grumbach:
Reviewed strictly against this PR's diff (3 files: `src/payment/pricing.rs`, `src/payment/verifier.rs`, `src/storage/handler.rs`), codex xhigh effort. The wall-clock removal itself is clean, but the replacement freshness heuristic has two real issues I'd want addressed before this lands.

## Blockers

1. **Pre-signature panic on oversized prices.** `src/payment/pricing.rs:55` — `derive_records_stored_from_price()` ends with `n_squared.root(2).to::<u64>()`. In `ruint`, `to()` is `expect`-based and panics on overflow. The function is called from `src/payment/verifier.rs:475` (`validate_quote_freshness`) **before** `validate_peer_bindings()` and `verify_quote_signature()` — so a deserialized proof carrying a huge `quote.price` can crash the verifier before any authenticity check runs. Suggested fix: use `try_to::<u64>()` (or a saturating conversion) and return `Error::Payment` for out-of-range prices.

2. **Cross-peer counter comparison + counter never seeded / updated on all paths.** `src/payment/verifier.rs:574` compares each quote's `derive_records_stored_from_price(quote.price)` against `self.records_stored`, but the price encodes the *quoting* node's count, not the verifier's. Combined with how `records_stored` is maintained, this is fragile:
   - Counter starts at 0 on construction (`src/payment/verifier.rs:262`) and is never seeded from disk in `src/node.rs:394`.
   - It's only ever updated from one call site: `src/storage/handler.rs:263` on a successful client PUT path.
   - Replication stores at `src/replication/mod.rs:1209` and `:2438` don't call `set_records_stored`.
   - Prune deletes at `src/replication/pruning.rs:496` don't decrement it.

   Net effect: a freshly started or restarted verifier sitting at 0 will reject any non-trivial fresh quote (a quote pricing >5 records will exceed the `max(5, 5%)` tolerance), and long-running nodes drift away from reality. Suggested fix: either seed from `LmdbStorage` on startup and hook all store/delete paths (incl. replication and prune), or don't compare remote quote prices against this verifier's local count — apply the storage-delta heuristic only to this node's own quote.

## Concerns

1. **No absolute age bound remains.** After this diff `quote.timestamp` is no longer read by the single-node verifier path (only signature still covers it). If the storage-delta heuristic happens to pass, an arbitrarily old quote can be replayed indefinitely. Worth adding an authenticated expiry (timestamp window, quote epoch, or block height) as a belt-and-braces check.

2. **Tests don't exercise the failure modes above.** The new "timestamp ignored" tests use `make_fake_quote()` with `price: Amount::from(1)` (`src/payment/verifier.rs:1674`), which derives to 0 records against a default zero-count verifier — so they prove the old error strings disappeared but don't exercise non-zero quotes, restart state, replication/prune drift, or the panic path. The dedicated delta tests around `src/payment/verifier.rs:1704` are meaningful but narrow. Adversarial tests for the overflow/panic case and for a non-zero verifier counter would be worth adding.

## Nits

- `src/payment/verifier.rs:455` still documents step 3 as "Quote timestamps are fresh"; doc no longer matches behavior.
- `src/storage/handler.rs:263` uses `as u64` for `records_stored()`; `u64::try_from(...)` would avoid a silent truncation on wider `usize` platforms.

## Positives

- `src/payment/pricing.rs:55` handles `price <= baseline` cleanly; integer arithmetic throughout, no lossy `f64` conversion. Round-trip tests for valid prices are useful.
- The "within tolerance" / "beyond tolerance" tests at `src/payment/verifier.rs:1704` directly exercise the intended heuristic instead of only asserting text.

## Questions

1. Was the intent really to compare every peer's quoted load against this verifier's local store count? Or should freshness be tied to the quoting peer's own previously observed count / a network-level metric?
2. If this heuristic stays, where should the verifier seed `records_stored` from persistent state on startup, and which paths should keep it in sync with replication writes and prune deletes?

PR #127, opened 2026-06-05
fix(payment): gate quote freshness on own quote only, not neighbours'

jacderida:
## Problem

The price-staleness gate introduced in 56dd370 checks **every quote in the payment bundle** against the **verifying node's own** record count:

```rust
for (encoded_peer_id, quote) in &payment.peer_quotes {
    if quote.price < min_acceptable_price { // floor derived from MY records
```

Fullness across a close group is wildly heterogeneous on a real network. On ant-prod-01 (2026-06-04, PROD-UL-01 upload run) the close group for a public DataMap chunk spanned **47..=1788 records**. The client collected 7 quotes and paid 3× the median on-chain (14,520,949,218,750,000 atto — more than 2× even the fullest node's current price), yet:

- node holding 1788 records: floor 5,271,199,218,750,000 → rejected the honest quote from the 978-record peer (4,840,316,406,250,000)
- nodes holding 1180 / 1310 records: rejected the honest quote from the 47-record peer (3,908,407,226,562,500) — the 1180-record node by a margin of just **1.05%**

All quotes were **10 seconds old** — no staleness was involved, only cross-node fullness spread. With a fourth close-group node disk-full, the PUT landed on 3/7 peers (majority is 4): the upload failed *after* the on-chain payment was burned, and the 500 already-stored body chunks became unreferenced. The same gate also rejects the replication paid-notify path (observed on two further nodes at 1180/1443 records), so the chunk cannot heal past the 3 copies.

Every close group whose fullest member's 75% floor exceeds its emptiest member's price will reject all single-node payments — guaranteed steady-state on prod where freshly joined nodes coexist with established ones.

## Fix

A node can only re-derive *its own* price from *its own* record count, so its own quote is the only one it can legitimately call stale (the doc comment's stated intent — "a node storing a few replicated records between quoting and verifying" — is exactly the self-staleness case). The gate now:

- resolves the node's own peer ID from the attached `P2PNode` (same handle the merkle closeness check uses) and skips every quote in the bundle that isn't ours;
- fails open with a debug log when no identity source is attached, mirroring the existing missing-record-count posture (production always attaches);
- rewords the rejection message — it previously claimed "paid X" while actually printing the quote price, not the on-chain payment.

The on-chain median payment binding (`completedPayments`, 3× median) is unaffected.

## Tests

- `test_neighbour_cheap_quote_not_rejected` — regression test reproducing the incident shape: own fresh quote at 1788 records alongside honest neighbour quotes at 47 and 978; must pass.
- `test_own_stale_quote_still_rejected_among_neighbours` — the gate still bites on a genuinely stale own quote.
- `test_freshness_skipped_without_self_peer_id` — fail-open without an identity source.
- Existing freshness tests updated to mark the quote under test as the node's own (`set_peer_id_for_tests`, cfg-gated like `set_records_stored_for_tests`).

All 61 `payment::verifier` tests pass; clippy clean. (`config::tests::test_bootstrap_peers_discover_env_var` fails identically on untouched `main` — environment-sensitive, unrelated.)

## Possible follow-up

An additional/alternative gate could compare the **verified on-chain amount** against the node's current price after `single_payment.verify()` — that would also fix the "rejected despite aggregate overpayment" semantics. Kept out of this PR to stay surgical.

Companion PR (reporting): WithAutonomi/ant-client#107 — `ant --json file upload` emits no JSON when the public DataMap store fails, so harnesses report 0/0 chunks for uploads like this one.

### DIFF
--- a/src/payment/verifier.rs
+++ b/src/payment/verifier.rs
@@ -162,6 +162,12 @@ pub struct PaymentVerifier {
     /// [`Self::set_records_stored_for_tests`] so unit tests that don't wire a
     /// real `LmdbStorage` can still drive the freshness logic.
     test_records_override: RwLock<Option<u64>>,
+    /// Test-only override for this node's own peer ID, used by
+    /// `validate_quote_freshness` to pick out the node's own quote from the
+    /// payment bundle. Production code derives it from the attached
+    /// [`P2PNode`]; set via [`Self::set_peer_id_for_tests`] so unit tests can
+    /// drive the freshness logic without wiring a real `P2PNode`.
+    test_peer_id_override: RwLock<Option<[u8; 32]>>,
     /// Configuration.
     config: PaymentVerifierConfig,
 }
@@ -280,6 +286,7 @@ impl PaymentVerifier {
             p2p_node: RwLock::new(None),
             storage: RwLock::new(None),
             test_records_override: RwLock::new(None),
+            test_peer_id_override: RwLock::new(None),
             config,
         }
     }
@@ -316,6 +323,30 @@ impl PaymentVerifier {
     #[cfg(any(test, feature = "test-utils"))]
     pub fn set_records_stored_for_tests(&self, count: u64) {
         *self.test_records_override.write() = Some(count);
+    }
+
+    /// Test-only setter for the node's own peer ID used by the quote
+    /// freshness check. Lets unit tests mark which quote in a payment bundle
+    /// is "ours" without wiring a real `P2PNode`. Has no effect in production
+    /// code because production code is expected to call
+    /// [`Self::attach_p2p_node`] instead.
+    #[cfg(any(test, feature = "test-utils"))]
+    pub fn set_peer_id_for_tests(&self, peer_id_bytes: [u8; 32]) {
+        *self.test_peer_id_override.write() = Some(peer_id_bytes);
+    }
+
+    /// Snapshot this node's own peer ID for the quote freshness check.
+    ///
+    /// Prefers the attached [`P2PNode`] (authoritative). Falls back to a test
+    /// override if one was set. Returns `None` only when no source is
+    /// available (mis-configured production startup); the caller treats that
+    /// as "unknown" and skips the freshness gate rather than rejecting — the
+    /// same fail-open posture as a missing record-count source.
+    fn self_peer_id_bytes(&self) -> Option<[u8; 32]> {
+        if let Some(node) = self.p2p_node.read().as_ref() {
+            return Some(*node.peer_id().as_bytes());
+        }
+        *self.test_peer_id_override.read()
     }
 
     /// Snapshot the current record count for freshness comparisons.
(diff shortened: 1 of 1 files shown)


grumbach (commented):
Reviewed the fix end to end (read `verify_evm_payment`, `validate_quote_freshness`, the self-ID resolution, the signing/binding path, and the merkle path). This is a clean, correct, well-scoped fix and the diagnosis matches the incident. Approving, with a couple of notes for your call.

### What I verified

- **Root cause is real and the fix is the right shape.** A close group spanning 47..=1788 records makes a full node's 75% floor exceed an empty node's honest price, so the old gate rejected perfectly fresh neighbour quotes. A node can only re-derive *its own* price from *its own* record count, so gating only its own quote is correct. The neighbour's stale-replay risk is genuinely that neighbour's gate to enforce when the PUT reaches it.
- **The self-ID comparison actually matches in production.** `self_peer_id_bytes()` returns `*node.peer_id().as_bytes()`, and an `EncodedPeerId` is constructed from exactly those bytes (`EncodedPeerId::new(*ant_peer_id.as_bytes())` in this file). Same 32-byte encoding, so `encoded_peer_id.as_bytes() != &self_peer_id` is sound and won't silently skip our own quote.
- **No bypass from the pre-binding ordering.** `validate_quote_freshness` runs before `validate_peer_bindings` and signature verification, so the `encoded_peer_id` label isn't authenticated yet when we pick out "our" quote. I chased the relabeling angle: an attacker could mislabel a forged cheap quote under a neighbour's peer ID to dodge the own-quote gate, but that quote then fails the BLAKE3 peer-id binding + signature (price is signed via `bytes_for_signing`). Both paths reject, so there's no surviving relabeled quote. The only untrusted-data op freshness performs (`derive_records_stored_from_price`) is already saturating/panic-proof for exactly this pre-auth reason. Fine as-is.
- **Scope is right.** `validate_quote_freshness` is only called from the single-node path; the merkle/batch path has no analogous own-fullness gate, so nothing to mirror there.

### Notes (non-blocking)

1. **Two independent self-identity notions now coexist in the verifier.** `validate_local_recipient` identifies "us" by `quote.rewards_address == local_rewards_address`; the new gate identifies "us" by `P2PNode` peer ID. If those two ever disagree for a given node (config mismatch, or one handle attached but not the other), `validate_local_recipient` could still consider the node a recipient while the freshness gate matches no quote and silently no-ops. It fails *open*, so it's not a security hole, just a latent inconsistency. Might be worth a one-line comment noting the two are expected to refer to the same node, or a debug log when the bundle contains our rewards address but no quote under our peer ID.

2. **Tiny doc-comment nit.** The new doc on `validate_quote_freshness` is excellent, but it says "this node has no basis to judge it against its own record count" for neighbour quotes, which is the key insight worth keeping front and center. Consider also stating explicitly that a bundle is expected to contain exactly one quote per peer (so the loop matches at most one own quote) so a future reader doesn't wonder about multiple-own-quote handling.

### Tests

The three new tests cover the regression shape (own fresh @1788 + honest neighbours @47/@978), own-still-bites, and fail-open-without-self-id. The fail-open test relies on no override being set, which is the right negative. Since these call `validate_quote_freshness` directly, the `pub_key: [0u8; 64]` fakes are fine (binding/signature aren't on this path). Coverage matches the change well.

Nice surgical fix. LGTM once you've eyeballed note 1.

jacderida:
Thanks for the thorough review @grumbach — both notes addressed in the amended commit (`51f7d55` → `8f69ae1`, history rewritten rather than stacked):

**Note 1 (two self-identity notions):** added a comment in `validate_quote_freshness` stating that the rewards-address identity (`validate_local_recipient`) and the peer-ID identity (this gate) are expected to refer to the same node, plus the `debug!` breadcrumb you suggested: when the bundle contains our rewards address but no quote under our peer ID, the fail-open is now logged instead of silent. That divergence is actually routine in our own fleet (all our nodes share a rewards address, so a PUT reaching a node whose own quote isn't in the bundle passes `validate_local_recipient` on a sibling's quote) — exactly the kind of path we'd have wanted in the node logs while reconstructing this incident.

**Note 2 (at-most-one own quote):** added a doc sentence noting that `validate_quote_structure` rejects duplicate peer IDs and runs before this gate on every path, so the loop matches at most one own quote.

No behaviour change on any rejecting path; clippy (CI flags) and all 61 verifier tests still pass.

PR #128, opened 2026-06-05
Gossip-triggered subtree audit + responsible-chunk audit (ADR-0002)

grumbach:
## Summary

Implements ADR-0002 on top of the v12 storage-bound audit base. The network now runs **two complementary storage audits** side by side:

- **Storage-commitment audit** (gossip-triggered): on ingesting a peer's commitment, a neighbour pins the just-gossiped root and runs a two-round contiguous-subtree proof — round 1 rebuilds the nonce-selected subtree to the pinned root; round 2 demands the original chunk bytes for a nonce-selected sample **from the audited node** (possession is non-delegable — the auditor holds none of the peer's chunks). Proves a node stores what it **claims** (no deletion, no relay).
- **Responsible-chunk audit** (periodic): audits a close peer for the chunks it **should** hold (responsibility + prior repair hint), via per-key possession digests. Proves a node stores what it **should**.

This PR replaces the closed #113 (the v12 base is squashed into one commit here; the ADR-0002 work is the second commit).

## Gossip-anchored retention (keeps commitments truthful)

A responder stays answerable for the chunk data behind its current commitment plus its **last two gossiped roots**, for a wall-clock TTL after each emission. The pruner consults `is_held` before deleting an out-of-range key, so an in-flight round-2 challenge can't false-positive an honest node; the commitment rebuild commits only to keys the node is still responsible for, so out-of-range chunks stop being re-committed and **age out** (bounded reprieve, not a permanent pin). A node with no responsible keys **retires** (stops advertising) its stale commitment while staying answerable until its TTL lapses; gossip emission snapshots-and-marks atomically so the node never advertises a root retention no longer covers.

## Accounting & rollout safety

Deterministic failures act on the first occurrence and revoke holder credit; only timeouts are graced (adaptive grace driven by liveness signals, never by deterministic failures). Timeout-driven eviction stays gated off this release (`TIMEOUT-EVICTION-DISABLED`) pending fleet upgrade; confirmed crypto/possession failures report. Density-aware closeness is observe-only. Both audits carry main's richer `AuditFailureSummary` diagnostics.

## Testing

- Lib unit tests (subtree honest-proof across sizes/nonces; retention TTL + retire-current regressions; adaptive-grace; pruner veto).
- Two PoC attack suites; live multi-node e2e: **honest → pass, data-deleter → confirmed fail, no honest false-positive, pruner-retention veto over the wire**.
- Both audits independently adversarially reviewed (codex xhigh + a second reviewer) across multiple rounds; the rebase onto main was re-reviewed for silent feature loss.
- cfd clean; all-targets `-D warnings` clean.

## Known follow-up

Timeout eviction re-enable once the fleet has upgraded (flip `TIMEOUT-EVICTION-DISABLED`).


mickvandijke (approved):
Great work! 🚢

PR #136, opened 2026-06-11
feat(payment): accept flexible single-node proof bundles

mickvandijke:
## Summary

This PR changes node-side payment verification for the existing `0x01` single-node proof format. It keeps the proof tag and wire serialization compatible, allows upgraded nodes to verify smaller quote bundles, and separates proof verification from receiver/admission placement checks.

The payment verifier now validates proof contents plus paid-quote issuer locality. The call sites decide whether this node is allowed to accept the write before payment verification runs:

- direct client PUT and fresh chunk replication require this node to be in the configured close group for the chunk address according to the local routing table
- fresh paid-list replication requires this node to be in the paid-list close group / K closest peers for the chunk address according to the local routing table
- the paid quote issuer must be in the configured close group for the quoted chunk address according to the verifier's local routing table

## Added

- Flexible `0x01` single-node proof verification for any `1..=CLOSE_GROUP_SIZE` quote bundle:
  - `0` quotes are rejected.
  - `1` quote is treated as the median-paid quote.
  - `2..=7` quotes are sorted by price and the median-priced quote is treated as the paid quote.
  - more than `CLOSE_GROUP_SIZE` quotes are rejected.
  - duplicate peer IDs are rejected.
- Paid median candidate verification without using `SingleNodePayment::from_quotes`, so partial bundles can be checked directly against `completedPayments(quote_hash)`.
- Live checks for each paid median candidate:
  - quote content matches the chunk address
  - quote public key derives the claimed peer ID
  - quote ML-DSA signature verifies
  - paid quote issuer is in this node's local configured close group for the chunk address
  - quote price satisfies the local 20% price-floor tolerance
  - on-chain `completedPayments(quote_hash)` is at least `3 * median_price`
- Direct client PUT storage responsibility in `AntProtocol` before payment verification/cache lookup.
- Context-aware verified-payment cache entries:
  - `ClientPut` entries satisfy later client PUT/fresh replication and paid-list lookups.
  - `PaidListAdmission` entries satisfy paid-list lookups only.
- `PaymentVerifierConfig::close_group_size`, wired from `ReplicationConfig` in production, devnet, and e2e test setup.
- `AntProtocol::attach_p2p_node`, so startup paths attach P2P through the protocol and keep direct PUT responsibility, paid-issuer, and merkle closeness checks wired consistently.
- Fresh-replication content-address validation before payment verification, matching direct PUT behavior.
- E2E/unit coverage for:
  - flexible single-node proof bundles
  - single-quote 3x payment requirement
  - paid-list cache entries not authorizing client PUTs
  - direct PUT storage responsibility running before payment cache hits
  - fresh chunk replication and paid-list notification propagation through a local testnet
  - merkle pay-yourself fabricated-pool rejection through the live DHT path

## Changed

- `PaymentVerifier::verify_payment` no longer checks whether this node is a receiver/member for the request. `VerificationContext` now controls cache strength only.
- Paid quote issuer locality is checked against the configured close group, not K closest peers.
- Direct client PUTs reject locally when this node is not in the configured close group for the chunk address, even if payment is cached.
- Fresh chunk replication checks close-group responsibility before payment verification, then uses `VerificationContext::ClientPut` because the receiver immediately stores the chunk as part of fresh write fan-out.
- Fresh paid-list replication checks paid-list close-group / K membership before payment verification, then uses `VerificationContext::PaidListAdmission` so it writes a paid-list-strength cache entry.
- Quote pricing/verifier comments were updated from "freshness" terminology to local paid-quote price-floor terminology.
- Devnet, production node startup, and e2e testnet startup now attach P2P via `AntProtocol` rather than attaching only the verifier directly.

## Removed

- Removed receiver/self membership checks from payment verifier functions.
- Removed the verifier's test-only receiver-membership override; direct PUT storage responsibility now has its own handler-level test override.
- Removed K-wide paid-quote issuer acceptance for single-node proofs.
- Removed the old `VerificationContext::Replication` proof-verification mode.
- Removed the old replication-verified cache level and replaced it with explicit `PaidList` vs `ClientPut` cache strength.
- Removed the exact-`CLOSE_GROUP_SIZE` requirement for single-node proof bundles.
- Removed full validation of non-median quotes in single-node proofs. Non-median quotes are now used only to derive the median price; the paid median candidate is the one whose content, peer binding, signature, issuer locality, local price floor, and on-chain payment are authoritative.
- Removed the old local-recipient/rewards-address check from single-node verification. Fresh receiver authorization now comes from call-site local placement checks before cache/proof verification, plus paid-issuer close-group locality for the paid quote.

## Unchanged

- No new proof tag is introduced.
- Existing 7-quote `0x01` proofs remain valid.
- Merkle proof verification still runs candidate signature checks, candidate closeness, on-chain merkle payment lookup/cache, merkle proof validation, and per-node payment checks.
- Later neighbour-sync repair paths still do not carry proof bytes and do not call the payment verifier. They continue to use existing closest-set quorum and paid-list evidence.

## Verification

Ran locally:

- `git diff --check`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test --lib --features test-utils -- --nocapture` — 526 passed
- `cargo test --test e2e --features test-utils data_types::chunk::tests::test_chunk_store_retrieve_small -- --nocapture` — passed, local minimal testnet
- `cargo test --test e2e --features test-utils replication::test_fresh_replication_propagates_to_close_group -- --nocapture` — passed, local replication testnet
- `cargo test --test e2e --features test-utils replication::scenario_24_fresh_replication_propagates_paid_notify -- --nocapture` — passed, local replication testnet
- `cargo test --test e2e --features test-utils payment_flow::helper_tests::test_init_testnet_and_evm -- --nocapture` — passed, local 10-node payment-enabled testnet + Anvil setup
- `cargo test --test e2e --features test-utils merkle_payment::test_attack_merkle_pay_yourself_fabricated_pool -- --nocapture` — passed, local merkle/DHT attack testnet

Note: `scripts/test_e2e.sh` was not run from this checkout because it requires `target/release/ant-cli`, which is not built/present in this repository checkout.

### DIFF
--- a/src/devnet.rs
+++ b/src/devnet.rs
@@ -10,6 +10,7 @@ use crate::payment::{
     EvmVerifierConfig, PaymentVerifier, PaymentVerifierConfig, QuoteGenerator,
     QuotingMetricsTracker,
 };
+use crate::replication::config::ReplicationConfig;
 use crate::storage::{AntProtocol, LmdbStorage, LmdbStorageConfig};
 use evmlib::Network as EvmNetwork;
 use evmlib::RewardsAddress;
@@ -550,9 +551,11 @@ impl Devnet {
         };
 
         let rewards_address = RewardsAddress::new(DEVNET_REWARDS_ADDRESS);
+        let replication_config = ReplicationConfig::default();
         let payment_config = PaymentVerifierConfig {
             evm: evm_config,
             cache_capacity: DEVNET_PAYMENT_CACHE_CAPACITY,
+            close_group_size: replication_config.close_group_size,
             local_rewards_address: rewards_address,
         };
         let payment_verifier = PaymentVerifier::new(payment_config);
@@ -611,10 +614,9 @@ impl Devnet {
         *node.state.write().await = NodeState::Running;
 
         if let (Some(ref p2p), Some(ref protocol)) = (&node.p2p_node, &node.ant_protocol) {
-            // Wire the P2PNode into the payment verifier for merkle-closeness checks.
-            protocol
-                .payment_verifier_arc()
-                .attach_p2p_node(Arc::clone(p2p));
+            // Wire P2P into AntProtocol for payment receiver-membership and
+            // payment-proof closeness checks.
+            protocol.attach_p2p_node(Arc::clone(p2p));
 
             let mut events = p2p.subscribe_events();
             let p2p_clone = Arc::clone(p2p);
--- a/src/devnet.rs
+++ b/src/devnet.rs
@@ -614,7 +614,7 @@ impl Devnet {
         *node.state.write().await = NodeState::Running;
 
         if let (Some(ref p2p), Some(ref protocol)) = (&node.p2p_node, &node.ant_protocol) {
-            // Wire P2P into AntProtocol for payment receiver-membership and
+            // Wire P2P into AntProtocol for direct PUT responsibility and
             // payment-proof closeness checks.
             protocol.attach_p2p_node(Arc::clone(p2p));
 
--- a/src/node.rs
+++ b/src/node.rs
@@ -107,11 +107,15 @@ impl NodeBuilder {
             Some(Self::build_upgrade_monitor(&self.config, node_id_seed))
         };
 
+        let repl_config = ReplicationConfig::default();
+
         // Initialize ANT protocol handler for chunk storage and
         // wire the fresh-write channel so PUTs trigger replication.
         let (ant_protocol, fresh_write_rx) = if self.config.storage.enabled {
             let (fresh_write_tx, fresh_write_rx) = tokio::sync::mpsc::unbounded_channel();
-            let mut protocol = Self::build_ant_protocol(&self.config, &identity).await?;
+            let mut protocol =
+                Self::build_ant_protocol(&self.config, &identity, repl_config.close_group_size)
+                    .await?;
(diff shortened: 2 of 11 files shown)


dirvine:
Re-review after `5e54ef6598512ccfc8d6084f737c1ce6206a712f`:

The paid-issuer-width concern from my earlier review looks addressed.

What I verified:
- The paid quote issuer is now checked with the configured close-group width via `find_closest_nodes_local_with_self(xorname, close_group_size)`, rather than `K_BUCKET_SIZE` / top-20.
- Direct PUT receiver responsibility is checked in `AntProtocol` before payment verification/cache use.
- Fresh replication still checks `config.close_group_size` responsibility before proof verification.
- `PaidListAdmission` keeps weaker cache semantics, so a paid-list cache entry does not authorise a later client PUT.

Local checks run on this head:
- `cargo test payment::verifier --lib --features test-utils -- --nocapture` — passed, 74 tests.
- `cargo test storage::handler::tests::test_put_rejects_out_of_range_receiver_before_payment_cache --lib --features test-utils -- --nocapture` — passed.

CI at time of review: builds, clippy, format, docs, security audit, and no-logging test are green; OS matrix tests are still pending. No remaining blocker from my focused review, pending those matrix tests completing green.

jacderida (approved):

PR #149, opened 2026-06-17
feat: commitment-bound quote pricing (ADR-0004)

grumbach:
**ADR-0004: commitment-bound quote pricing** — node side. Builds on the ADR-0002 storage-audit work (#128, merged).

Binds a quote's price to the node's audited storage commitment so a node cannot lie about how much it stores to inflate what it charges. The delivered guarantee is a **price ceiling**: to be paid above the empty-node baseline, a node must surrender a signed commitment that passes binding checks before payment and faces audit after it.

## What

- **Forced price.** A quote prices from the live commitment's committed key count and pins that commitment; price is the public formula of the count by *exact recomputation* (never price inversion). Both quote types (single-node + merkle candidate) carry and sign the binding.
- **Ship the commitment.** The node sends the signed commitment alongside each quote, so any receiver verifies the binding with no extra round trip.
- **Storers re-check + cross-check.** Every storer re-runs the arithmetic on every quote and cross-checks each claimed count against the pinned commitment (resolved from the shipped sidecar, fresh gossip, or a capped off-hot-path `GetCommitmentByPin` fetch), emitting portable mismatch evidence on a contradiction.
- **Deterministic first audit.** Monetized commitments enter a per-peer first-audit queue (newest-per-peer, dedup on real audit, retried across the per-peer cooldown), so the latest commitment earning a peer money faces an audit soon.
- **Grace removed (Amendment 1).** Audit answerability is made restart-durable and challenged only in-window, so a responsive audit miss is always a confirmed failure — no `UnknownCommitment`/`Transient` grace loophole.
- **No receipt bloat.** Commitment sidecars are stripped before a receipt is persisted or replicated.
- The commitment wire type, its pin/verification, and the pricing formula move to ant-protocol; this crate re-exports them.

Enforcement ships behind observe-only flags for a hard cutover, per the ADR's rollout. The ADR document and implementation-slices notes are included.

## Release ordering

Part of the coordinated ADR-0004 cutover — publish order **evmlib 0.9.0 → ant-protocol 2.3.0 → ant-node / ant-client**. evmlib's change is merged (WithAutonomi/evmlib#11) pending a `0.9.0` tag/publish; the shared layer is WithAutonomi/ant-protocol#15. This branch builds today via a committed, temporary `[patch.crates-io]` git-branch dep block (self-labeled *strip before merge*); at cutover that block is deleted and the `[dependencies]` pins (already `evmlib 0.9.0` / `ant-protocol 2.3.0`) resolve from the published crates.

Companion: ant-client resolve-before-pay (the client half).


jacderida (approved):

jacderida:
### Testnet analysis — DEV-01 (5h20m, 58 nodes ×15, 30% NAT) vs release 0.14.2 baseline

Ran this branch (paired with ant-client #126) on a staging testnet and compared against a `0.14.2` baseline on identical topology over a matched 5h20m window.

**Clean result** — 0 upload failures (2,977 uploads) and 0 download failures (662). Payment path healthy: 876k `Merkle payment verified` with **zero** admission-margin / reachability-rerank / `InvalidPut` / hard-reject signatures, so the new commitment-bound pricing did **not** produce rejections. ERROR volume is ~100% storage-commitment audit **timeouts** (`confirmed_failed_keys=0`, no data loss). No node service restarted during the run.

**Worth noting before release:**
- **One node panic/crash:** `JoinHandle polled after completion` (tokio) at 16:46 UTC → `SIGABRT`/core-dump. Units are `Restart=no` so it stayed down. Looks like a code bug (not resource exhaustion — disk 55%); worth a follow-up.
- **Node memory higher than 0.14.2** — fleet mean 33% vs 28%, peak 58% vs 37%. The commitment/audit machinery appears to carry a larger per-node footprint.
- Store latency rises as the data disk fills (3%→51%): `put_rpc` p95 14s→24s, p99 20s→38s over the run — drove a gradual upload slowdown. Nothing saturated (disk max 65%, CPU/mem healthy).

No blockers found; the panic and memory footprint are the two things to keep an eye on.

jacderida (approved):

PR #150, opened 2026-06-17
perf(replication): reduce neighbor sync CPU spikes

mickvandijke:
## Summary

Reduces the CPU amplification in outbound replication neighbor sync by batching hint construction per sync batch, and adds targeted timing logs around the hot paths identified during the `0.12.0` CPU spike investigation.

Semver: patch

## Background

The investigation pointed at two related changes in runtime behavior since `0.12.0`:

- replication receipts now reach the more expensive payment verification path instead of being rejected earlier
- successful verification grows local stored-key and paid-list datasets, making periodic neighbor sync scans more expensive over time

Before this change, outbound neighbor sync built hints independently for each peer. Each peer sync scanned all local stored keys and all paid-list keys, then ran DHT close-group checks for each key. For a batch of peers this made the cost scale as roughly `stored_keys * peers` plus `paid_keys * peers`, which can become visible as periodic CPU spikes on nodes with larger datasets.

## Changes

- Add `build_sync_hints_for_peers` to construct outbound sync hints for a whole peer batch with one storage scan and one paid-list scan.
- Wire bootstrap neighbor sync, scheduled neighbor sync, and replacement-peer sync through the prebuilt hint path.
- Keep the existing single-peer sync API by delegating it through the new batch builder for compatibility with current call sites.
- Add neighbor-sync hint build logs with peer count, stored-key count, paid-key count, emitted hint counts, and elapsed time.
- Add payment verification timing logs including context, detected proof type, proof byte length, result status, and elapsed time.
- Add quorum verification round timing logs with key count, peer count, requested key refs, and elapsed time.
- Add full replication verification cycle timing logs with pending/fetch/in-flight queue counts and terminal key count.

## Expected Impact

The sync-side work should now scale by scanning the local datasets once per selected peer batch, then distributing matching hints to the peers in that batch. This should reduce periodic CPU spikes from neighbor sync without changing replication protocol semantics.

The added logs should make any remaining spikes attributable to one of the following areas:

- neighbor-sync hint construction
- payment proof verification
- quorum verification requests
- full verification-cycle queue processing

## Validation

Ran successfully:

```text
cargo test --lib neighbor_sync
cargo test --lib quorum
cargo test --lib payment::verifier
cargo fmt --all -- --check
cargo clippy --all-targets --all-features -- -D warnings
```

### DIFF
--- a/src/payment/verifier.rs
+++ b/src/payment/verifier.rs
@@ -559,8 +559,14 @@ impl PaymentVerifier {
                     "Payment verification: context={context:?}, proof_type={proof_type}, proof_bytes={proof_bytes}, status={status:?}, elapsed_ms={elapsed_ms}",
                 );
             }
-            Err(e) => {
+            Err(e) if elapsed_ms >= PAYMENT_VERIFY_SLOW_LOG_MS => {
                 warn!(
+                    target: "ant_node::payment::verify",
+                    "Slow payment verification failed: context={context:?}, proof_type={proof_type}, proof_bytes={proof_bytes}, elapsed_ms={elapsed_ms}: {e}",
+                );
+            }
+            Err(e) => {
+                debug!(
                     target: "ant_node::payment::verify",
                     "Payment verification failed: context={context:?}, proof_type={proof_type}, proof_bytes={proof_bytes}, elapsed_ms={elapsed_ms}: {e}",
                 );
--- a/src/payment/verifier.rs
+++ b/src/payment/verifier.rs
@@ -30,6 +30,7 @@ use saorsa_core::P2PNode;
 use std::collections::HashMap;
 use std::num::NonZeroUsize;
 use std::sync::Arc;
+use std::time::Instant;
 
 /// Minimum allowed size for a payment proof in bytes.
 ///
@@ -60,6 +61,7 @@ const PAID_QUOTE_PRICE_FLOOR_TOLERANCE_PCT: u64 = 20;
 
 const PERCENT_DENOMINATOR: u64 = 100;
 const PAID_QUOTE_PAYMENT_MULTIPLIER: u64 = 3;
+const PAYMENT_VERIFY_SLOW_LOG_MS: u128 = 500;
 
 /// Number of nearest DHT peers accepted for paid-quote issuer locality.
 ///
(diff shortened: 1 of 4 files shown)


dirvine:
Reviewed for RC risk, especially against the concern that this was more than just the SNP security-tightening revert.

**Verdict: low risk / OK for RC** from my review.

What I checked:

- **SNP/payment security:** no weakening found.
  - `verify_payment` is wrapped with timing/logging, then delegates to the previous logic now in `verify_payment_inner`.
  - No new `2*K`, `storage_admission_width`, network fallback, or receiver/issuer authority widening in the SNP path.
  - Proof size/type handling, median quote selection, issuer locality, signature/pubkey binding, price floor, on-chain payment check, and cache/context semantics are preserved.

- **CPU spike mitigation:** directionally sound.
  - Previously neighbour sync rebuilt hints per peer, scanning all stored keys and paid-list keys repeatedly.
  - New path scans once per selected peer batch and distributes hints to peers.
  - With the default small batch size, this should materially reduce the periodic CPU spike shape.

- **Replication/repair correctness:** looks conservative.
  - Failed syncs still do not record repair proofs.
  - Replacement peers get freshly built hints, not reused hints from the failed peer.
  - The sent hint list and repair-proof metadata still come from the same `SentReplicaHint` vector.

Non-blocking caveats:

- Batch prebuilding widens the “hint snapshot may be slightly stale” window from per-peer to per-batch. Given small batches and audit-time reconciliation, I don’t see this as RC-blocking.
- There’s no direct equivalence test proving batch-built hints exactly match old per-peer-built hints; worth adding later.
- Slow-path info/warn logs could add noise during the very spikes being investigated, but fast payment failures have been demoted to debug, and no raw proof bytes are logged.

Validation checked:

- GitHub checks all green on head `afaaed356ebcc5efd3a75cfb454f201627490076`.
- Local targeted checks passed:
  - `cargo test -q payment::verifier --lib --no-fail-fast` — 75 passed.
  - `cargo test -q replication::neighbor_sync --lib --no-fail-fast` — 24 passed.

**Recommendation:** OK to merge into the RC. It is not just the SNP revert, but the extra delta is small, well-scoped, and addresses the CPU spike investigation without relaxing the SNP security model.

PR #156, opened 2026-06-22
fix(replication): handle inbound overload backpressure

mickvandijke:
## Summary
- Bound inbound replication request handling with global and per-peer semaphores, plus capped overload response tasks so the P2P event loop does not run expensive replication handlers inline.
- Add an explicit `Overloaded` replication message for request types without typed rejection responses, and map overload into neutral deferrals for neighbor sync, verification, prune audit, fetch, and audit flows.
- Track overload-deferred bootstrap syncs, priority retry those peers, and prevent bootstrap from draining before deferred sync work is resolved.
- Add per-peer overload-claim budgets for audit/fetch so an unauthenticated overload reason cannot indefinitely avoid accountability.
- Preserve pending verification work across inconclusive rounds with a bounded retry counter, and let bootstrapping replica hints fetch from present holders even when full quorum is not available.

## Details
- Inbound request handling now accepts work only when both global and per-peer limits have capacity. Accepted work is spawned into tracked tasks; rejected request-response messages receive typed overload responses where possible.
- `NeighborSyncAttempt` distinguishes completed, failed, and overloaded sync attempts. Overloaded bootstrap syncs block drain until a priority retry completes, fails, or the peer leaves.
- Verification handles `Overloaded` responses as unresolved evidence, caps consecutive inconclusive rounds, and returns stale evictions to bootstrap bookkeeping.
- Fetch overload errors are retried against the same source within a small budget, then treated as source failures to allow alternate sources.
- Audit overload claims are honored briefly, then counted as audit failures once the peer exceeds the consecutive-claim budget. Prune audits treat overload as neutral because it only keeps the local replica.

## Validation
- `git diff --check`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test --lib --all-features replication::`
- `cargo test --test e2e --all-features replication::test_late_joiner_replicates_responsible_chunks -- --exact`

## Note
A broader `cargo test --all-targets --all-features replication::` run completed the library suite and 29/30 replication e2e tests but had one late-joiner convergence timeout after 90s. The exact failed e2e test passed on rerun in 71.36s.

### DIFF
--- a/src/replication/audit.rs
+++ b/src/replication/audit.rs
@@ -19,6 +19,7 @@ use crate::replication::protocol::{
 use crate::replication::types::{
     AuditFailureReason, AuditFailureSummary, FailureEvidence, PeerSyncRecord, RepairProofs,
 };
+use crate::replication::{is_inbound_replication_overloaded_reason, OverloadClaimTracker};
 use crate::storage::LmdbStorage;
 use saorsa_core::identity::PeerId;
 use saorsa_core::P2PNode;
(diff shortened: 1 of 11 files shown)


mickvandijke:
superseded by #158

PR #169, opened 2026-07-09
Add ADR-0005 node-side audit tally for earned reward eligibility

grumbach:
## What

The node side of **ADR-0005: earned reward eligibility**. Every node already audits its neighbours (ADR-0002/0004); this PR just writes down the results, per peer per day, so a client can decide at quote collection whether a quoter has earned a clean audited week at the size it wants to charge for. Nothing here gates payment on its own; the client-side gate lands in the ant-client PR, and the shared wire types are in ant-protocol#19.

## Changes

- **`replication/audit_tally.rs` (new)** — per-peer, per-day tally of subtree-audit passes and the size proven. Sticky convictions (a caught cheater re-earns its whole week), and a fence that freezes an observer's vouch when a monetized-pin challenge goes unanswered until the peer next passes. Rows and an evidence-less fence age out with the window. The on-disk snapshot carries a magic + version tag so a layout change is rejected cleanly rather than mis-decoded positionally, and the observer safely starts fresh (a lost snapshot only costs one testimony window, re-earned through ordinary audits).
- **`payment/quote.rs`** — attach the node's signed audit report, bound to the client's `report_nonce`, to quote responses.
- **`replication/*` + `storage/handler.rs`** — wire pass / conviction / fence recording into the existing audit paths and persist the tally with the same write-on-change atomic-rename discipline as the ADR-0004 retention snapshot.

## Dependency

Builds against the ADR-0005 wire types in **WithAutonomi/ant-protocol#19**. `Cargo.toml` temporarily pins `ant-protocol` at that PR branch so CI can build; this reverts to a published version bump once the wire types are released (same release dance as ADR-0004).

## Testing

`cargo test --lib` (695 pass, including the tally, fence age-out, and snapshot-tag tests), `cargo clippy --all-targets --all-features` with the panic/unwrap/expect denies clean, `cargo fmt --check` clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -861,9 +861,8 @@ dependencies = [
 
 [[package]]
 name = "ant-protocol"
-version = "2.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5b665880152b4e124094acc1faaff5f20a275594e6364b8a729b50f3bc326674"
+version = "3.0.0"
+source = "git+https://github.com/grumbach/ant-protocol?rev=2015113270bbde8204a1740674bed2ada52d2ca6#2015113270bbde8204a1740674bed2ada52d2ca6"
 dependencies = [
  "blake3",
  "bytes",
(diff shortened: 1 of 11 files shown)


grumbach:
Parking ADR-0005 (earned reward eligibility) for now: deprioritized while we focus on the ADR-0004 production fixes and capacity recovery. Closing to keep the review queue tidy, will reopen when we revisit. Branch stays intact.

PR #172, opened 2026-07-10
feat: add cumulative replication traffic accounting

jacderida:
## Summary

Part of **V2-623** (log-based traffic accounting). Chunk fetch/offer and
replication payloads are invisible in the logs today. This adds cumulative,
monotonic per-variant byte/message counters as INFO summary lines so the
telegraf→Elasticsearch pipeline can attribute how much bandwidth each
replication message type moves, per direction.
**Purely additive: no wire changes, no behaviour changes.**

- A process-global per-variant counter table in the replication protocol module
  (relaxed atomics), incremented at the single `encode` (tx) and `decode` (rx)
  choke points that every replication message passes through.
- New `replication traffic summary (cumulative)` INFO lines (target
  `ant_node::replication::traffic`) from a periodic task in the replication
  engine.

**Note on line splitting:** the 17 variants × 4 flat keys (68) exceed
`tracing`'s 32-field event cap, so the flat keys (`fresh_offer_tx_bytes`, …) are
split across three lines sharing the same target and message and distinguished
by a `group` field. The shipping pipeline lifts each key into its own field, so
the acceptance query (max per field per hour → delta → per-variant MB/h) is
unaffected.

## Test plan

- [x] `cargo check` (default features)
- [x] `cargo check --no-default-features` (logging compiled out)
- [x] `cargo clippy --all-features -- -D warnings -D clippy::panic -D clippy::unwrap_used -D clippy::expect_used`
- [x] `cargo fmt --all -- --check`

### DIFF
--- a/src/replication/mod.rs
+++ b/src/replication/mod.rs
@@ -165,6 +165,11 @@ const COMMITMENT_ROTATION_INTERVAL_SECS: u64 = 3600;
 /// guard keeps idle nodes from needless disk writes.
 const RETENTION_PERSIST_INTERVAL_SECS: u64 = 30;
 
+/// Cadence of the `replication traffic summary (cumulative)` INFO lines
+/// (V2-623). A `const` so testnets can drop it to 60s; 300s is the production
+/// default that keeps log volume negligible.
+const TRAFFIC_SUMMARY_INTERVAL_SECS: u64 = 300;
+
 /// Maximum tolerated auditor↔responder wall-clock skew for the first-audit
 /// in-window screen (ADR-0004 A1 guardrail A). The screen accepts a monetized pin
 /// for first audit only if its SIGNED `quote_ts` lands in
@@ -597,6 +602,8 @@ impl ReplicationEngine {
         // ADR-0004: deterministic first audit of commitments that backed a
         // payment (surfaced by the verifier cross-check).
         self.start_first_audit_drainer();
+        // V2-623: periodic cumulative per-variant traffic accounting.
+        self.start_traffic_summary_loop();
 
         info!(
             "Replication engine started with {} background tasks",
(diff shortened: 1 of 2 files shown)


dirvine (commented):
## Hermes review — conditionally ready after CI and ingestion smoke test

No wire-format or replication-behaviour change found. The encode/decode accounting is internally consistent, and local `cargo check` plus `cargo check --no-default-features` passed at head `a86521345d8cdd63d795537765f341610532122a`.

Caveats:
- TX counters measure successfully encoded replication messages, not confirmed network sends. Failed sends/retries can therefore exceed actual egress. Please name/document these as encoded/attempted traffic unless accounting is moved to the send-success path.
- The fixed 17-variant index/table is currently exhaustive and in bounds, but future enum/table drift could panic. A focused test asserting unique in-range indices and coverage of all emitted groups would cheaply guard this.
- The three sparse INFO events are workable for per-field cumulative queries, but please verify all three groups and all 68 fields survive the real Telegraf/Elasticsearch pipeline.

Current CI: format, Clippy, docs, security audit, no-logging test, and Linux/macOS builds pass; Windows build and platform tests are still pending.

Recommendation: no code-level production blocker found for diagnostic use, subject to all required CI completing green and one >5-minute ingestion smoke test. Not approved yet while CI is pending.

dirvine (approved):
## Follow-up approval

No code-level blocker found at head `a86521345d8cdd63d795537765f341610532122a`.

The counters consistently measure successfully encoded/decoded replication protocol traffic. TX is encoded/attempted volume rather than confirmed network delivery; that is acceptable for diagnostic use provided the query retains that interpretation.

Format, Clippy, documentation, security audit, no-logging and all platform builds are green. macOS and Windows tests were still pending at the latest check; approval does not waive those required jobs.

A Telegraf/Elasticsearch smoke test is optional canary validation of the three grouped events and field mappings, not a code-safety gate.

PR #174, opened 2026-07-13
feat(ant-devnet): opt-in LAN / external-devnet mode (--host / --evm-network / --serve-port)

Nic-dorman:
## What

Adds a **LAN / external-devnet scenario** to the existing `ant-devnet` harness so a devnet can be tested from another device — a phone, an iOS simulator, or a second machine — rather than only from the host box. Three opt-in flags, all **default off**:

| Flag | Effect |
|---|---|
| `--host <ipv4>` | Bind `0.0.0.0` and advertise this LAN IP in the manifest's bootstrap addresses, so other LAN devices can connect. |
| `--evm-network arbitrum-sepolia` | Verify payments against the **real deployed Arbitrum Sepolia contracts** (no local Anvil, empty wallet key) — exercises the external-signer payment flow. |
| `--serve-port <port>` | Serve the manifest over a small read-only HTTP API (`GET /api/devnet-manifest.json` + `/api/info`) so devices fetch it instead of copying files. |

```bash
# unchanged default behavior (single machine, local Anvil)
ant-devnet --preset small --enable-evm

# LAN devnet backed by Arbitrum Sepolia, manifest served over HTTP
ant-devnet --preset small --host 192.168.1.100 --evm-network arbitrum-sepolia --serve-port 8088
```

## Why here

`ant-devnet` is already the maintained local-devnet tool; this is a focused enhancement to it, not a new binary or module — **no new dependencies** (the manifest server is hand-rolled HTTP/1.1). It's the node half of enabling on-device testing of the mobile SDK; the client half already shipped in the Autonomi mobile FFI.

## Backward compatibility / blast radius

- **Opt-in and default-off.** No `--host` → `advertise_ip = None` → `.local(true)` + `127.0.0.1` bootstrap addresses, byte-identical to today.
- **Confined to the harness.** Only the `ant-devnet` binary and the `Devnet`/`DevnetConfig` type change. The real node binary (`ant-node`) and the wire protocol are untouched, and nothing here is on the production node path — so **deployed testnets are unaffected**.
- **Minor API note:** adds one field to the public `DevnetConfig` (`advertise_ip: Option<Ipv4Addr>`). It's covered by the `Default` impl, so `..Default::default()` construction is unaffected; only a struct-literal build without `..Default::default()` would need the new field.

## Tests

- `cli.rs`: LAN flags default to `None` (backward-compat contract); parse into the expected typed values; non-IPv4 `--host` is rejected.
- `devnet.rs`: `DevnetConfig::default().advertise_ip` is `None`.
- `cargo fmt --all --check` clean; strict clippy gates (`-D clippy::panic/unwrap_used/expect_used`) pass; release build green.

Proven end-to-end out-of-band: a physical Android device completed a full paid upload + byte-perfect download round-trip against a Sepolia-backed LAN devnet launched with these flags.

### DIFF
--- a/src/bin/ant-devnet/cli.rs
+++ b/src/bin/ant-devnet/cli.rs
@@ -80,8 +80,9 @@ pub struct Cli {
     /// Serve the manifest over a read-only HTTP API on this port (binds
     /// 0.0.0.0). Any LAN device can then GET
     /// `http://<host>:<port>/api/devnet-manifest.json` (and `/api/info`) —
-    /// no file copying. Open CORS. Suggested: 8088.
-    #[arg(long)]
+    /// no file copying. Open CORS. Suggested: 8088. Requires `--host` (the API
+    /// advertises a LAN URL, so a loopback-only devnet would be misleading).
+    #[arg(long, requires = "host", value_parser = clap::value_parser!(u16).range(1..))]
     pub serve_port: Option<u16>,
 }
 
@@ -121,5 +122,20 @@ mod tests {
     #[test]
     fn host_rejects_non_ipv4() {
         assert!(Cli::try_parse_from(["ant-devnet", "--host", "not-an-ip"]).is_err());
+    }
+
+    /// `--serve-port` requires `--host` (it advertises a LAN URL).
+    #[test]
+    fn serve_port_requires_host() {
+        assert!(Cli::try_parse_from(["ant-devnet", "--serve-port", "8088"]).is_err());
+    }
+
+    /// `--serve-port 0` is rejected (an ephemeral port would be advertised as `:0`).
+    #[test]
+    fn serve_port_rejects_zero() {
+        assert!(
+            Cli::try_parse_from(["ant-devnet", "--host", "192.168.1.5", "--serve-port", "0"])
+                .is_err()
+        );
     }
 }
--- a/src/bin/ant-devnet/cli.rs
+++ b/src/bin/ant-devnet/cli.rs
@@ -62,4 +62,64 @@ pub struct Cli {
     /// payments against the local chain.
     #[arg(long)]
     pub enable_evm: bool,
+
+    /// Advertise this IPv4 to peers/clients and bind 0.0.0.0, so the devnet is
+    /// reachable from other devices on the LAN. When omitted, nodes bind
+    /// loopback (127.0.0.1) as before (single-machine only).
+    #[arg(long)]
+    pub host: Option<std::net::Ipv4Addr>,
+
+    /// EVM network for node payment verification. `arbitrum-sepolia` makes
+    /// nodes verify against the real deployed Arbitrum Sepolia contracts —
+    /// no local Anvil and no embedded wallet key (bring your own funded
+    /// wallet via an external signer). Omit for the local-Anvil devnet
+    /// (`--enable-evm`).
+    #[arg(long)]
+    pub evm_network: Option<String>,
+
+    /// Serve the manifest over a read-only HTTP API on this port (binds
+    /// 0.0.0.0). Any LAN device can then GET
+    /// `http://<host>:<port>/api/devnet-manifest.json` (and `/api/info`) —
+    /// no file copying. Open CORS. Suggested: 8088.
+    #[arg(long)]
+    pub serve_port: Option<u16>,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::net::Ipv4Addr;
+
+    /// Backward-compatibility contract: with none of the LAN flags, the new
+    /// options parse to `None`, so behavior is identical to before.
+    #[test]
+    fn lan_flags_default_off() {
+        let cli = Cli::parse_from(["ant-devnet", "--preset", "small"]);
+        assert!(cli.host.is_none());
+        assert!(cli.evm_network.is_none());
+        assert!(cli.serve_port.is_none());
+    }
+
+    /// The LAN flags parse into the expected typed values.
+    #[test]
+    fn lan_flags_parse() {
+        let cli = Cli::parse_from([
+            "ant-devnet",
+            "--host",
+            "192.168.1.100",
+            "--evm-network",
+            "arbitrum-sepolia",
+            "--serve-port",
+            "8088",
+        ]);
+        assert_eq!(cli.host, Some(Ipv4Addr::new(192, 168, 1, 100)));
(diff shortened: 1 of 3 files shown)


Nic-dorman:
Addressed the review feedback in 96ab80a:

- **`--host` validation** — a loopback/unspecified `--host` is now rejected early with a clear error (it would otherwise stamp unreachable bootstrap addresses into the manifest).
- **GET enforcement** — the manifest API now returns `405` for any method other than GET.
- **`--evm-network` / `--enable-evm`** — made mutually exclusive at the clap level (`conflicts_with`).
- **Manifest server read timeout** — handler sockets get a 5s read timeout so a slow/idle client can't tie up a thread (the listener binds 0.0.0.0).

Not changed, with rationale:
- **`local_ip_guess()` returning IPv6** — the socket binds `0.0.0.0` (IPv4) and connects to an IPv4 address, so `local_addr()` is always IPv4; it can't return an IPv6 literal here.
- **`host_ip` use-after-move** — this compiles as written (green build on all three OSes); `json!`/`format!` don't consume `host_ip` the way the comment suggests.

fmt + clippy (`-D warnings`) + tests green.

dirvine (commented):
Reviewed current head `96ab80aaa2fbee966a00e83c08b30c36fda32677` with a multi-reviewer pass and local runtime checks.

**Verdict: changes requested before merge.** The latest commit fixes loopback/unspecified `--host`, GET-only handling, conflicting EVM flags, and idle-read timeout. Two functional issues remain:

1. **`src/bin/ant-devnet/main.rs:260-267` — `--serve-port` bind failure is swallowed.** `spawn_manifest_server` only warns and returns, then `serve_manifest_api` returns `Ok(())` and the CLI reports `Devnet running`. I reproduced this with an occupied port on the current head: the process logged `failed to bind ... Address already in use`, then continued successfully. Please bind synchronously and propagate the error.

2. **`src/bin/ant-devnet/cli.rs:84-85` / `main.rs:237,261` — port `0` is accepted but advertised as `:0`.** The OS actually selected port `49241` in my current-head run, while the log and `/api/info` construction retained port `0`. Either reject zero or return/use `listener.local_addr()`.

**Also worth fixing in this pass:**
- `--serve-port` without `--host` binds `0.0.0.0` and publishes a LAN-looking API URL, but the served manifest still contains loopback bootstrap addresses. Require `--host`, or bind/report loopback when it is absent.
- The five-second timeout bounds idle connection lifetime, but the server still creates an unbounded OS thread per connection. A small concurrency cap/inline handler would remove the remaining burst-exhaustion path.
- `--host` now rejects loopback/unspecified addresses, but multicast and broadcast values remain accepted and are not usable bootstrap destinations.
- Current `ant-cli` does not infer Sepolia from the manifest alone; users need `--evm-network arbitrum-sepolia` (or `local` to consume the manifest's custom RPC/contracts). This is not necessarily a blocker for the mobile flow, but should be documented unless the manifest schema gains an explicit network/chain ID.

Local verification on the current head:
- `cargo test --bin ant-devnet` — 3 passed
- `cargo test default_config_has_no_advertise_ip --lib` — 1 passed
- strict `cargo clippy --bin ant-devnet --all-features` — passed
- runtime occupied-port and port-zero cases reproduced as above

The reported Android paid upload/download result was not independently reproduced during this review.

Nic-dorman:
Thanks @dirvine — addressed in 998493d.

**Blockers:**
1. **Swallowed bind failure** — `serve_manifest_api` now binds the listener **synchronously** and propagates the error via `?`, so an occupied port fails the run instead of the devnet reporting `running` without its API.
2. **`--serve-port 0`** — rejected at the parser (`value_parser = value_parser!(u16).range(1..)`), so `:0` can't be advertised. (`ant-devnet --host … --serve-port 0` → `error: invalid value '0' … 0 is not in 1..=65535`.)

**Also fixed this pass:**
- **`--serve-port` without `--host`** — `--serve-port` now `requires = "host"` at the clap level, so it can't publish a LAN URL over a loopback-only devnet.
- **Unbounded thread-per-connection** — the server now handles connections **inline in its single thread** (no per-connection thread), keeping the 5s read timeout. Payloads are tiny and a devnet serves a handful of devices, so serial handling bounds resource use with no burst-exhaustion path.
- **Multicast/broadcast `--host`** — the early validation now also rejects `is_multicast()` / `is_broadcast()`, not just loopback/unspecified.
- **Bootstrap serialization** — `/api/info` now propagates the `serde_json::to_value(&bootstrap)` error instead of substituting `null`.

**Not changed here (documentation / follow-up):**
- **`ant-cli` can't infer Sepolia from the manifest alone** — correct: the manifest's `evm` block carries `rpc_url`/token/vault but no network identifier, so consumers pass `--evm-network arbitrum-sepolia` (or `local` to use the manifest's custom RPC/contracts). This is a manifest-schema matter rather than an `ant-devnet` behavior; the cleanest fix is adding an explicit `chain_id` to the manifest `evm` block so clients can infer it. That touches the shared manifest schema (and its FFI consumers), so I'd propose it as a separate follow-up rather than fold it in here — happy to open that if you agree.

Local: `cargo fmt --check`, strict `clippy -D warnings`, and `cargo test --bin ant-devnet` (5 passed) all green; the occupied-port, port-0, serve-port-without-host, and multicast-host cases verified by hand.

dirvine (commented):
Re-reviewed current head `998493dd8a13f842ec2b16d00ffa51fa31240916` against the previous reviewed head.

The previous blockers are addressed:
- manifest-listener bind errors now propagate;
- port zero and `--serve-port` without `--host` are rejected;
- loopback/unspecified/multicast/broadcast hosts are rejected;
- unbounded per-connection threads are removed;
- bootstrap serialisation errors propagate.

Local/runtime verification:
- `cargo test --bin ant-devnet` — 5 passed
- `cargo test default_config_has_no_advertise_ip --lib` — passed
- strict clippy and build — passed
- occupied-port case now exits with the expected bind error
- normal GET succeeds and invalid host classes are rejected

**Recommendation: approve once the remaining platform CI tests pass.** No remaining blocker was found for the stated opt-in trusted-LAN devnet scope. The single-read HTTP parser, process-scoped detached listener, intentional `DevnetConfig` field addition, and possible data-directory residue on a post-start error are non-blocking limitations here.

At review time, format, clippy, documentation, all three builds, no-logging tests and security audit are green. Ubuntu/macOS/Windows test jobs remain in progress.

jacderida (approved):

PR #176, opened 2026-07-14
Receiver-side price floor for single-node store admissions

grumbach:
## Summary

Node-side companion to [ant-client #150](https://github.com/WithAutonomi/ant-client/pull/150) (one-quote uploads). ADR-0004 gives every quote a price **ceiling** (price = f(committed key count)); removing the old local floor (`b52f4577`) left no revenue **floor**. Since the verifier accepts 1..=7 quote proofs and fully validates only the median candidate, a modified client can fetch quotes from the whole neighbourhood and settle only the cheapest valid one. The flexible bundles from #136 were safe while the old floor stood; the combination is the hole, and one-quote clients make it routine.

This PR restores the floor on the receiver, priced from the receiver's own live commitment:

```text
settled_amount >= 3 x tolerance% x calculate_price(local committed responsible key count)
```

alongside the existing `settled >= 3 x median` check. The floor compares the **settled amount**, so an honest client may overpay a cheap quote to clear stricter receivers.

## Design points

- **Same pricing basis as our own quotes.** The floor reads the commitment via a new non-mutating `CommitmentSource::current_binding_snapshot()` — the exact source `QuoteGenerator` prices from, wired in `AntProtocol::attach_commitment_source`. Never `current_chunks()` (the unlike-counts comparison that false-rejected honest quotes under the pre-ADR-0004 floor), and never `current_binding_for_quote()` (which would extend commitment answerability as a side effect).
- **No commitment = baseline floor.** A fresh/retired/restarting receiver (or one with nothing attached) prices the floor at `calculate_price(0)`, making it vacuous rather than an outage mode.
- **Shadow-first rollout.** Default is observe-only: one telemetry line per evaluated admission (target `ant_node::payment::price_floor`) with paid price, settled amount, local count/price, and the would-reject verdict. Enforcement is a per-node canary opt-in via `ANT_PRICE_FLOOR_ENFORCE=1` (tolerance override `ANT_PRICE_FLOOR_TOLERANCE_PERCENT`, default 50%); unsetting it is the kill switch. Clients need 4-of-20 stores, so canary enforcement cannot fail honest uploads while telemetry calibrates the tolerance.
- **Scoped to fresh economics only.** New `VerificationContext::FreshReplication` splits fan-out from direct PUT — verified identically everywhere (store-strength cache, cross-check, first-audit) so behaviour is unchanged; the floor applies to both (a single cheap accepting node must not fan a below-floor proof around everyone else's floor), and never to paid-list admission, repair, or cache hits. No historical receipt is repriced. Merkle proofs untouched.
- **Economic decision only.** A floor rejection is a payment decline; it feeds no trust/misbehaviour scoring.

## Tests

- Cheapest-of-K adversarial gate: a 3x-baseline settlement against a fuller receiver is rejected under enforcement, on both direct PUT and fresh replication.
- Exactly-at-floor overpayment accepted; shadow mode never rejects; no-commitment receivers stay vacuous; paid-list admission never repriced.
- Full lib suite (692 tests), clippy (all targets + `-D warnings`), fmt, and docs clean.

## Rollout

1. Ship in shadow mode fleet-wide, watch `price_floor` telemetry for the settled/local-price ratio distribution.
2. Enable enforcement on a few canary nodes, confirm zero honest rejections.
3. Widen enforcement ahead of the one-quote client becoming the common path; tune tolerance from data.
4. Follow-up: companion ADR documenting the floor as its own economic policy (it amends ADR-0004's "a node may always charge less").

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -158,6 +158,14 @@ name = "poc_bootstrap_stall"
 path = "tests/poc_bootstrap_stall.rs"
 required-features = ["test-utils"]
 
+# Live-Anvil verification of the ADR-0006 price floor against a real deployed
+# vault + real on-chain settlement. Needs the test-utils fail-open for the
+# issuer K-closest check so the floor is the isolated load-bearing check.
+[[test]]
+name = "poc_price_floor_live"
+path = "tests/poc_price_floor_live.rs"
+required-features = ["test-utils"]
+
 [features]
 default = ["logging"]
 # Enable tracing/logging infrastructure.
--- a/docs/adr/ADR-0006-receiver-side-revenue-floor.md
+++ b/docs/adr/ADR-0006-receiver-side-revenue-floor.md
@@ -79,7 +79,11 @@ settled_amount >= 3 x tolerance% x calculate_price(receiver's current
 - **Settled amount, not quote price.** An honest client may overpay a cheap
   quote to clear stricter receivers.
 - **No commitment = baseline.** A fresh, retired, or restarting receiver
-  prices the floor at `calculate_price(0)` — vacuous, never an outage mode.
+  prices the floor at `calculate_price(0)`. Every on-curve honest quote clears
+  a baseline floor at a 3x settlement, so this is not an availability
+  regression. (Only an off-curve sub-baseline quote — which the ADR-0004
+  arithmetic gate rejects once enforced — could fall below it, an economic
+  decline rather than an outage.)
 - **Scope.** Applies to `ClientPut` and `FreshReplication` (a below-floor
   proof must not fan out through one cheap accepting node). Never applies to
   paid-list admission, later repair, cache hits, or merkle-batch proofs: no
--- a/docs/adr/ADR-0006-receiver-side-revenue-floor.md
+++ b/docs/adr/ADR-0006-receiver-side-revenue-floor.md
@@ -1,0 +1,157 @@
+# ADR-0006: Receiver-side revenue floor for single-node store admissions
+
+- **Status:** Proposed
+- **Date:** 2026-07-14
+- **Decision owners:** Anselme (@grumbach)
+- **Reviewers:** <pending>
+- **Supersedes:** none (amends the economic scope of ADR-0004; see Decision)
+- **Superseded by:** none
+- **Related:** ADR-0004 (commitment-bound quote pricing), ant-node #136
+  (flexible 1..=7 proof bundles), ant-node #176 (implementation),
+  ant-client #150 (one-quote uploads)
+
+## Context
+
+ADR-0004 made a quote's price an exact function of the issuer's signed
+storage commitment. That is a **price ceiling**: a node cannot charge more
+than the storage it can prove. ADR-0004 explicitly noted that "a ceiling is
+not a revenue floor" and left the floor to a follow-up policy. Its
+implementation also retired the old receiver-side floor
+(`validate_paid_quote_price_floor`), which compared quote prices against the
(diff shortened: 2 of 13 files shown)


dirvine (commented):
## Review — receiver-side price floor

Reviewed current head `bfe16d733c9be1ca887db3097b2e7acbb9a3d52b` with six independent passes and local verification.

**Verdict: no protocol/security blocker for merging the default-shadow implementation.** This is not approval to enable enforcement yet.

### What holds

- `ClientPut` and `FreshReplication` both use store-strength verification and floor evaluation; `PaidListAdmission`, repair, Merkle proofs and existing store-cache hits remain deliberately exempt.
- Paid-list cache entries cannot satisfy store admission, and insertion happens only after the authenticated paid quote, on-chain settlement and floor checks succeed.
- The floor uses the receiver's non-mutating live commitment snapshot and compares the actual settled amount, not the cheap quote price.
- Commitment rotation/snapshot locking is sound; no context/cache downgrade bypass was found.
- The focused local floor suite passed: `cargo test price_floor --lib --features test-utils` — **7/7**.
- Current GitHub checks are all green across builds, platform tests, clippy, fmt, docs, audit, ADR validation and no-logging tests.

### Rollout gates before any canary enforcement

1. **Make shadow evidence observable.** `ant_node::payment::price_floor` is an `info!` event, but normal startup installs no subscriber unless `--enable-logging` / `ANT_ENABLE_LOGGING` is set, and no-logging builds cannot emit it. Confirm a logging-enabled artefact and service configuration before calling shadow collection active.
2. **Make the signal queryable.** The values are currently embedded in one high-volume formatted string containing the full XOR name. Prefer structured low-cardinality fields/counters/histograms (separate `ClientPut` and `FreshReplication`) or publish and test the exact collector parser/query.
3. **Run the live settlement suite with retained evidence.** `poc_price_floor_live` genuinely covers a real Anvil payment, production `completedPayments` read and real commitment state, but no PR workflow executes it. Add one Ubuntu CI invocation, or run and retain equivalent evidence, before enforcement. It deliberately bypasses live DHT issuer locality and production startup/network wiring, so describe it as a focused vault/floor integration test rather than a full network E2E.
4. **Soften the 4-of-20 claim.** Compatible clients require four acceptances and can fall back across up to twenty discovered targets; this gives headroom but does not guarantee success when fewer targets are usable or enforcing nodes are XOR-local. Replace “cannot fail honest uploads” with a conditional statement and define the maximum canary count.
5. **Document rollback accurately.** The environment is read once at startup. The kill switch is: unset `ANT_PRICE_FLOOR_ENFORCE`, restart the named canary, then positively verify shadow mode. Unsetting alone does nothing to a running node.
6. Define the shadow window, sample size, rejection bound, canary count, stop condition and sign-off owner in ADR-0006 before moving it from Proposed or widening enforcement. “~0” is not an operational threshold.

### Minor, non-blocking notes

- The arithmetic computes `(local_price × tolerance / 100) × 3`; the written formula can be read as `local_price × tolerance × 3 / 100`. The difference is at most two atomic units, but code/ADR/tests should name one canonical rounding rule.
- Present-but-non-Unicode tolerance values are treated as absent, and unrecognised enforcement values silently select shadow mode. Availability-safe, but an explicit effective-config startup signal would make operator mistakes visible.
- The existing process-global environment test would be safer behind a static mutex.

**Review action:** comment only — no changes requested, no approval/merge action. Human review is still required for the Proposed ADR and enforcement policy.

mickvandijke (approved):

jacderida (approved):

PR #181, opened 2026-07-22
Replace full-chunk audit responses with BLAKE3 verified slices

grumbach:
## Linear issue

- https://linear.app/autonominetwork/issue/V2-685 (replace full-chunk byte audit responses with BLAKE3 verified slices)

## Risk tier

This changes the possession-proof representation itself. A weak slice proof would let under-storing nodes pass audits and keep holder credit, so it sits squarely in T3.

## Compatibility

- Wire: changed for the audit message families only; core replication is unchanged and is **not** partitioned. Details below.
- Storage: none. No on-disk format change; commitments are rebuilt from the existing store.
- API: additive. New `replication::slice` module and public helpers; metric field names deliberately held stable.

**Wire detail.** The four subtree-audit bodies move to `SUBTREE_AUDIT_PROTOCOL_ID` (`autonomi.ant.replication.subtree-audit.v1`) and the `AuditChallenge`/`AuditResponse` pair to `POSSESSION_AUDIT_PROTOCOL_ID` (`autonomi.ant.replication.possession-audit.v2`). Round 1's `SubtreeLeaf` carries `content_len + nonced_root` instead of a flat `nonced_hash`; round 2's `SubtreeByteChallenge/Response` become `SubtreeSliceChallenge/Response`. A receive guard drops any body arriving on the wrong family id, and responses are routed by the same rule.

Core replication stays on `autonomi.ant.replication.v2`, so fresh replication, neighbour sync, verification, fetch, repair and commitment fetch keep working across versions for the whole upgrade window. Only cross-version *audits* pause — see Mitigation / rollback for the trust-accounting effect during that window.

**API detail.** `SubtreeLeaf`'s field changes are within an already-public replication type. The metric field names (`subtree_byte_*`, `audit_dropped_byte`) are held stable on purpose: they are what the V2-685 testnet queried to compare a 0.15.0 cohort against this branch with no field mapping, and renaming them would silently zero existing dashboards.

## Semver impact

- [ ] breaking
- [x] feature
- [ ] fix

Stated precisely, since a review pass raised it: the Rust API *is* source-breaking for a library consumer. `ReplicationMessageBody::SubtreeByteChallenge/Response` are renamed to `...Slice...` and `SubtreeLeaf` changes fields, and both are publicly reachable through `ant_node::replication::protocol`. On a `0.x` crate that is a minor bump, which is what "feature" delivers, so the box is right either way. Nothing outside this repo consumes these types today — the only callers are the crate's own tests — so no deprecated aliases are provided.

## Test evidence

**Testnet (jacderida, 990 nodes / 66 VMs, DEV-01, 07-27): 9/9 acceptance criteria pass.** Full data in the testnet registry, row 472.

The mixed-version cohort gave a same-network control — both cohorts in one network, at the same time, under the same client load, differing only in the audit protocol version:

| cohort | responses | mean round-2 response |
|---|---|---|
| DEV-02 (v2, released 0.15.0) | 14,197 | **6.19 MB** |
| DEV-01 (v3, this branch) | 69,903 | **14.49 KB** |

**About 427x smaller, with every confound held constant.** The v2 figure independently reproduces production, measured at 5.79–6.15 MB/response on `ant-prod-01` across four separate days.

Other criteria: corrupted node caught by 3 independent auditors in +29 min (consistent with the 30-minute per-peer cooldown); zero false positives (`digest_mismatch`/`key_absent`/`malformed` all 0 across ~43k audits over 5.5 h); 4,049/4,049 uploads and 797/797 downloads; 736/736 and 135/135 mixed-version transfers with zero failures; all 7 v2 hosts grew both `chunks.mdb` and `paid_list.mdb`, i.e. the v2 cohort kept accepting paid stores from v3 peers throughout; round-1 CPU 41.3% vs 45.7/45.2 baseline and RSS 908 MB vs 916/901 — no measurable regression despite round 1 now building a nonced Merkle tree over all block bytes; audit-timeout rate 1.57/service-hour vs run 470's ~178.5; 990/990 services up, 0 restarts, 0 failed units (both baselines showed 2/990 and 5/990 panics on 0.15.0).

Detection was validated with a construction that zeroes chunk bytes **in place with no restart**, so the node keeps its identity, uptime and already-gossiped commitment while its bytes no longer hash to their addresses. Deleting `chunks.mdb` and restarting is an invalid test — the node rebuilds its commitment from the empty store and then honestly declares holding nothing, which is correct behaviour, not cheating.

**What the fleet run does and does not cover.** It ran at `78cc1b0`, and the branch has moved since. The measurements that carry — round-2 response size, round-1 CPU and RSS, detection of an in-place corrupted node, mixed-version transfer success — are all properties of the proof shape, which has not changed since. What landed *after* the run and therefore has no fleet evidence: the three digest lanes moving to their own protocol id (with `GRACE_POSSESSION_AUDIT_TIMEOUTS` covering the resulting cross-version silence), the pre-decode ceiling now classifying every inbound replication message by its own discriminant, the responder-wide round-1 work budget being charged on every exit path, and round 2 revoking commitment-scoped credit on silence. The first two touch paths every message crosses, so a short re-run on the final head before this rides a train is the honest close — happy to spec it narrowly.

Add the two-slot round-1 admission pool to that list: it also post-dates the run, and it is the tightest bound the responder applies. Set too low it capacity-drops honest auditors into the graced timeout lane, so audit coverage falls with no peer penalised and nothing in the logs to show it. That is the metric pair worth watching on the re-run — `audit_dropped_byte` against the audit pass rate — and it is now `subtree_round1_max_concurrent` on `ReplicationConfig` rather than a compiled-in constant, so a fleet that lands on the wrong number is retuned in place instead of waiting for another release.

**Local (on the current head, rebased onto `main` at 0.16.0):**
- `cfd` clean: clippy `-D warnings`, fmt, `cargo doc --deny=warnings`.
- 832 lib tests green, including `replication::slice`, `storage_commitment_audit`, the round-2 credit and work-budget tests, the family message ceilings, the rollout dialect selection, and the round-1 pool's config plumbing.
- `poc_commitment_audit_attacks` 19/19: relay-holds-addresses, predict-and-fetch under fresh sampling, fabricated-fraction, deleter, substitution / replay / structural.
- `poc_audit_handler_live` 14/14: the production responder against real LMDB, including a deep-block opening of a multi-block chunk verified on both chains, responder coalescing under duplicate and interleaved openings, and the round-1 work charge on a served proof, a mid-run rejected read, and a first-leaf failure.
- `tests/e2e` audit + prune 10/10: real multi-node QUIC + LMDB. Honest node passes, data-deleting node is caught, repeated audits produce no false positives.
- `tests/e2e` replication 36/36 on the final head, re-run after the changes below (32 min, real QUIC + LMDB).

**Rollout compatibility.** A peer on the previous release sends its possession challenges on the core protocol id and converts an unanswered one into an audit-severity failure. Those challenges are answered rather than dropped, on the id they arrived on and in the digest construction that peer verifies with, so the release does not penalise the nodes that take it. That is a temporary weakening for old-auditor pairings only, bounded and never below what is deployed today, and it is removed together with `GRACE_POSSESSION_AUDIT_TIMEOUTS`. ADR-0009 records both halves and the criterion.

## New dependency

`bao = "0.13.1"` — Bao verified-slice encoding over BLAKE3. Shares the `blake3` crate already in the tree. This is the mechanism that makes a 1 KiB opening verifiable against a chunk's existing content address, and is what the egress reduction rests on. Needs explicit acknowledgement at review.

## ADR

https://github.com/WithAutonomi/ant-node/blob/v2-685-blake3-verified-slice-audit/docs/adr/ADR-0009-audit-proof-shape-and-protocol-families.md — ADR-0009: Audit proof shape and protocol families.

Numbered 0009 rather than 0007: `main` took 0007 for the Windows LMDB map head-room decision while this branch was in review.

ADR-0002 is left byte-for-byte as it was: it records what was decided then. ADR-0009 records what replaces its round-2 proof shape, the decision to unify freshness derivation across all four audit lanes, and the decision to give each changed message family its own protocol id. Every ADR already on `main` is byte-identical here; this branch adds one file and touches no other.

`docs/REPLICATION_DESIGN.md` is updated in the same commit, since it is the normative document and still mandated the flat freshness digest this branch replaces.

## Mitigation / rollback

Revert the branch. No stored data changes, so a rollback needs no migration: nodes rebuild commitments from their existing store either way. Because core replication never left `autonomi.ant.replication.v2`, a partially-rolled-back fleet keeps replicating normally — the only effect is that audits pause between mismatched versions, landing in the timeout lane which `GRACE_POSSESSION_AUDIT_TIMEOUTS` holds penalty-free for the window. Audit rate and admission caps are config constants, so audit load can also be reduced without a code change.

**Keep the mixed-version window short.** Stated because it is a real operational constraint and not obvious from the grace gate's name. Gracing a timeout stops silence being penalised, but it cannot create holder credit: credit is written only by a completed two-round subtree audit and expires after 40 minutes, and the subtree lane has no legacy accommodation because its round-1 leaf changed shape. So across a version boundary neither side refreshes the other's credit, and once the entry expires a cross-version peer counts as uncredited. The effect is bounded to the presence quorum, where its `Present` vote downgrades to `Unresolved` and the key verifies through the paid-list path or is retried; pruning does not consult holder credit and is unaffected. The direction is conservative, so the cost is extra verification traffic and a reward-eligibility gap for cross-version holders, not lost data.

It is deliberately not suppressed. Waiving the downgrade for the window would count `Present` votes from peers that had proven nothing, which can let a false claim stand in for a replica and suppress a repair — a conservative failure traded for an unsafe one. ADR-0009 records the reasoning, and the note also sits on the gate constant so whoever removes it sees both halves. The gap closes on its own as the fleet converges, at the same point the follow-up below unblocks.

**Follow-up owed:** remove `GRACE_POSSESSION_AUDIT_TIMEOUTS` (set `false`, delete the constant, inline the guarded branches) once the fleet is past the upgrade window, restoring the timeout penalty on the three digest lanes. ADR-0009 now records the owner and an objective criterion for that: at least 99% of nodes seen in the routing table over a seven-day window on the release carrying this change, and no supported release still on the old audit protocol in the field. While the gate is set the suppression applies to every peer, not only old ones, so this should stay open rather than be closed with this PR.

---

## What this changes and why

Round 2 of the storage-commitment audit made the audited node return the **complete original bytes** of the nonce-sampled chunks, up to two full 4 MiB chunks per response. This replaces that with a **verified slice**: a few KB per opened 1 KiB block, while proving *more* than the old check.

Framing the benefit correctly, since the original justification has aged: the "~4.2 TB/day" figure predates 0.15.0. Querying V2-623 counters on `ant-prod-01`, `subtree_byte_response` was 5.98 TB/day on 07-22 and 32.3 TB/day on 07-24, then fell to 0.46 and 0.22 TB/day on 07-26/27 as the fleet auto-upgraded to 0.15.0. Response *size* held at 5.8–6.2 MB throughout; only the *rate* moved. So the incremental saving on today's fleet is nearer 0.2–0.5 TB/day than 4.

The stronger argument is that this **bounds the cost per audit event regardless of rate**, which frequency caps cannot do. On 07-24 the audit rate spiked and cost 32 TB/day at 6 MB per response; after this change that same spike costs ~32 GB.

### How

A chunk's address is `BLAKE3(content)`, and BLAKE3 is internally a Merkle tree over 1 KiB blocks, so a **Bao verified slice** proves a block is the real content at a given offset against the address the auditor already knows — content authenticity for free. But the address is public, so authenticity alone is delegable: a node storing nothing could pass by fetching the block on demand from an honest holder. Possession must be bound separately.

So round 1 additionally commits, per leaf, a fresh **nonced block tree**: a Merkle root over the same 1 KiB blocks whose leaves are **keyed** BLAKE3 hashes, `keyed_BLAKE3(key = f(nonce ‖ peer ‖ key), block_index ‖ len ‖ block)`. The nonce is fed as the BLAKE3 *key*, not a message prefix, so it mixes into every chunk compression of every leaf — there is no nonce-independent chaining value to precompute. The auditor draws the block to open with fresh randomness **after** the roots are committed (cut-and-choose), and round 2 verifies both chains over the *same* block bytes.

A responder cannot fold an after-the-fact-fetched block to a garbage committed root without a preimage break.

### What a pass proves — stated honestly

**Against the realistic threat this keeps security close to what `main` gives, at roughly 1/430 of the egress.** It is a sampling check, and the claim should be read that way rather than as a proof of complete possession.

The auditor holds none of the audited bytes, so it cannot tell a correct `nonced_root` from a responder-chosen one. A peer retaining a fraction `p` of a chunk's blocks passes whenever every fresh draw lands on a kept block — roughly `p^leaves` per audit over the 3–5 sampled leaves. The mandatory final-block opening anchors the claimed length but adds no detection, since a partial holder simply keeps the final block. The full-byte round 2 this replaces caught any missing byte of a sampled chunk with certainty.

That difference is inherent: verifying every byte means moving every byte. Stating the comparison in both directions:

- Against a node under-storing **in bulk** — the realistic case, a node dropping data to save disk — detection is close to what the full-byte audit gave, at ~1/430 of the egress. The fleet run caught a 256 MB in-place corruption on the first audit that reached the node, by three independent auditors, against zero false positives across ~43,000 audits in the preceding 5.5 hours.
- Against a **fine-grained** partial deleter, one shaving a little off every chunk, this is strictly weaker per audit than serving every byte. The compensating lever is audit *frequency*, which is what the cost reduction buys — the old shape made frequent auditing unaffordable.

`partial_holder_passes_on_kept_blocks_and_fails_on_dropped_ones` in `replication::slice` makes the property executable rather than asserted.

For `p^leaves` to describe *detection* rather than merely a pass, it also has to cost something not to answer — see the next section.

### Round 2 changes: silence is no longer free

Round 2 names the blocks to open only after the round-1 roots are committed, which is what makes the sample a cut-and-choose. The same ordering means the responder sees the draw before deciding whether to reply. Silence was fully graced and left holder credit in place, so an unfavourable draw cost nothing while credit from an earlier favourable one survived. The published probability then describes an answered pass rather than detection.

An unanswered round 2 following a valid round-1 proof now revokes the holder credit carried by the commitment under audit — the same treatment an explicit `Transient` reject already had. It stays in the graced timeout lane and takes no trust penalty, and the revocation is scoped to the pinned commitment, so credit re-earned under a newer commitment survives. An honest peer that drops a reply keeps its standing and re-earns credit on its next audit; a peer on the older protocol cannot reach this state, because it never completes the new round 1.

The decision is a small predicate (`round2_revokes_pinned_credit`) rather than an inline match arm, so it is pinned by a direct test, with a companion test on the scoping.

### Responder work bounds

Round 1 had a two-slot pool and a per-peer responder cooldown. Both are keyed by peer identity or by concurrency, and neither bounds *sustained* work: a slot refills the moment it frees, and a per-peer cooldown refills by presenting a different peer id.

Round 1 is now also charged against a responder-wide budget over the chunk bytes it reads and hashes, refilled at a fixed rate and keyed by nothing at all. The balance carries debt, so an expensive proof is admitted proportionally less often than a cheap one and the sustained rate settles at the refill rate whatever the caller's identity. It is sized above honest demand even at the largest commitment the system allows, since starving honest audits would cost more than it saves — at a mid-sized commitment the headroom over measured honest load is more than an order of magnitude.

Charging happens after the proof is built, because the cost is not known until the pinned commitment has been resolved and its subtree selected. Only the proof path needs charging: every rejection a caller can provoke returns before a chunk is read, and the two paths that do read first turn on this node's own storage rather than on anything the caller sends.

### Bounding work before admission

Family, session and admission checks all read fields of a *decoded* message, so none of them can run before decoding. That made decoding the first thing an unknown peer could demand, under a 10 MiB ceiling sized for hint batches that no audit body carries.

The encoded length needs no parsing, so the audit families now take their own wire ceiling, checked before decode and sized against the largest legitimate audit body — the round-1 proof at the commitment key-count cap. A test builds that worst case and pins the headroom, so a later change to leaf size or commitment encoding cannot quietly start dropping honest proofs.

### Binding the content length

`content_len` and `nonced_root` travel in the round-1 leaf but are **not** covered by the signed commitment root (it hashes only `key ‖ bytes_hash`), so a malicious responder can claim any length. Unchecked, that forges possession two ways: a deflated length collapses the block count so only block 0 is ever opened, and Bao does not bind the slice length header for a prefix slice that never reaches EOF.

Both are closed on the auditor side: `verify_block_slice` requires the slice's own length header to equal the claimed `content_len`, and the auditor also opens the **claimed final block** per sampled leaf. The final-block decode reaches EOF, where Bao authenticates the encoded length against the address, so a forged-short length fails there. Signing `content_len` into the commitment would not help — the attacker would just commit the false length from the start; only anchoring it against the address does.

Both verifiers now also reject a block index past the end of the chunk. Nothing the auditor generates is ever out of range, so this is about failing cleanly rather than a live hole: the block range clamps an over-large index to an empty range, which would otherwise read back as a verified empty block.

### Unifying the audit lanes

There are **four** lanes that ask a peer to prove it holds bytes: the subtree audit, plus three digest-based callers — the periodic responsible-chunk audit, the post-replication possession probe, and prune confirmation. The digest lanes derived their freshness by prefixing the challenge material to the hashed stream, which binds the nonce to the leading bytes rather than the whole content. All lanes now use one construction: derive a BLAKE3 key from the challenge material under a versioned domain-separation context, then hash the content under that key.

That changes what an older peer computes for the same bytes. Left on the shared core id, a cross-version exchange would read as a *confirmed* digest mismatch and penalise an honest peer for the whole upgrade window — hence the dedicated protocol ids described under Compatibility.

Also: the responsible-chunk audit no longer reports a pass when a challenged key's own local copy is unreadable. That key cannot be checked, so the tick yields no verdict rather than a pass reflecting only the readable keys.

## Why not the literal proposal

The ticket's original wording (verify the slice **only** against the chunk address, with a nonce-derived offset) is unsound: it drops the possession binding entirely, so a storage-less relay farms audit passes. That defeats proof-of-storage and lets one warehouse back many replica identities, breaking replication multiplicity. The offset must also be **fresh post-proof randomness**, not nonce-derived, or the cut-and-choose collapses. This PR implements the corrected dual-chain design.

## Commits

1. `Replace full-chunk audit responses with BLAKE3 verified slices` — the dual-chain slice audit.
2. `Authenticate slice-audit content_len to defeat possession forgery` — the content-length binding (header check + final-block open).
3. `Harden slice audit and isolate its protocol id` — dedicated subtree-audit protocol family.
4. `Address audit review: resource bounds, geometry, contract` — admission caps, canonical sibling depth, response contract.
5. `Tidy slice-audit comments and lock responder coalescing with a live test` — coalesced/order-independent round-2 contract, locked by a live responder test.
6. `Close two audit holes: key-content forgery and round-1 work amplification` — `bytes_hash == key` enforcement and round-1 work bounds.
7. `Bind every audit lane to the fresh nonce and isolate the audit protocols` — the unified keyed digest and the possession-audit protocol id.
8. `Take the responder permit before consuming the round-2 session` — admission ordering fix so a transient capacity drop no longer permanently kills a round-2 exchange.
9. `Lock the round-2 accounting boundary with direct tests` — round-2 classification as a pure function, tested at the production boundary.
10. `Grace audit timeouts through the rollout, and state the proof's real strength` — the rollout gate, and the corrected strength claim.
11. `Grace only silence in the prune audit lane` — split "no answer" from "undecodable answer" at the source.
12. `State the strength comparison in both directions` — both-directions framing in the docs.
13. `Cite the PR and issue in the ADR's Related field`.
14. `Renumber the audit ADR and align the replication design doc` — ADR-0009, corrected lane count, grace-removal criterion and owner, `REPLICATION_DESIGN` brought into line.
15. `Cost an unanswered round 2 the pinned commitment's credit`.
16. `Bound audit responder work by rate and by message size`.
17. `Reject out-of-range block openings, and restate the egress figures`.
18. `Separate admitted absence from failed proofs, and make the round-1 pool tunable` — the two pre-release diagnostics/tunability fixes, and the mixed-version credit-gap note.

## Related

Supersedes #186, which pursued the same egress goal by having the auditor verify co-held leaves against its own copy. Its keyed-digest and Idle-on-unreadable changes are carried here (commit 7); its local-verify optimisation is not, because once round 2 is 14.49 KB the auditor-side cost of reading and hashing its own 4 MiB chunk exceeds the wire bytes it would save. Reasoning recorded on that PR.

The ADR-0008 payment-economics document that briefly appeared on this branch has been dropped: it belongs to #188, which has since merged to `main`. This branch now carries no ADR content other than its own.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2433,7 +2433,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -3124,7 +3124,7 @@ dependencies = [
  "js-sys",
  "log",
  "wasm-bindgen",
- "windows-core 0.62.2",
+ "windows-core",
 ]
 
 [[package]]
@@ -4334,7 +4334,7 @@ dependencies = [
  "once_cell",
  "socket2 0.6.4",
  "tracing",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -4347,7 +4347,7 @@ dependencies = [
  "libc",
  "socket2 0.6.4",
  "tracing",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -4689,9 +4689,9 @@ dependencies = [
 
 [[package]]
 name = "ruint"
-version = "1.19.0"
+version = "1.20.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "45caf26f647c19115bf9c453c70ffe4a4a3a6390dceebd942610584f99b8ddce"
+checksum = "f5e99bff0393163bb25029a6af25d3d8d202ba5b5438a74d1bd8789f5c822970"
 dependencies = [
  "alloy-rlp",
  "ark-ff 0.3.0",
@@ -4777,7 +4777,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -4844,7 +4844,7 @@ dependencies = [
  "security-framework",
  "security-framework-sys",
  "webpki-root-certs",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -5697,7 +5697,7 @@ dependencies = [
  "getrandom 0.4.3",
  "once_cell",
  "rustix",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -6418,7 +6418,7 @@ version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c2a7b1c03c876122aa43f3020e6c3c3ee5c05081c9a00739faf7503aeba10d22"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -6433,7 +6433,7 @@ version = "0.58.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dd04d41d93c4992d421894c18c8b43496aa748dd4c081bac0dc93eb0489272b6"
 dependencies = [
- "windows-core 0.58.0",
+ "windows-core",
  "windows-targets 0.52.6",
 ]
 
@@ -6443,24 +6443,11 @@ version = "0.58.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6ba6d44ec8c2591c134257ce647b7ea6b20335bf6379a27dac5f1641fcf59f99"
 dependencies = [
- "windows-implement 0.58.0",
- "windows-interface 0.58.0",
- "windows-result 0.2.0",
- "windows-strings 0.1.0",
+ "windows-implement",
+ "windows-interface",
+ "windows-result",
+ "windows-strings",
  "windows-targets 0.52.6",
-]
-
-[[package]]
-name = "windows-core"
-version = "0.62.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b8e83a14d34d0623b51dce9581199302a221863196a1dde71a7663a4c2be9deb"
-dependencies = [
- "windows-implement 0.60.2",
- "windows-interface 0.59.3",
- "windows-link",
- "windows-result 0.4.1",
- "windows-strings 0.5.1",
 ]
 
 [[package]]
@@ -6475,32 +6462,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "windows-implement"
-version = "0.60.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "053e2e040ab57b9dc951b72c264860db7eb3b0200ba345b4e4c3b14f67855ddf"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn 2.0.118",
-]
-
-[[package]]
 name = "windows-interface"
 version = "0.58.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "053c4c462dc91d3b1504c6fe5a726dd15e216ba718e84a0e46a88fbe5ded3515"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn 2.0.118",
-]
-
-[[package]]
-name = "windows-interface"
-version = "0.59.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3f316c4a2570ba26bbec722032c4099d8c8bc095efccdc15688708623367e358"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -6523,31 +6488,13 @@ dependencies = [
 ]
 
 [[package]]
-name = "windows-result"
-version = "0.4.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7781fa89eaf60850ac3d2da7af8e5242a5ea78d1a11c49bf2910bb5a73853eb5"
-dependencies = [
- "windows-link",
-]
-
-[[package]]
 name = "windows-strings"
 version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4cd9b125c486025df0eabcb585e62173c6c9eddcec5d117d3b6e8c30e2ee4d10"
 dependencies = [
- "windows-result 0.2.0",
+ "windows-result",
  "windows-targets 0.52.6",
-]
-
-[[package]]
-name = "windows-strings"
-version = "0.5.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7837d08f69c77cf6b07689544538e017c1bfcf57e34b4c0ff58e6c2cd3b37091"
-dependencies = [
- "windows-link",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -813,6 +813,7 @@ version = "0.16.0"
 dependencies = [
  "alloy",
  "ant-protocol",
+ "bao",
  "blake3",
  "bytes",
  "chrono",
@@ -1320,6 +1321,16 @@ dependencies = [
  "object",
  "rustc-demangle",
  "windows-link",
+]
+
+[[package]]
+name = "bao"
+version = "0.13.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3f9125f93587b894c196d58b0a752ce2213552d0d483c98ee41530e38202a511"
+dependencies = [
+ "arrayvec",
+ "blake3",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -111,6 +111,7 @@ page_size = "0.6"
 
 # Protocol serialization
 postcard = { version = "1.1.3", features = ["use-std"] }
+bao = "0.13.1"
 
 [target.'cfg(unix)'.dependencies]
 libc = "0.2"
--- a/docs/REPLICATION_DESIGN.md
+++ b/docs/REPLICATION_DESIGN.md
@@ -461,6 +461,8 @@ Rules:
 
 ## 15. Storage Audit Protocol (Anti-Outsourcing)
 
+Protocol families: the audit message families do not ride the core replication protocol id. The digest-based lanes described in this section (responsible-chunk audit, post-replication possession probe, prune confirmation) share one possession-audit id, and the gossip-triggered subtree audit has its own. Core replication keeps its id unchanged, so an audit-protocol change never partitions replication across a mixed-version fleet. A receive guard MUST drop any message whose family disagrees with the id it arrived on, and responses MUST be sent on the same id the guard would accept them on. See ADR-0009.
+
 Challenge-response for claimed holders:
 
 1. Challenger creates unique challenge id + nonce.
@@ -470,7 +472,7 @@ Challenge-response for claimed holders:
 5. If `PeerKeySet` is empty, the audit tick is idle.
 6. Challenger sends `challenged_peer_id` an ordered challenge key set equal to `PeerKeySet(challenged_peer_id)`.
 7. Target responds with either per-key `AuditKeyDigest` values or a bootstrapping claim:
-    a. Per-key digests: for each challenged key `K_i` (in challenge order), target computes `AuditKeyDigest(K_i) = H(nonce || challenged_peer_id || K_i || record_bytes_i)`, where `record_bytes_i` is the full raw bytes of the record for `K_i`. Target returns the ordered list of per-key digests. If the target does not hold a challenged key, it MUST signal absence for that position (e.g., a sentinel/empty digest); it MUST NOT omit the position silently.
+    a. Per-key digests: for each challenged key `K_i` (in challenge order), target computes `AuditKeyDigest(K_i) = keyed_H(key = derive_key(context, nonce || challenged_peer_id || K_i), record_bytes_i)`, where `record_bytes_i` is the full raw bytes of the record for `K_i`, `derive_key` is BLAKE3 key-derivation mode and `context` is a versioned domain-separation string. Target returns the ordered list of per-key digests. If the target does not hold a challenged key, it MUST signal absence for that position (e.g., a sentinel/empty digest); it MUST NOT omit the position silently.
     b. Bootstrapping claim: target asserts it is still bootstrapping. Challenger applies the bootstrap-claim grace logic (Section 6.2 rule 4b): record `BootstrapClaimFirstSeen` if first observation, accept without penalty within the one-time `BOOTSTRAP_CLAIM_GRACE_PERIOD`, emit `BootstrapClaimAbuse` evidence if past grace period or if this is a repeated claim after the peer previously stopped claiming bootstrap. Audit tick ends (no digest verification).
 8. On per-key digest response, challenger recomputes the expected `AuditKeyDigest(K_i)` for each challenged key from local copies and verifies equality per key before deadline. Each key is independently classified as passed (digest matches) or failed (mismatch, absent, or malformed).
 9. On any per-key audit failures (timeout, malformed response, or one or more `AuditKeyDigest` mismatches/absences), challenger MUST perform a responsibility confirmation for each failed key before emitting penalty evidence:
@@ -483,7 +485,7 @@ Audit-proof requirements:
 
 1. Challenger MUST hold a local copy of each challenged record to recompute per-key digests. Audit selection is therefore limited to records the challenger stores.
 2. Records are opaque bytes for replication; digest construction MUST operate over raw record bytes (no schema dependency) and be deterministic.
-3. Each `AuditKeyDigest(K_i)` input MUST be exactly: `H(nonce || challenged_peer_id || K_i || record_bytes_i)`. Including `K_i` binds each digest to its specific key and prevents digest reordering attacks.
+3. Each `AuditKeyDigest(K_i)` MUST be computed as `keyed_H(key = derive_key(context, nonce || challenged_peer_id || K_i), record_bytes_i)`. Including `K_i` in the derived key binds each digest to its specific key and prevents digest reordering attacks. The challenge material MUST enter as the hash key, not as a prefix to the hashed stream: a prefix binds only the leading bytes of the record, leaving the remainder independent of the nonce, whereas a derived key makes every compression of the record depend on it.
 4. Each `AuditKeyDigest` MUST include full record bytes; key-only digests are invalid.
 5. Nodes that advertise audit support MUST produce valid responses within `AUDIT_RESPONSE_TIMEOUT`.
 6. Responses MUST include exactly one digest entry per challenged key in challenge order. A response is invalid if it has fewer or more entries than challenged keys.
@@ -495,7 +497,8 @@ Audit challenge bound:
 
 Failure conditions:
 
-- Timeout, malformed response, or per-key `AuditKeyDigest` mismatch/absence — subject to responsibility confirmation (step 9) before penalty.
+- Malformed response or per-key `AuditKeyDigest` mismatch/absence — subject to responsibility confirmation (step 9) before penalty.
+- Timeout — also subject to responsibility confirmation, but see the rollout gate below: for as long as a protocol-family change is rolling out across the fleet, the digest-based lanes MUST NOT penalise a timeout, because a peer on the older family cannot answer at all and its silence is not its fault. A response that arrives is judged on what it says throughout: mismatch, absence, malformed reply and explicit rejection are all penalised while the gate is set. The gate MUST be removed once the fleet has upgraded (see ADR-0009 for the criterion and owner), restoring the timeout penalty on these lanes.
 - Bootstrapping claim past `BOOTSTRAP_CLAIM_GRACE_PERIOD`, or repeated after the peer previously stopped claiming bootstrap (emits `BootstrapClaimAbuse`, not `AuditFailure`).
 
 Audit trigger and target selection:
--- a/docs/adr/ADR-0002-gossip-triggered-contiguous-subtree-audit.md
+++ b/docs/adr/ADR-0002-gossip-triggered-contiguous-subtree-audit.md
@@ -90,8 +90,10 @@ tree is a clean binary shape when N is not a power of two.
 ## Decision
 
 We will make the audit **gossip-triggered** and replace its proof shape with a
-**single contiguous-subtree storage proof**, reusing the existing tree,
-commitment, and freshness-hash primitives.
+**single contiguous-subtree storage proof**, reusing the existing tree and
+signed commitment. Possession is proven by a keyed nonced block-tree opened by
+Bao verified slices (per the V2-685 amendment above), not by per-leaf freshness
+hashes.
 
 - **Trigger.** When a node ingests a neighbour's commitment during normal
   (steady-state) operation, it may start an audit of that neighbour — not every
@@ -233,7 +235,7 @@ commitment, and freshness-hash primitives.
 - **Big-block deletion is caught only proportionally.** An attacker who deletes data in large contiguous blocks is caught, per audit, with probability roughly equal to the fraction deleted — independent of N and of subtree size. We accept this: there is no economic reason to delete a *small* fraction (you save almost nothing and are still eventually caught), and a node that deletes a large fraction to actually save resources is caught within one or two audits. If ever needed, the lever is auditing *more often*, not bigger subtrees.
 - **Inflating the claimed size is not fully prevented.** Only the selected subtree and the path summaries are verified each audit, so filler leaves elsewhere could inflate the claimed chunk count. Both the regular audits and the closeness check mitigates this over time. Fully auditing the entire claimed set would be too much effort. We accept this probabilistic approach in which over time cheaters are detected. 
 - **Retention has a storage cost.** A node must keep the chunk data behind its recently published commitments for a bounded TTL window (with a slot backstop). This is an accepted cost.
-- **The audit format change is breaking.** The whole network must upgrade before the new audit can be relied on and before eviction is enabled.
+- **The audit format change is contained, not whole-network breaking.** Only the subtree-audit messages move to a dedicated protocol id; core replication stays on v2, so mixed-version nodes still interoperate for everything else. During rollout a cross-version subtree audit may go unanswered — a bounded, self-healing trust effect (see the `SUBTREE_AUDIT_PROTOCOL_ID` note), strictly milder than a global protocol bump. Eviction can be relied on once the fleet has upgraded.
 
 ### Neutral / Operational
 
--- a/docs/adr/ADR-0002-gossip-triggered-contiguous-subtree-audit.md
+++ b/docs/adr/ADR-0002-gossip-triggered-contiguous-subtree-audit.md
@@ -1,12 +1,21 @@
 # ADR-0002: Gossip-triggered contiguous-subtree storage audit
 
 - **Status:** Proposed
-- **Date:** 2026-06-04
+- **Date:** 2026-06-04 (round-2 proof shape updated 2026-07 for the V2-685 verified-slice audit)
 - **Decision owners:** Anselme (@grumbach)
 - **Reviewers:** <pending>
 - **Supersedes:** none
 - **Superseded by:** none
-- **Related:** none
+- **Related:** ADR-0004 (audit grace / Amendment 1)
+
+> **Amendment (V2-685, 2026-07).** Round 2 no longer returns full chunk bytes: it
+> returns a few-KB **BLAKE3/Bao verified slice** per opened block, checked against
+> the chunk address (authenticity) and a keyed **nonced block-tree root** committed
+> in round 1 (possession). Possession is now a round-1 cryptographic commitment
+> rather than a deadline property; per-leaf "freshness hashes" become keyed nonced
+> roots; retention is TTL-bounded (not "last two"); the subtree audit rides a
+> dedicated protocol id; and round 1 gains responder rate/session/heavy-pool
+> controls. The Decision and Consequences below reflect this.
 
 ## Context
 
@@ -46,7 +55,7 @@ tree is a clean binary shape when N is not a power of two.
 - Ensure all nodes actually store the data they claim they are storing
 - Keep each proof small and keep steady-state audit traffic low.
 - Catch the three real cheating strategies: storing nothing and fetching on demand; deleting some fraction of data; and keeping only chunk *addresses* (which are public) while never holding the actual bytes, then fabricating proofs.
-- Reuse the existing cryptographic building blocks (the Merkle tree, the signed commitment, the freshness hash) without inventing new ones.
+- Reuse the existing cryptographic building blocks (the Merkle tree, the signed commitment, and a per-leaf keyed nonced block-tree root) without inventing new ones.
 - Never wrongly penalise honest nodes, even in extreme cases like on small or dense networks where every node legitimately holds almost all of the data.
 
 ## Considered Options
@@ -115,53 +124,67 @@ commitment, and freshness-hash primitives.
   branch is exactly the one the random value selects and that it contains at least
   the square root of the claimed held chunks in real leaves.
 
-- **The proof.** The audited node returns every leaf of the selected subtree —
-  each given both as the plain content hash and as a freshness hash (the content
-  mixed with the auditor's random value) — plus one summary hash per level for the
-  unselected siblings along the path to the root. Everything outside the selected
-  branch costs a single hash; nothing there is touched.
+- **The proof (round 1).** The audited node returns every leaf of the selected
+  subtree — each given as the plain content hash (the chunk address), its content
+  length, and a fresh **nonced block-tree root**: a Merkle root over the chunk's
+  1 KiB blocks whose leaves are *keyed* BLAKE3 over each block, keyed by the fresh
+  audit nonce (plus peer and key). It also returns one summary hash per level for
+  the unselected siblings along the path to the root. Everything outside the
+  selected branch costs a single hash; nothing there is touched. (This replaces
+  the earlier flat per-leaf "freshness hash": keying every block by the nonce
+  leaves no nonce-independent chaining value to precompute, so building a correct
+  root requires *all* of a chunk's bytes under that nonce.)
 
 - **Verification, three independent checks.**
   - *Structure:* rebuild the root from the returned subtree and the sibling
     summaries; it must equal the freshly-published root the audit was started
     against. This proves the subtree genuinely belongs to the committed tree.
-  - *Real bytes:* after the full subtree proof is in hand, pick a small fixed
-    number of its leaves and demand the original chunk bytes for exactly those
-    keys from the audited node itself (a second-round surprise challenge), then
-    confirm both the plain hash (the chunk's content address) and the freshness
-    hash match the served bytes. The sample is drawn with **fresh randomness
-    chosen by the auditor after round 1 — NOT derived from the round-1 nonce**.
-    This is essential: the structural root check binds only `(key, bytes_hash)`,
-    both of which are public (the `bytes_hash` *is* the chunk's network address),
-    not the per-leaf freshness hash. If the sample were predictable at
-    proof-build time, a relay could fabricate the freshness hash on every leaf,
-    fetch only the few leaves it knew would be opened, and pass while holding
-    almost nothing. Drawing the sample after the proof commits turns this into a
-    cut-and-choose: the node must have produced a correct freshness hash — which
-    needs the real bytes — for essentially every leaf, or be caught. Possession
-    is non-delegable: the auditor needs to hold none of the node's chunks, and a
-    committed key the node cannot serve is a deterministic failure, never bad
-    luck. So a node that rebuilt the tree from public chunk addresses but never
-    held the bytes cannot serve content that hashes to the committed address;
-    faking a fraction of leaves survives only with probability (1 − fraction)
-    raised to the number of spot-checks.
-  - *Possession in time:* the whole response must arrive within a deadline sized
-    to hashing the subtree from local disk. A node that doesn't hold the data must
-    fetch it across the network first and misses the deadline.
+  - *Possession by verified slice (round 2):* after the full subtree proof is in
+    hand, pick a small fixed number of its leaves and, for each, a **freshly-random
+    1 KiB block index** (plus the claimed final block, which pins the true content
+    length via Bao's EOF check). Demand a **verified slice** per opened block, not
+    the whole chunk: a Bao slice proving the block against the chunk address
+    (authenticity, ~KB not MB), plus a nonced-tree opening proving it against the
+    round-1 `nonced_root` (possession), both over the same served bytes. The sample
+    is drawn with **fresh randomness chosen by the auditor after round 1 — NOT
+    derived from the round-1 nonce**: the structural root binds only
+    `(key, bytes_hash)`, both public, so a predictable sample would let a relay
+    fabricate a `nonced_root` on every leaf and hold only the few it knew would be
+    opened. Drawing the sample after the proof commits makes this a cut-and-choose;
+    a node that never held the bytes cannot have committed a correct nonced root,
+    and faking a fraction of blocks survives only with probability (1 − fraction)
+    raised to the number of spot-checks. Possession is non-delegable and the
+    auditor holds none of the node's chunks.
+  - *Deadline (liveness, not the possession proof):* the response must still
+    arrive within a deadline, but — unlike the earlier full-byte round 2 — the
+    deadline is no longer load-bearing for possession (that is now the round-1
+    `nonced_root` commitment). It bounds liveness and how long the auditor waits;
+    a missed deadline is graced (see Accounting).
 
 - **Retention — "you stay answerable for what you publish."** A node keeps the
-  chunk data behind its **last two published commitments**. Two, not one, absorbs
-  the normal race where an auditor is asking about the commitment a node published
-  just before its newest one. Because of this, an honest node can always answer an
-  audit about a commitment it published recently — so "I don't recognise that
-  commitment" about a recently-published root is now provably misbehaviour, not
-  lag.
+  chunk data behind its recently published commitments for a **bounded TTL
+  window** (with a slot backstop), so an honest node can always answer an audit
+  about a commitment it published recently — meaning "I don't recognise that
+  commitment" about a recently-published root is provably misbehaviour, not lag.
+
+- **Responder resource controls.** The heavy round 1 (which hashes the whole
+  `sqrt(N)` subtree, up to gigabytes for a maximal commitment) runs on a blocking
+  thread pool with its own tight admission (a small global + one per peer),
+  separate from the light responsible/slice audits, plus a per-peer rate cooldown;
+  a round-2 slice challenge is served only after a matching single-use round-1
+  session. This keeps a burst of audit requests from starving honest audits or the
+  async runtime, without changing the audit's security properties.
+
+- **Rollout.** Only the subtree-audit messages changed wire format, so they ride a
+  dedicated protocol id while core replication stays on its existing id; a
+  mixed-version fleet keeps replicating, syncing, fetching and repairing across
+  versions, and only cross-version subtree audits pause during the upgrade window.
 
 - **Accounting and False Positives** "That chunk isn't in my commitment" 
   can never occur, because the auditor only ever challenges leaves of the node's
   *own* committed tree, so every challenged leaf is in the commitment by
   construction. Failures that are deterministic and cannot be caused by bad luck — a
-  rebuilt root that doesn't match, a content or freshness hash that doesn't match,
+  rebuilt root that doesn't match, a content-address or nonced-root check that doesn't match,
   or repudiating a recently-published commitment — are acted on **the first time
   they occur**, because re-asking cannot turn a genuine failure into a pass.
   Failures that *can* be caused by transient bad luck — a missed response deadline
@@ -202,19 +225,19 @@ commitment, and freshness-hash primitives.
 - The probabilistic approach to verification ensures that verification is cheap but over time efficient. 
 - Each proof is small and contiguous (about the square root of N leaves plus a handful of summary hashes) instead of many scattered inclusion paths.
 - Audits are surprise exams pinned to the *freshly published* commitment, so there is no stale-data ambiguity unlike in the previous audit design
-- Three independent defences cover the three cheating strategies: structure (belongs to the committed tree), real bytes (actually held, not fabricated from public addresses), and timeliness (held locally, not fetched on demand).
+- Independent defences cover the cheating strategies: structure (belongs to the committed tree), possession (a keyed nonced-root commitment in round 1, opened by verified slices in round 2, so it cannot be fabricated from public addresses), with the response deadline bounding liveness rather than carrying the possession proof.
 - Acting on the first deterministic failure roughly cuts time-to-detection compared with requiring several strikes, with no added risk of false positives.
 
 ### Negative / Trade-offs
 
 - **Big-block deletion is caught only proportionally.** An attacker who deletes data in large contiguous blocks is caught, per audit, with probability roughly equal to the fraction deleted — independent of N and of subtree size. We accept this: there is no economic reason to delete a *small* fraction (you save almost nothing and are still eventually caught), and a node that deletes a large fraction to actually save resources is caught within one or two audits. If ever needed, the lever is auditing *more often*, not bigger subtrees.
 - **Inflating the claimed size is not fully prevented.** Only the selected subtree and the path summaries are verified each audit, so filler leaves elsewhere could inflate the claimed chunk count. Both the regular audits and the closeness check mitigates this over time. Fully auditing the entire claimed set would be too much effort. We accept this probabilistic approach in which over time cheaters are detected. 
-- **Retention has a storage cost.** A node must keep the chunk data behind its last two published commitments. This is an accepted cost. 
+- **Retention has a storage cost.** A node must keep the chunk data behind its recently published commitments for a bounded TTL window (with a slot backstop). This is an accepted cost.
 - **The audit format change is breaking.** The whole network must upgrade before the new audit can be relied on and before eviction is enabled.
 
 ### Neutral / Operational
 
-- Introduces a few tunable settings: the per-gossip audit probability, the per-neighbour cooldown, the number of real-byte spot-checks, and the retention count (two). The grace allowance for missed deadlines reuses the existing strike threshold and applies to deadline misses only.
+- Introduces a few tunable settings: the per-gossip audit probability, the per-neighbour cooldown, the number of block spot-checks per audit, the retention TTL window, and the heavy round-1 responder pool / rate cooldown / session TTL. The grace allowance for missed deadlines reuses the existing strike threshold and applies to deadline misses only.
 - The storage-commitment audit needs no periodic timer of its own — it is driven by gossip. (The separate responsible-chunk audit keeps its periodic tick; the two run side by side.) The related "node is capable but has no current commitment" special case is unnecessary on the gossip-triggered path, since that path always has a freshly-published commitment to pin. A silent node needs no special handling for this audit — it simply stops earning storage credit, so all nodes are naturally motivated to gossip. 
 - At the chosen settings, steady-state audit load is on the order of a handful of small audits per node per hour.
 
@@ -234,15 +257,16 @@ How we will know this decision remains correct:
   and identical on the auditor and the audited node; selection never lands on an
   all-padding branch across many awkward sizes (a regression test for the
   fixed-depth flaw this ADR fixes); the root rebuilds correctly from a single-branch
-  proof; possession verifies from the bytes the audited node itself serves in the
-  second-round byte challenge (the auditor holding none of them); a committed key
-  the node cannot serve is a deterministic failure; the real-byte spot-check catches a node that fabricated
-  freshness hashes, at the expected probability; deterministic failures are acted on
+  proof; possession verifies from the verified slices the audited node itself
+  serves in round 2 (the auditor holding none of the bytes); a committed key
+  the node cannot serve is a deterministic failure; the slice spot-check catches a
+  node that fabricated a `nonced_root` or under-stored, at the expected
+  probability; deterministic failures are acted on
   the first time while deadline misses honour the grace allowance; the adaptive
   timeout grace responds to widespread timeouts but never to deterministic failures;
-  repudiating a recently-published commitment fails; the last two published
-  commitments stay answerable; the response deadline is sized correctly; and a flood
-  of gossip does not multiply audits.
+  repudiating a recently-published commitment fails; recently published
+  commitments stay answerable within the TTL window; the response deadline is sized
+  correctly; and a flood of gossip does not multiply audits.
 
 - **Operational signals and re-open triggers.** Audits per node per hour stay within
   budget; false-positive penalties on a small, dense test network stay at zero
--- a/docs/adr/ADR-0002-gossip-triggered-contiguous-subtree-audit.md
+++ b/docs/adr/ADR-0002-gossip-triggered-contiguous-subtree-audit.md
@@ -1,21 +1,12 @@
 # ADR-0002: Gossip-triggered contiguous-subtree storage audit
 
 - **Status:** Proposed
-- **Date:** 2026-06-04 (round-2 proof shape updated 2026-07 for the V2-685 verified-slice audit)
+- **Date:** 2026-06-04
 - **Decision owners:** Anselme (@grumbach)
 - **Reviewers:** <pending>
 - **Supersedes:** none
 - **Superseded by:** none
-- **Related:** ADR-0004 (audit grace / Amendment 1)
-
-> **Amendment (V2-685, 2026-07).** Round 2 no longer returns full chunk bytes: it
-> returns a few-KB **BLAKE3/Bao verified slice** per opened block, checked against
-> the chunk address (authenticity) and a keyed **nonced block-tree root** committed
-> in round 1 (possession). Possession is now a round-1 cryptographic commitment
-> rather than a deadline property; per-leaf "freshness hashes" become keyed nonced
-> roots; retention is TTL-bounded (not "last two"); the subtree audit rides a
-> dedicated protocol id; and round 1 gains responder rate/session/heavy-pool
-> controls. The Decision and Consequences below reflect this.
+- **Related:** none
 
 ## Context
 
@@ -55,7 +46,7 @@ tree is a clean binary shape when N is not a power of two.
 - Ensure all nodes actually store the data they claim they are storing
 - Keep each proof small and keep steady-state audit traffic low.
 - Catch the three real cheating strategies: storing nothing and fetching on demand; deleting some fraction of data; and keeping only chunk *addresses* (which are public) while never holding the actual bytes, then fabricating proofs.
-- Reuse the existing cryptographic building blocks (the Merkle tree, the signed commitment, and a per-leaf keyed nonced block-tree root) without inventing new ones.
+- Reuse the existing cryptographic building blocks (the Merkle tree, the signed commitment, the freshness hash) without inventing new ones.
 - Never wrongly penalise honest nodes, even in extreme cases like on small or dense networks where every node legitimately holds almost all of the data.
 
 ## Considered Options
@@ -90,10 +81,8 @@ tree is a clean binary shape when N is not a power of two.
 ## Decision
 
 We will make the audit **gossip-triggered** and replace its proof shape with a
-**single contiguous-subtree storage proof**, reusing the existing tree and
-signed commitment. Possession is proven by a keyed nonced block-tree opened by
-Bao verified slices (per the V2-685 amendment above), not by per-leaf freshness
-hashes.
+**single contiguous-subtree storage proof**, reusing the existing tree,
+commitment, and freshness-hash primitives.
 
 - **Trigger.** When a node ingests a neighbour's commitment during normal
   (steady-state) operation, it may start an audit of that neighbour — not every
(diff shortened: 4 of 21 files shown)


dirvine (changes requested):
Reviewed exact head `bae97e9c867b5b1e3b00159cf316ef68a0708eaf` with a six-seat panel (two groups of three), then independently reproduced the blocking cryptographic finding and ran the focused checks locally.

**Verdict: changes requested.** The bandwidth direction is good, and the Bao length-forgery fix is well tested, but the current construction does not yet establish the possession property claimed by the PR.

## Blocking findings

### 1. High — nonce-independent BLAKE3 CV still permits deterministic under-storage

`src/replication/slice.rs:51,99-115`

The leaf input is:

`domain(38) || nonce(32) || peer(32) || key(32) || index(4) || len(4) || block(1024)`

The 142-byte prefix plus the first 882 block bytes fill BLAKE3 chunk 0. The final 142 block bytes form chunk 1 and are nonce-independent. A node can retain that chunk's 32-byte chaining value instead of the 142 original bytes, saving **110 bytes per full block / 10.7421875%**, while reconstructing the exact round-one leaf/root for every fresh nonce.

I reproduced this against the production layout using BLAKE3's low-level subtree API: `merge_subtrees_root(chunk0_cv, precomputed_chunk1_cv)` exactly matched `blake3::hash(prefix || block)`.

Round two still needs the missing suffixes, but at most ten openings require only about 1.4 KiB of missing data. The generous slice deadline therefore becomes security-relevant again, contrary to the comments claiming the preprocessing gap is closed and all bytes are required.

A fresh nonce-derived BLAKE3 key (`new_keyed(nonce)` or a domain-separated key derived from the nonce) makes every BLAKE3 chunk nonce-dependent and closes this specific attack. Merely prefixing the nonce does not.

### 2. High — round-two requests amplify full-chunk reads and synchronous hashing

`src/replication/storage_commitment_audit.rs:1148-1203`; `src/replication/slice.rs:144-207,248-263`; `src/replication/mod.rs:2527-2594`

Each opening independently:

1. reads the full chunk from LMDB;
2. creates a full Bao outboard and clones the full content;
3. rebuilds the complete nonced tree.

The normal random-plus-final pair does this twice for the same key. Duplicate openings are accepted, so one ten-opening request can cause up to 40 MiB of reads plus repeated full scans/copies. Sixteen globally admitted responder tasks bound concurrency, but there is no rate/work budget or binding of round two to responder-held round-one state; repeated requests can keep the pool and Tokio workers busy.

Please coalesce by key, read each chunk once, reuse Bao/nonced tree state for all openings, reject/coalesce duplicates, and add byte/work-aware admission or bind the request to live round-one state. The CPU-heavy construction should not run unbounded on Tokio worker threads.

### 3. Medium — correlated mid-audit `Bootstrapping` preserves holder credit

`src/replication/storage_commitment_audit.rs:351-357,782-785`; `src/replication/mod.rs:4471-4507`

After serving a valid round-one proof, a responder can answer round two with a matching `Bootstrapping` response. It is mapped to `Timeout`; subtree timeouts are fully graced and preserve existing `recent_provers` credit. This responsive contradiction bypasses both confirmed-failure handling and bootstrap-claim-abuse accounting.

Treat it as a protocol failure, or at minimum revoke the pinned commitment's credit and route it through bootstrap-claim accounting.

### 4. Rollout gate — v3 partitions all replication operations

`src/replication/config.rs:247-262`; `src/replication/mod.rs:1237-1257`

The single global protocol ID is used for fresh replication, neighbour sync, verification, fetch, repair/pruning/possession audits and commitment fetches—not just the changed subtree audit. During a staggered v2/v3 rollout, DHT routing still mixes cohorts while cross-version peers exchange no replication traffic at all.

This needs either a dedicated/dual-registered audit protocol or a quantified and tested coordinated-rollout plan. The current test only asserts the string constant; it does not test mixed-version bootstrap, fresh replication or repair.

## Design documentation

ADR-0002 is unchanged and still specifies full chunk responses, flat freshness hashes and deadline-based possession. Several code comments also still describe full bytes/one opening or incorrectly imply the Bao header check alone authenticates prefix-slice length. These are material security-invariant mismatches, not wording nits; please update the ADR and comments with the implementation.

## Verified locally

- `cargo test -q replication::slice --lib --no-fail-fast` — 15/15 pass
- `cargo test -q storage_commitment_audit --lib --no-fail-fast` — 16/16 pass
- `cargo test -q --features test-utils --test poc_commitment_audit_attacks` — 19/19 pass
- `cargo test -q --features test-utils --test poc_audit_handler_live` — 9/9 pass
- `cargo clippy --all-targets --all-features -- -D warnings` — pass
- Max-chunk measurement: Bao slice 1,800 bytes + 12 nonced siblings/384 bytes per opening; about 22 KiB for ten openings before small envelope overhead.

## Review-team consensus

All six seats completed. Both cryptographic reviewers independently reproduced the BLAKE3 preprocessing gap. Both runtime reviewers confirmed duplicate full-chunk work amplification. Both rollout/accounting passes confirmed the global v3 split and round-two bootstrap-credit path. No dissent on the blockers; the main qualification is that the CV attack still needs retrieval/delegation for the few opened suffixes.

CI is green except the Windows test job is still pending at review time.

dirvine (changes requested):
Reviewed exact head `712b4c994ecb04522d357f6330f7e398a5a13e2d` (incremental from `bae97e9c867b5b1e3b00159cf316ef68a0708eaf`).

**Verdict: changes still requested.** The concrete CV attack, round-two duplicate work, responsive-bootstrap credit path, and global protocol partition are fixed. Two material items from the prior review remain, plus focused regression gaps.

## Resolved

- **Nonce-independent suffix CV:** closed. `nonced_block_key` binds domain/nonce/peer/key and every leaf uses keyed BLAKE3, so every BLAKE3 chunk compression is nonce-dependent (`src/replication/slice.rs:60-83,128-142`).
- **Round-two repeated work:** substantially fixed. Openings are coalesced per key, the chunk is read once, Bao/nonced state is reused, and CPU work runs through `spawn_blocking` (`src/replication/storage_commitment_audit.rs:1079-1131,1207-1298`).
- **Responsive round-two `Bootstrapping`:** now revokes the pinned commitment's credit and becomes a confirmed rejection (`src/replication/storage_commitment_audit.rs:360-367,796-808`).
- **Rollout partition:** core replication remains on v2; the four subtree-audit bodies use a dedicated protocol ID with symmetric body/ID guards (`src/replication/config.rs:245-282`; `src/replication/mod.rs:175-208,2342-2358,3360-3377`).

## Remaining blockers

### 1. High — expensive audit work is concurrency-bounded, but still not rate- or session-bounded

`src/replication/config.rs:132-158`; `src/replication/mod.rs:2242-2309`; `src/replication/storage_commitment_audit.rs:995-1067,1171-1298`

The 16-global/4-per-peer admission caps prevent unbounded fan-out, but a peer can continuously refill its four slots; a few identities can keep the shared pool saturated. Round two is still accepted from any source presenting a retained commitment hash, without a short-lived/single-use responder-side round-one session.

Round one is the heavier path and still builds each leaf synchronously on Tokio workers after the LMDB read. At the maximum legal subtree hint, a request may process 1,000 × 4 MiB = **3.906 GiB** of chunk content; four in-flight requests from one peer are **15.625 GiB**, and sixteen globally are **62.5 GiB** of worst-case concurrent work. This can starve honest audits and turn dropped honest requests into timeout/failure accounting.

Please add a per-peer rate/work budget, bind round two to a live single-use round-one exchange, and move/bound round-one hashing off Tokio workers. Separate budgets for the much heavier subtree round one versus responsible/slice audits would also avoid one work class consuming the whole shared pool.

### 2. High — ADR-0002 still specifies a different protocol

`docs/adr/ADR-0002-gossip-triggered-contiguous-subtree-audit.md:118-150,152-158,208-218,233-245`

The ADR remains unchanged and still specifies flat freshness hashes, original full-chunk responses, deadline-led possession, “last two” retention, and a generic whole-network breaking rollout. The implementation now uses `content_len + nonced_root`, keyed block leaves, Bao random+final openings, TTL-based retained gossiped commitments, and a dedicated subtree protocol with bounded mixed-version trust effects.

Several source comments also retain removed names/semantics, including `nonced_hash`, `verify_byte_response`, full-byte serving, and one opening per leaf (`src/replication/storage_commitment_audit.rs:49-53,478-487,1134-1141`; `src/replication/subtree.rs:235-242`; `src/replication/protocol.rs:892-895`). For a possession protocol, this is normative design drift rather than a wording nit. Please update ADR-0002 and the stale comments in this PR.

### 3. Medium — the fixed production paths lack direct regressions

There are no focused tests driving:

- responder coalescing/one-read/one-build behavior and duplicate/interleaved requests;
- `ResponsiveBootstrap` through scoped credit revocation and confirmed-failure classification;
- an actual old-v2 subtree wire fixture being dropped on the core ID;
- dedicated request/response routing and correlation.

The helper tests are useful, but these were explicit prior findings and should be locked at the production-path boundary. Also reconcile the declared response contract (“one item per requested opening, in requested order”, `src/replication/protocol.rs:927-942`) with grouped/deduplicated emission (`src/replication/storage_commitment_audit.rs:1207-1245`): either restore original order/cardinality after reusing work, or document and test the coalesced order-independent contract.

## Non-blocking hardening

`verify_nonced_block` accepts any sibling-vector depth that folds to `nonced_root` (`src/replication/slice.rs:254-273`). This does not recreate the deterministic under-storage attack—the root precedes fresh sampling—but enforcing the canonical depth from `block_count(content_len)` (including odd-tail self-pair semantics) would bind the claimed left-packed tree geometry and reject malformed work early.

## Verification

Locally on this exact head:

- `cargo test -q replication::slice --lib --no-fail-fast` — 16/16 pass
- `cargo test -q storage_commitment_audit --lib --no-fail-fast` — 16/16 pass
- `cargo test -q --features test-utils --test poc_commitment_audit_attacks` — 19/19 pass
- `cargo test -q --features test-utils --test poc_audit_handler_live` — 9/9 pass
- `git diff --check` — pass

Current-head CI: builds, format, clippy, docs, security audit, and no-logging tests pass; Ubuntu/macOS/Windows test jobs remain pending.

## Review-team result

Five delegated seats plus independent verification reviewed the exact SHA. The cryptographic/adversarial seats agree the concrete CV attack is closed; runtime/release seats agree the rate/session bound and ADR drift remain blockers. The only dissent was severity of canonical sibling-depth validation (low/non-blocking). A required direct Z.AI GLM-5.2 seat was attempted repeatedly but the direct API returned HTTP 429, so this is not represented as full six-model consensus.

dirvine:
Follow-up review on head `712b4c9`: the keyed-BLAKE3, round-two coalescing, responsive-bootstrap, and dedicated-protocol fixes are confirmed. Changes remain requested for sustained audit work lacking rate/session bounds, unchanged ADR-0002/design comments, and direct regression coverage of the fixed production paths. Full review: https://github.com/WithAutonomi/ant-node/pull/181#pullrequestreview-4761709519

jacderida:
## Testnet evaluation — 990-node DEV run, 9 acceptance criteria, all pass

Ran this branch (`78cc1b0`, = PR head) on a full-scale testnet against the V2-685 evidence plan. **9/9 criteria pass.** Summary below; full data is recorded in the testnet registry as row 472.

| | |
|---|---|
| Treatment | DEV-01, 990 nodes / 66 VMs, 7 bootstraps, 10 uploaders + 2 downloaders, 30% NAT, 4 providers |
| Control cohort | DEV-02, 105 nodes on released **0.15.0** (v2 audit), extension sharing DEV-01's bootstraps |
| Profile | `staging-short` @ `47165ba` — identical to baseline runs 470 and 471 (both ant-node 0.15.0) |
| Phases | T0 = 2026-07-27T14:44:35Z · warm-up 1h · **measurement 15:44:35–18:44:35Z** · mixed-version 18:51:22–19:51:22Z |

### Results

| # | Criterion | Verdict | Evidence |
|---|---|---|---|
| 1 | Round-2 response size | **PASS** | **14.49 KB** mean over 69,903 responses |
| 2 | Corrupted node detected | **PASS** | 3 independent auditors, `digest_mismatch`, +29 min |
| 3 | No false positives | **PASS** | `digest_mismatch`/`key_absent`/`malformed` all **0** across ~43k audits / 5.5h |
| 4 | Transfers | **PASS** | 4,049/4,049 uploads, 797/797 downloads = 100.000% |
| 5 | Mixed-version transfers | **PASS** | 736/736 and 135/135 during the phase, zero failures |
| 6 | v2 nodes storing | **PASS** | all 7 v2 hosts grew `chunks.mdb` **and** `paid_list.mdb` |
| 7 | Round-1 cost | **PASS** | CPU 41.3% vs 45.7 / 45.2 baseline; RSS 908 MB vs 916 / 901 |
| 8 | Audit-timeout rate | **PASS** | 1.57 per service-hour vs run 470's ~178.5 |
| 9 | Node stability | **PASS** | 990/990 services + 7/7 bootstraps, 0 restarts, 0 failed units |

### The size reduction, with a same-network control

The mixed-version cohort turned out to be the cleanest possible measurement. Both cohorts ran **in one network, at the same time, under the same client load**, differing only in the subtree-audit protocol version:

| cohort | responses | mean round-2 response |
|---|---|---|
| DEV-02 (v2, released 0.15.0) | 14,197 | **6.19 MB** |
| DEV-01 (v3, this branch) | 69,903 | **14.49 KB** |

**417× reduction** with every confound held constant. The v2 figure independently reproduces what production is currently doing — I measured `subtree_byte_response` on `ant-prod-01` at **5.79–6.15 MB/response**, stable across four separate days, before the run.

14.49 KB rather than the ~6 KB implied by "~1000x" is expected and correct: round 2 now opens two blocks per leaf with two hash chains each.

### Round 1 costs nothing measurable

This was the regression I most expected, since the responder still reads every byte of each opened chunk off disk (per the `audit_response_timeout` sizing comment) while round 1 now builds a nonced Merkle tree over all block bytes rather than the old flat `nonced_hash`. On 15 services per VM, CPU came out **~9% below** both baselines and per-service RSS was flat. No cost visible at this scale.

### Detection parity — note the test-design trap

My first attempt was **an invalid test**, and it is worth recording so nobody repeats it: deleting `chunks.mdb` and restarting the service lets the node **rebuild its commitment from the empty store**, so it then honestly declares holding nothing. No detection occurs, and that is the correct behaviour — the node isn't cheating, it's just smaller.

The valid construction zeroes chunk bytes **in place with no restart**, so the node keeps its identity, uptime and already-gossiped commitment while its bytes no longer hash to their addresses. (LMDB is mmap'd, so deleting the file under a live process achieves nothing — it keeps serving from the unlinked inode.)

Done that way: 256 MB zeroed at 20:26:29Z → caught by **three independent auditors** at 20:55:31Z (+29 min, consistent with the 30-min per-peer cooldown), against a counter that had read **0 fleet-wide across ~43,000 round-2 responses over the preceding 5.5 hours**. That is the Bao authenticity chain doing exactly what it is designed to do.

### What this run does *not* establish

Node egress fell to **34.6 GB per GB uploaded** (baselines: 82.0 and 85.7) on aligned windows, while moving more data. **I am not attributing that to this PR.** Round-2 audit egress this run was 1.04 GB out of 3,644 GB of total replication egress — **0.03%**. At the old ~6 MB size those same responses would have been ~256 GB, so this PR accounts for roughly **6%** of a ~4 TB gap. This branch carries **25 commits of main beyond v0.15.0** (audit scheduler, cooldown and admission fixes) that runs 470/471 predate, which is the more likely explanation. The same caveat applies to criterion 8's timeout improvement.

### Three notes on the PR itself

1. **The PR description is stale in one important respect.** It says `REPLICATION_PROTOCOL_ID` advances to v3 and that "v2 and v3 nodes simply don't exchange replication traffic". At `78cc1b0` the subtree audit rides its own `SUBTREE_AUDIT_PROTOCOL_ID`, so core replication is *not* partitioned. The run confirms the current behaviour: all 7 v2 hosts grew `paid_list.mdb` as well as `chunks.mdb`, i.e. the v2 cohort was accepting **paid** stores from v3 peers throughout, with zero transfer failures. Worth updating the description — the current text understates the change.

2. **The "~4.2 TB/day out" figure predates 0.15.0.** Querying the V2-623 counters on `ant-prod-01`: `subtree_byte_response` was 5.98 TB/day on 07-22 and 32.3 TB/day on 07-24, then fell to 0.46 TB/day on 07-26 and 0.22 TB/day on 07-27 — the collapse coincides with the fleet auto-upgrading to 0.15.0. Response *size* held at 5.8–6.2 MB throughout; only the *rate* moved. So the incremental saving on today's fleet is nearer 0.2–0.5 TB/day than 4. That is not an argument against this PR — arguably the reverse: on 07-24 the audit rate spiked and cost 32 TB/day at 6 MB per response; this change makes that same spike cost ~32 GB. It bounds the cost per event regardless of rate, which frequency caps alone don't. Might be worth reframing the justification that way, since the headline number invites a "we already fixed that" response.

3. **Minor naming inconsistency.** `SubtreeByteResponse` → `SubtreeSliceResponse` in the enum, but `log_traffic_summary` still emits `subtree_byte_response_*` at variant index 14. Convenient for us — it made the counters directly comparable across 0.15.0 and this branch with no field mapping — but the name is now misleading.

### Method notes

- Counters are cumulative from process start, so figures are per-`(host, service)` `max − min`; that undercounts across a restart and cannot overcount. There were no restarts.
- Whole-run deltas are used in preference to timestamp-ranged queries because the profile includes a deliberately +3h clock-skewed VM that a range filter displaces.
- Criterion 9 was checked via systemd, not Elasticsearch — panics reach journald only, so ES-based crash queries read clean regardless.
- Stability was better than both baselines, which showed 2/990 and 5/990 `JoinHandle polled after completion` panics on 0.15.0. This run had none.

grumbach:
## Review round: what changed at `1d40303`

Rebased onto `main` (clean, no conflicts), then worked through @dirvine's open items, the Copilot findings, @jacderida's testnet notes, and three independent adversarial review passes. Summary of what moved.

### Two claims in this PR were wrong. Both are fixed.

**1. The digest lanes were not graced on timeout.** Moving the responsible-chunk audit, possession probe and prune confirmation onto `POSSESSION_AUDIT_PROTOCOL_ID` means a v3 auditor's challenge to a still-v2 peer is never answered and times out. The subtree lane graces timeouts (`handle_subtree_failed_audit` returns early); **these three did not** — they reported `ApplicationFailure(AUDIT_FAILURE_TRUST_WEIGHT)`, the weight-5 confirmed-failure event. Every honest peer across the version boundary would have taken that repeatedly for the entire upgrade window. The doc on the protocol-id constant claiming otherwise was simply false.

`GRACE_POSSESSION_AUDIT_TIMEOUTS` graces them for the window. It covers **silence only** — a peer that answers is judged on what it said, so digest mismatch, absent key, malformed reply and explicit rejection are all still penalised throughout. `rollout_gate_graces_only_timeouts` pins that. The prune lane originally collapsed "no answer" and "undecodable answer" into one `None`; those are now split at the source so the gate cannot excuse garbage from a matched-version peer.

**This gate must be removed in a follow-up** — set `false`, delete the constant, inline the branches — once the fleet is past the window. While it is set, a peer silently dropping challenges is under-penalised (it still takes the upstream unit transport decrement, not the audit-severity one). The guarded branches stay compiled rather than commented out so they cannot rot, and the test flips with them.

**2. The proof-strength claim was overstated.** "Building the correct nonced root requires all of a chunk's bytes" is true but does not carry the weight it was given: a responder does not need the *correct* root. The auditor holds none of the audited bytes, so it cannot distinguish a correct root from a responder-chosen one. A peer keeping a fraction `p` of a chunk's blocks commits a root with genuine leaves for what it kept and arbitrary leaves for what it dropped, and passes whenever every draw lands on a kept block — about `p^leaves` per audit over the 3–5 sampled leaves. The mandatory final-block opening anchors the claimed length but adds no detection, since a partial holder keeps the final block.

So this is a sampling check. The comparison is stated in both directions, because a one-line summary was what made it misleading the first time:

- Against a node under-storing **in bulk** — the realistic case — detection is close to what the full-byte audit gave, at ~1/1000 of the egress. That is what the fleet run measured.
- Against a **fine-grained** partial deleter it is strictly weaker per audit. The compensating lever is audit *frequency*, which is what the cost reduction buys.

Two axes move in opposite directions and are now separated in the docs: *freshness* gets stronger (the nonce enters every block leaf, closing a preprocessing gap in the old flat `nonced_hash`), *coverage* gets weaker (sampling, not every byte). `partial_holder_passes_on_kept_blocks_and_fails_on_dropped_ones` builds the forged root and demonstrates both halves, so the property is executable rather than asserted.

### Other fixes

- **Response routing was asymmetric.** Possession `AuditResponse`s were sent on the *core* id while the receive guard required the possession id. It worked only because saorsa-core correlates RR replies by `(peer, msg_id)` rather than protocol name; a bare response would have been dropped. Send and receive now share one `response_protocol_for`, with a test asserting the id we send on is always one the guard accepts.
- **ADR-0007's "one freshness derivation" was not true.** The possession lane used BLAKE3 `derive_key` mode; the slice lane used a plain hash over a domain prefix, which separates domains only by convention. The slice lane now uses `derive_key` under its own context, with a test pinning that both lanes share the construction and derive different keys.
- **Round-2 admission ordering** (Copilot): the single-use session was consumed *before* responder admission, so a transient capacity drop destroyed the exchange. Admission now comes first; the regression test fails on the old order. Honest note: `request_slice_proof` issues one `send_request` and maps failure to `Timeout`, so there is no in-audit retry today — the ordering keeps a refusal truthful rather than recovering an audit, and costs nothing over the alternative.
- **`AuditDropKind::Byte` → `Slice`** at code level, with the emitted counter names (`audit_dropped_byte`, `subtree_byte_*`) deliberately pinned — those are the fields @jacderida queried to compare a 0.15.0 cohort against this branch with no field mapping, and renaming them would silently zero existing dashboards.
- **Three e2e audit tests** were timing out against the family guard, still sending on the core id. The helper now derives the id from the body, so a future family move cannot silently turn them into timeouts.
- **PR description** rewritten against the new template from #185, with @jacderida's three corrections applied: the v3-partitions-replication claim removed (that was #186's design), the ~4.2 TB/day figure reframed as bounding cost-per-event, and the commit list completed.

### @dirvine's open items, rechecked at this head

Your review was against `712b4c9`, which predates four of these commits. Rechecking:

- **Rate/session bounds** — in place: `MAX_CONCURRENT_SUBTREE_ROUND1 = 2` (separate tighter pool), `SUBTREE_ROUND1_RESPONDER_COOLDOWN = 30 min` per peer, `MAX_SUBTREE_ROUND1_PER_PEER`, bounded `MAX_SUBTREE_SESSIONS`, single-use sessions, round-1 hashing on `spawn_blocking`.
- **ADR drift** — ADR-0002 restored byte-for-byte; ADR-0007 records what replaced its round-2 proof shape. The remaining `nonced_hash` mentions are deliberate contrasts with the old design, not stale claims.
- **Canonical sibling depth** (your non-blocking item) — enforced in `verify_nonced_block` against `nonced_tree_depth(block_count)`.
- **Response contract** — reconciled: the coalesced, order-independent contract is documented on the type and locked by a live responder test.
- **Direct regressions** — the gap was `ResponsiveBootstrap`. The round-2 classification is now a pure `classify_slice_response`, tested directly: mid-audit `Bootstrapping` is a confirmed failure, and every arm is guarded on `challenge_id` so a responder cannot answer one challenge with a body minted for another.

### Deferred, with reasons

Raised by review, judged out of scope for this PR rather than dismissed:

- **A transient LMDB read error is reported as proven absence** (`audit.rs`, the `Err(e) => ABSENT_KEY_DIGEST` arm). An honest peer with a disk hiccup takes a confirmed failure. This is **pre-existing on `main`**, not introduced here, and fixing it needs a wire change — that message family has no `Transient` reply kind, unlike the new subtree handlers. Worth its own PR.
- **The responsible-audit key count has no absolute cap** (`2 * sqrt(stored_chunks)`), so a large store admits a large read/hash batch into the light pool. Also pre-existing sizing.
- **`SubtreeLeaf::bytes_hash` duplicates the content-addressed key**, and round 2 re-sends session data the responder already holds. Both are genuine simplifications that would shave further bytes, but both are wire changes and would re-open the crypto review this PR has already been through.
- **`poc_commitment_audit_attacks` reimplements the verifier** rather than calling it, so it can stay green while production drifts. Real concern; a test refactor, partly mitigated by the live-handler PoC and the real-QUIC e2e.

### Verification

`cfd` clean (clippy `-D warnings`, fmt, `cargo doc --deny=warnings`). 788 lib tests, `poc_commitment_audit_attacks` 19/19, `poc_audit_handler_live` 11/11, and the audit e2e 8/8 including the real-QUIC honest-passes / deleter-caught / no-false-positives set.

A final adversarial pass re-checked each fix against the updated diff and confirmed the response routing, the key derivation, and the round-2 admission ordering (no permit or per-peer slot leaks on any path; session consumed exactly once on success). It rejected two things, both now addressed: the prune lane still collapsed unanswered and undecodable into one case — split at the source, so the gate cannot excuse garbage — and the strength summary was still overstated in two headline sentences even though the detailed explanation was accurate. Those sentences are the ones rewritten above.

One correction to the record: an earlier automated pass reported this PR contained ~930 lines of unrelated CI and governance changes. That was an artifact of a stale local `main` reference. The real diff is 18 files, all replication code.

grumbach:
## CI note: the Windows e2e failure is a pre-existing flake on `main`

`Test (windows-latest)` is red, and it is **not** this branch. Recording the evidence so nobody re-diagnoses it.

**Signature.** Every occurrence is the same: exactly one test dies during testnet startup, never on an assertion.

```
setup: Testnet(Startup("Failed to create node N after 4 attempts: Transport error:
  Setup failed: Failed to create dual-stack network nodes: Failed to create transport"))
test result: FAILED. 87 passed; 1 failed; 3 ignored
```

**The victim is random, which is the tell.** Three runs, three different tests, one cause:

| run | victim test | node |
|---|---|---|
| this branch, 1st | `merkle_payment::test_attack_merkle_pay_yourself_fabricated_pool` | node 13 |
| this branch, rerun | `replication::test_paid_list_persistence` | node 0 |
| **`main`** (run 30068433100) | `replication::scenario_12_bootstrap_quorum_aggregation` | node 0 |

A code regression fails the same test every time. This picks a different one each run and always at transport creation, ~2,500 s into the job.

**It reproduces on `main`.** 2 of the last 8 `main` CI runs failed `Test (windows-latest)` with that exact string — roughly a 25% rate, independent of this PR.

**The audit paths pass on Windows.** In the failing run, all ten replication-audit and prune e2e tests passed, including the three protocol-id tests fixed here and the full `subtree_audit_testnet` set — honest node passes, data-deleting node is caught, repeated audits produce no false positives.

**Not caused by the one test-harness change here.** This branch touches `tests/e2e/testnet.rs` only to set `subtree_round1_responder_cooldown` to 1 ms on the SMALL preset, so repeated audits in tests are not rate-dropped. It changes no node count, port range or startup path.

### Worth fixing separately

This looks like a real resource ceiling rather than luck. The suite creates many nodes in one long-lived process, and `Failed to create node N after 4 attempts` means the retry budget was exhausted — the Windows runner is running out of sockets or ephemeral ports late in the run. It will keep hitting unrelated PRs until the e2e job either reuses transports, spreads across jobs, or widens the retry/backoff. Happy to raise it if it is not already tracked.

dirvine (changes requested):
Reviewed exact head `3802265dd5bbc3ecddf65e81f708530d0366597e` with five completed delegated review seats plus independent verification. One additional adversarial seat timed out and is not represented as completed consensus; the configured environment also had no available GLM-5.2 credential, so I am not claiming that seat.

**Verdict: changes requested.** The earlier nonce/CV, Bao length, proof-geometry, protocol-partition, duplicate-work and session-ordering findings are substantially fixed. The focused and full local suites are green. Two security/resource blockers remain, alongside architecture-document drift that should not merge in this T3 PR.

## Blocking findings

### 1. High — selective round-two silence turns failed samples into eventual passes

`src/replication/storage_commitment_audit.rs:849-886`; `src/replication/mod.rs:5626-5645,5917-5942`; `docs/adr/ADR-0007-audit-proof-shape-and-protocol-families.md:108-133`

The responder sees the freshly selected `(key, block_index)` openings after committing its round-one roots. A partial holder can answer only when every random opening lands on retained blocks and remain silent otherwise. Silence becomes `AuditFailureReason::Timeout`, is fully graced, and preserves existing holder credit.

That changes the documented `p^leaves` result materially: it is the probability of an answered pass, not detection. Unfavourable samples are inconclusive rather than failures. Repeating audits therefore lets a partial holder wait for a favourable sample; for example, retaining 75% of blocks gives a 23.7% favourable five-leaf draw and about a 95% chance of eventually passing within 12 attempts.

This is worse than the old full-chunk round two for fine-grained deletion: a partially deleted sampled chunk could never produce a valid full-byte response, whereas the new block sample eventually can.

Please make silence after a valid round-one proof security-relevant. At minimum revoke the pinned holder credit immediately. Prefer a bounded retry of the same challenge/session (with an idempotent cached response), followed by confirmed-failure accounting if the peer remains silent. A version-skewed peer cannot reach this state because it never completed the new round one.

### 2. High — task caps are not a global work/rate budget; Sybil identities can sustain multi-GiB round-one work

`src/replication/config.rs:162-197`; `src/replication/mod.rs:3125-3155,3558-3586`; `src/replication/storage_commitment_audit.rs:1101-1164`; `src/replication/subtree.rs:173-196`

The two-slot global semaphore and one-slot-per-`PeerId` cap bound concurrency, but the 30-minute cooldown is identity-scoped. Two rotating identities can continuously occupy both heavy slots, and the cooldown map eventually evicts old identities at 2,048 entries.

At the maximum 1,024-leaf subtree and 4 MiB chunks, one request reads about 4 GiB and hashes about 8 GiB of input (plain address plus nonced tree); both permits admit about 8 GiB read / 16 GiB hashed per wave. This can saturate disk and the blocking pool while honest audits are immediately dropped. The honest 990-node CPU/RSS observation does not exercise this adversarial pattern.

Please add a responder-wide byte/work token budget or rate window, not only a semaphore, and gate expensive challenges on a harder-to-Sybil eligible-auditor relation where possible. Preserve fair capacity for honest audits. Add an adversarial test/measurement for sustained distinct-identity requests at the maximum legal subtree.

### 3. Medium — attacker-controlled collection decoding occurs before audit admission

`src/replication/protocol.rs:65-90,1012-1029`; `src/replication/mod.rs:3368-3412,3658-3721`

A sessionless peer can send a near-10-MiB postcard `SubtreeSliceChallenge.openings` vector. The serial replication handler fully allocates/decodes it before protocol-family/session/admission checks; the ten-opening cap is enforced only afterwards. Responder semaphores do not bound this path.

Use a fixed-capacity/bounded wire representation, a much smaller subtree-family message cap, or header-first per-peer byte/rate admission before decoding the collection.

## Architecture and scope blockers

- **Remove unrelated ADR-0008 from this PR.** `docs/adr/ADR-0008-storage-economics-and-payment-protocol.md` is the subject of separate PR #188 and the copy in #181 has already diverged from that PR's corrected text. Merging it here would land an inaccurate payment ADR and conflict with #188.
- **Update the current normative replication design.** `docs/REPLICATION_DESIGN.md:473,486` still mandates flat `H(nonce || peer || key || bytes)`, while this PR intentionally ships derive-key + keyed BLAKE3 (`src/replication/protocol.rs:1114-1147`). The same section still describes timeout as unconditionally penalised and does not record the dedicated possession family.
- **Correct ADR-0007's lane count.** `ADR-0007:21-30,70-78` lists three lanes but omits periodic responsible-chunk audit. There are four logical callers: subtree plus three digest-based callers (responsible, post-replication possession, prune confirmation).
- **Track removal of `GRACE_POSSESSION_AUDIT_TIMEOUTS`.** Link a dedicated issue/owner and an objective fleet-upgrade criterion, or explicitly keep V2-685 open. While enabled, all three digest callers under-penalise deliberate silence. Add direct tests for possession timeout/garbage and prune unanswered/malformed; the current helper test covers only the responsible-audit policy branch.

## Non-blocking corrections

- The displayed `6.19 MB / 14.49 KB` means imply about **427×** in decimal units, not 417×; the ADR's “roughly a thousandth” is also unsupported by those figures. Quote raw-byte aggregates or use approximately 1/400–1/430.
- `verify_block_slice` / `verify_nonced_block` should reject `index >= block_count` explicitly. Production-generated requests are in range, so I did not find an exploit from this alone.
- Refresh the stale commit/test-count prose and comments that still say possession stays on the core protocol or that timeout is always penalised.

## Verified locally on this head

- `git diff --check origin/main...HEAD` — pass
- `cargo test -q replication::slice --lib --no-fail-fast` — 19/19 pass
- `cargo test -q storage_commitment_audit --lib --no-fail-fast` — 20/20 pass
- `cargo test -q --features test-utils --test poc_commitment_audit_attacks` — 19/19 pass
- `cargo test -q --features test-utils --test poc_audit_handler_live` — 11/11 pass
- `cargo test -q --lib --no-fail-fast` — 788/788 pass

Current CI passes all completed jobs; Windows tests are still pending. The 990-node figures are detailed author/testnet-reported evidence, but I could not fetch the internal Slack/registry raw evidence from this environment.

The cryptographic construction now looks materially stronger and the egress direction is good. The selective-silence path is the main proof-soundness blocker: without closing it, the sampling probability is not an actual detection probability.

dirvine (changes requested):
Re-reviewed exact head `f64f66229c27b2365570b915692f6254b904d470` with six completed delegated review seats plus independent local verification.

**Verdict: changes requested.** The previous proof-soundness, selective-silence, protocol-routing, geometry and ADR findings are substantially addressed. Two resource-control blockers remain, plus the direct prune-policy regression test requested in the prior review.

## Blocking findings

### 1. High — the audit pre-decode ceiling is bypassable via the core protocol ID

`src/replication/mod.rs:1013-1017,3500-3537`; `src/replication/config.rs:384-409`

The 512 KiB guard is selected solely from `inbound_protocol`. An audit-family body sent on `REPLICATION_PROTOCOL_ID` skips that guard, is decoded under the 10 MiB core allowance, and is rejected by `body_matches_protocol` only after its variable-length collections have been allocated and decoded.

I reproduced this on the exact head with a valid `SubtreeSliceChallenge` carrying 200,000 openings:

```text
encoded bytes = 6,600,102
audit cap     =   524,288
core cap      = 10,485,760
```

So the new guard does not close the pre-admission allocation/decode path it claims to bound. Please use a protocol-independent safe pre-decode ceiling, a bounded wire collection/decoder, or a small envelope/discriminant that can be validated before decoding attacker-sized fields.

### 2. High — partial round-one read/hash work is not charged

`src/replication/storage_commitment_audit.rs:1141-1205`; `src/replication/mod.rs:3116-3182,3218-3224,3732-3758`

Round one reads and hashes leaves incrementally, but a later missing key, persistent read error, or blocking-task failure returns `Rejected` after discarding the already-built leaves. The global budget is charged only when the final response is `SubtreeAuditResponse::Proof`.

A retained commitment containing a failing key therefore permits repeated subtrees to consume substantial disk/hash work without touching the identity-independent budget; rotating identities still bypass the per-peer cooldown on this path. Please charge incrementally after successful leaf work, or return/guard consumed-work accounting so every exit is charged.

### 3. Medium — direct prune timeout-versus-malformed regression coverage is still missing

`src/replication/pruning.rs:1385-1418,1504-1545`

The implementation now correctly distinguishes `Unanswered` from undecodable `Malformed`, and only graces `Unanswered`. No direct test exercises those variants through the prune rollout decision. The generic `rollout_gate_graces_only_timeouts` test covers the responsible-audit predicate and existed while the prune-specific collapse was still present, so it cannot prevent this regression.

Please add direct assertions that, while the rollout gate is enabled, prune `Unanswered` is graced and prune undecodable/`Malformed` is penalised. The post-replication possession branch would benefit from the same production-boundary test.

## Non-blocking cleanup

Several security comments still say timeout is immediately penalised or keeps credit despite the active rollout gate and the new commitment-scoped round-two revocation. Please align at least `storage_commitment_audit.rs:253-257`, `possession.rs:67-69,98-100`, and the surrounding timeout-policy comments in `mod.rs` with the shipped behaviour.

## What is resolved

- Selective round-two silence now revokes the exact pinned commitment's holder credit without a peer-wide trust penalty.
- Bao length/EOF binding, keyed nonce derivation, canonical sibling depth and out-of-range rejection are present and tested.
- Subtree and possession audit families route symmetrically on dedicated IDs; core replication remains on v2.
- ADR-0009 and `REPLICATION_DESIGN.md` now cover all four lanes, rollout ownership and the objective grace-removal criterion.
- Round-two work is coalesced per key and heavy hashing runs through `spawn_blocking`.

## Verification

- `git diff --check origin/main...HEAD` — pass
- `cargo test -q --lib --no-fail-fast` — 816/816 pass
- `replication::slice` — 20/20 pass
- `storage_commitment_audit` — 22/22 pass
- `poc_commitment_audit_attacks` — 19/19 pass
- `poc_audit_handler_live` — 11/11 pass
- Current-head GitHub CI — all reported checks green across Linux/macOS/Windows, clippy, docs and security audit

Review-team result: the completed proof/crypto seat found the earlier construction blockers fixed; routing/docs seats agreed the family split and ADR work are sound; the runtime and adversarial seats independently converged on the two resource blockers above. No dissent on the reproduced protocol-ID bypass. The direct Z.AI GLM-5.2 credential was not available to the delegated runtime, so I am not claiming a distinct GLM model seat.

grumbach:
All three addressed at `9fcc404`. Thanks — the protocol-ID bypass was a real hole and the reproduction made it trivial to confirm.

### 1. Pre-decode ceiling bypass — fixed

You were right that the guard did not bound the path it claimed to. It selected the ceiling from `inbound_protocol`, so an audit body addressed to the core id skipped it entirely and decoded under the 10 MiB allowance before `body_matches_protocol` dropped it.

The ceiling now follows the **body**, not the address. A body's postcard discriminant sits behind a single varint (`request_id`), so it is readable before anything attacker-sized is touched — two bounded varint reads, no allocation. `family_of_variant` maps it to a family, and that one table now backs the pre-decode ceiling, the post-decode guard, the outbound response routing and both `is_*_audit` predicates. The pre- and post-decode views of "which family is this" can no longer disagree, which is what the bug was.

A prefix too malformed to classify takes the **strict** ceiling — failing to classify has to mean "do not decode generously", never "let it through".

Your reproduction is now a test verbatim: `oversized_audit_body_is_classified_as_audit_wherever_it_is_addressed` builds the 200,000-opening `SubtreeSliceChallenge`, asserts it exceeds `MAX_AUDIT_MESSAGE_SIZE`, and asserts it classifies as `SubtreeAudit` from the discriminant alone. Two more back it: `peeked_discriminant_matches_the_decoded_one_for_every_variant` round-trips all 17 variants through `encode` at request ids straddling postcard's varint width boundaries (so the peek is proven against the real encoding rather than my reading of the format), and `unclassifiable_prefix_yields_no_family` pins that a truncated or non-terminating prefix does not classify as `Core`.

### 2. Partial round-1 work — fixed

Also right, and the old comment argued the wrong thing: it reasoned the rejecting paths "turn on this node's own storage rather than on anything the caller sends", which misses that a commitment holding one unreadable key costs a full run of reads and keyed-BLAKE3 passes over every leaf before it, on every replay. With the per-peer cooldown escapable by rotating identity, the responder-wide budget was the only bound that applied to that, and it never saw the work.

`handle_subtree_challenge_measured` now returns the response together with the content bytes read, accrued **at the read** rather than at the end, and the caller charges unconditionally. The accumulator is threaded in rather than returned per-arm, so a future early return cannot forget to account for reads that already happened. A zero charge is a no-op, so the paths that touch nothing are unaffected. `handle_subtree_challenge` still exists as a thin wrapper, so the PoC call sites are unchanged.

Two production-boundary tests: `round1_proof_reports_the_content_it_read` fixes what the measurement means (it equals the sum over the proof's leaves), and `rejected_round1_still_reports_the_work_it_spent` deletes the bytes behind the **last** leaf of a real multi-leaf subtree, then asserts the response is the missing-bytes rejection *and* the charge is the baseline minus exactly that one leaf.

### 3. Prune timeout-vs-malformed coverage — added

Agreed on why the generic test could not have caught it: it lives on the responsible-audit predicate and passed while the prune collapse was still present.

Both lanes now expose the decision as a pure predicate — `prune_send_failure_is_penalised` and `probe_outcome_is_penalised` — used by the production paths and asserted directly. `prune_rollout_gate_graces_silence_but_never_a_bad_answer` asserts `Unanswered` is graced and `Malformed` is penalised; `possession_rollout_gate_graces_silence_but_never_a_bad_answer` does the same for the probe, and additionally that `Present`, `BootstrapClaim` and `Inconclusive` are not failures at all. Both tie the graced case to `GRACE_POSSESSION_AUDIT_TIMEOUTS` rather than to a literal, so removing the gate flips the assertions instead of leaving them quietly asserting the wrong thing. `probe_outcome_is_penalised` is exhaustive over `ProbeOutcome`, so a new outcome cannot silently inherit "not penalised".

The line the gate draws, stated once: it graces **silence**, never a wrong answer. A peer that replies at all is on this protocol, so a reply that does not decode is a protocol failure, not a version skew.

### Non-blocking comment drift — fixed

`storage_commitment_audit.rs` (the `TransientReject` doc contradicted the `Timeout` doc directly below it, which now revokes commitment-scoped credit), the `possession.rs` module header and `ProbeOutcome::Timeout`, and the two timeout-policy comments in `mod.rs`.

### Verification

`cfd` clean (clippy `-D warnings`, fmt, `cargo doc --deny=warnings`). 822 lib tests, `poc_commitment_audit_attacks` 19/19, `poc_audit_handler_live` 13/13, and the audit + prune e2e 10/10 including the real-QUIC honest-passes / deleter-caught / no-false-positives set.

dirvine (changes requested):
Re-reviewed current exact head `2021de9f3c1cf4d45c0ab6a59a6fa86baa0bb0ea` with six completed delegated seats plus independent code and runtime verification.

**Verdict: changes requested.** The three findings from the `f64f662` review are addressed: wrong-protocol audit requests now take the body-family ceiling before decode, partial successful round-one leaf work is charged on rejected outcomes, and direct prune/possession rollout tests now distinguish silence from malformed replies. One response-side resource-bound gap remains.

## Blocking finding

### Medium — solicited audit responses still decode under the generic 10 MiB ceiling

`src/replication/protocol.rs:65-84`; `src/replication/storage_commitment_audit.rs:211-232,326-353`; `src/replication/audit.rs:300-315`; `src/replication/possession.rs:325-347`; `src/replication/pruning.rs:1540-1565`

The new 512 KiB family-aware guard is only in `handle_replication_message`. Request-response replies are intercepted by `saorsa-core` and delivered directly to each `send_request` caller, where all five audit paths call the generic `ReplicationMessage::decode` before checking the expected response type or collection count.

I reproduced this on the reviewed tree with a valid oversized `AuditResponse::Digests`:

```text
encoded response = 6,400,007 bytes
audit cap       =   524,288 bytes
generic cap     = 10,485,760 bytes
ReplicationMessage::decode = accepted (200,000 digests allocated)
```

The responder must be the authenticated peer that was challenged, so this is less exposed than the earlier unsolicited/core-ID bypass. It is nevertheless material in this T3 PR because it contradicts the stated audit-family wire bound, and prune confirmation can have up to 32 solicited responses in flight. Apply `MAX_AUDIT_MESSAGE_SIZE` before every audit-response decode, preferably through one shared bounded audit decoder, and add a response-side regression using an oversized `AuditResponse` or `SubtreeSliceResponse`.

## Review-team dissent / non-blocking follow-up

A persistent LMDB error on the first selected round-one leaf performs three attempted reads and two 200 ms backoffs but reports `content_bytes == 0`, so the responder-wide byte budget is not debited. Two seats treated this as blocking; one treated it as bounded unhealthy-storage behaviour because no chunk bytes are returned or hashed, and I agree with the latter classification under a material bar. A fixed attempted-read charge plus a first-leaf-error regression would still make the documented resource guarantee more complete.

## Verification

- `git diff --check` — pass
- `cargo test -q --lib --no-fail-fast` — 822/822 pass on the reviewed resource-control tree
- `cargo test -q --features test-utils --test poc_audit_handler_live` — 13/13 pass
- focused protocol/prune/possession/storage-audit filters — pass
- current-head CI — completed jobs pass; macOS/Ubuntu/Windows tests and Windows build were still pending at final check

The cryptographic construction, selective-silence handling, protocol-family routing, ADR/design alignment, partial-success work accounting, and prune regression coverage now look sound. This remaining issue is the other half of the same pre-decode boundary: replies need the audit ceiling too.

dirvine (changes requested):
Re-reviewed exact head `56210cc707b75bce2311004c5042bcac26caacf5` with six completed delegated seats plus independent local verification.

**Verdict: changes requested.** The response-side ceiling blocker is fixed, the previous zero-charge first-leaf path is now non-zero, and the legacy challenge compatibility path is correctly scoped. One material sustained-work issue remains.

## Blocking finding

### High — the 4 KiB leaf floor does not make fixed per-leaf work hit the global budget

`src/replication/config.rs:162-246`; `src/replication/mod.rs:3159-3225,3269-3305,3773-3816`; `src/replication/storage_commitment_audit.rs:67-99,1220-1300`

The new charge is correctly placed before the read and totals `max(content_len, 4 KiB)` per attempted leaf. That fixes the literal zero counter. The calibration, however, leaves the identity-independent budget non-binding for the fixed work it now claims to cover: LMDB lookups/retries, permit occupancy, and one blocking-task dispatch/join per successful leaf.

Using the shipped constants:

```text
work refill                         64 MiB/s
attempt floor                        4 KiB
persistent first-leaf failure       3 reads + 2 × 200 ms backoff
2 heavy slots / 400 ms              ~5 requests/s
charge at that rate                 20 KiB/s
refill / charge                     3,276.8×
```

The two heavy slots can therefore remain occupied while the budget stays full. The healthy-storage variant has the same problem: a legal 1,024-leaf tiny-record proof is charged 4 MiB, so the bucket permits 16 such proofs/s — 16,384 LMDB lookups and blocking-task round trips/s — before it binds. In practice the two-slot pipeline saturates first, so rotating identities can continuously capacity-drop honest subtree audits without consuming the identity-independent allowance.

The new first-leaf test covers `Ok(None)` by deleting the record. It does not exercise persistent `Err` and its retries/backoffs, and the budget tests prove debt mechanics rather than that the chosen floor constrains this work.

Please keep the byte bucket for content read/hash, and add a separately calibrated responder-wide attempt/IOPS budget (charging each read attempt/retry and blocking dispatch), or otherwise justify and test a floor that becomes limiting before the heavy pool can be held continuously. Add regressions for sustained fresh-identity tiny-record work and persistent first-leaf `Err`, not only one `Ok(None)` attempt.

## Resolved

- All five solicited audit response paths now use the 512 KiB audit decoder before postcard allocation.
- The exact 200,000-digest / 6.4 MB reproducer is rejected by both the audit decoder and the generic family-aware decoder.
- First-leaf `Ok(None)`/`Err` and later failures can no longer report zero work.
- New nodes always ask on the dedicated possession ID; legacy core-ID challenges alone are accepted during the rollout gate and answered using the old digest on the old ID. Core-ID possession responses remain refused.
- ADR-0009 now states the compatibility weakening and proof delegability honestly.

## Review-team dissent

The completed panel was split on calibration: three focused/adversarial seats treated the underpriced operation work as blocking; one broad seat approved; the compatibility seat approved; the ceiling seat found only an unknown-discriminant hardening point. I independently classify the unknown-discriminant point as non-blocking because postcard rejects an unknown enum arm before decoding any trailing collection.

## Verification

- `git diff --check origin/main...HEAD` — pass
- `cargo test -q --lib --no-fail-fast` — 829/829 pass
- `cargo test -q --features test-utils --test poc_audit_handler_live` — 14/14 pass
- `cargo test -q --features test-utils --test poc_commitment_audit_attacks` — 19/19 pass
- exact response-ceiling, first-leaf and compatibility filters — pass
- current-head CI — all completed checks pass; macOS/Ubuntu/Windows test jobs were still pending at final check

mickvandijke (changes requested):
## Merged deep review — changes requested

Reviewed exact head `56210cc707b75bce2311004c5042bcac26caacf5` against base `c8f0ce15c8d146ccbedcbec251f0b101c43b225d`. This follow-up merges direct source analysis with independent Opus xhigh and GPT high reviews.

### Blocking

#### 1. Round-one's responder-wide budget does not meaningfully bound fixed per-leaf work

**Locations:** `src/replication/config.rs:199-246`, `src/replication/mod.rs:3249-3305`, `src/replication/storage_commitment_audit.rs:1220-1287`

Each attempted leaf is charged a 4 KiB minimum, but it may also cost an LMDB lookup, up to three read attempts/two 200 ms backoffs, a `spawn_blocking` dispatch/join, and keyed-BLAKE3 hashing. With two heavy slots, persistent first-leaf errors can occupy both slots at roughly five requests/s while charging only about 20 KiB/s; the bucket refills at 64 MiB/s, about 3,276.8× faster. A legal 1,024-leaf tiny-record proof costs only 4 MiB, so the nominal refill permits 16 such proofs/16,384 leaf operations per second—well above execution capacity.

A caller rotating peer identities can therefore keep both heavy slots occupied while the identity-independent budget remains full, causing honest audits to be capacity-dropped before the intended sustained-work control engages.

**Required fix:** retain byte accounting for content hashing, but separately meter fixed lookup/retry/dispatch work with a calibrated responder-wide attempt/IOPS budget (or use a demonstrably sufficient fixed-work unit). Add tests for sustained fresh-identity tiny-record traffic, persistent first-leaf errors including retries, and refusal before the heavy pool can remain continuously saturated.

#### 2. A round-two session authorizes arbitrary commitment keys, not only the subtree proven in round one

**Locations:** `src/replication/mod.rs:3153-3160`, `src/replication/mod.rs:3820-3829`, `src/replication/storage_commitment_audit.rs:1493-1501`

`SubtreeSession` retains the commitment hash and nonce but not the selected subtree range/leaf set. Round two accepts any key for which `built.proof_for(&key)` succeeds, i.e. any key anywhere in the pinned commitment.

An adversarial auditor can complete a cheap round one over tiny records, then request openings for five unrelated 4 MiB records elsewhere in the commitment, forcing up to 20 MiB of reads plus Bao/nonced-tree hashing. Round two has concurrency limits but no corresponding responder-wide sustained-work budget; identity rotation bypasses the per-peer cooldown. This lets cheap round-one work authorize materially more expensive unrelated round-two work.

**Required fix:** store the selected path/range or authorized leaf keys in `SubtreeSession`; reject every opening outside that exact round-one authorization before reading storage or constructing a proof. Add a regression showing an inside-subtree key is accepted, an existing outside-subtree key is rejected, and rejection performs no chunk read/proof construction.

### Non-blocking correctness and observability

#### 3. Explicit `Absent` is counted as `DigestMismatch`

**Locations:** `src/replication/storage_commitment_audit.rs:752-781`, `src/replication/storage_commitment_audit.rs:1048-1073`

`SubtreeSliceItem::Absent` becomes `Some(None)`, but the subsequent pattern maps it to `AuditFailureReason::DigestMismatch`. Thus `KeyAbsent` is not emitted, `absent_keys` remains zero, and admitted deletion is indistinguishable from a failed cryptographic proof in logs/counters. Penalty and credit effects are currently identical, so this is observability rather than proof soundness.

**Fix:** map `Some(None)` to `KeyAbsent`; reserve `DigestMismatch` for missing/wrong openings or failed proof chains. Add a summary-counter regression.

#### 4. The encoder does not enforce the audit-family ceiling

**Location:** `src/replication/protocol.rs:52` and the related `ReplicationMessage::encode` path

Audit decode paths enforce `MAX_AUDIT_MESSAGE_SIZE`, while general encoding checks only `MAX_REPLICATION_MESSAGE_SIZE`. Legitimate current responses fit, but future proof growth could let an honest responder encode a body every peer rejects as malformed, potentially penalising correct behaviour.

**Fix:** derive the family from the body variant during encoding and enforce the same ceiling as the decoder.

#### 5. Unknown discriminants default to the core family

**Location:** `src/replication/protocol.rs:332` and surrounding `family_of_variant` logic

The `_ => BodyFamily::Core` fallback lets a successfully parsed but unknown discriminant use the 10 MiB core pre-decode allowance rather than the strict 512 KiB audit allowance. Postcard rejects the unknown outer enum early, so no attacker-shaped allocation was demonstrated; the current impact is unnecessary transport/parse work and an invariant gap.

**Fix:** return `Option<BodyFamily>` or add `Unknown`, and apply the strict ceiling to unknown discriminants. Test known families, malformed/truncated prefixes, varint overflow, and validly encoded unknown discriminants.

### Verification

- Current head/base rechecked immediately before submission.
- Current-main merge tree was clean.
- `cargo fmt --all -- --check` passed.
- `replication::slice`: 20/20 passed.
- `storage_commitment_audit`: 22/22 passed.
- `round1_work_budget`: 2/2 passed.

The functional tests and testnet evidence are strong, but they do not exercise the adversarial tiny-record/retry-heavy and outside-subtree resource scenarios above. Findings 1 and 2 should be addressed before merge.

PR #182, opened 2026-07-23
fix(replication): stop the first-audit scheduler from targeting the local peer

grumbach:
## Problem

On every payment a node verifies, the payment's quote list includes the node's own quote, so the verifier emits a monetized-pin first-audit event targeting the local peer. The first-audit scheduler launched these audits without a self check, so the node tried to send an audit challenge to itself. The dial fails instantly (`Peer not found`, 0-1 ms, the local peer has no dialable address) and was counted as a subtree audit timeout.

Reported by @mickvandijke from two independent 400-node testnet runs: 396/396 and 382/423 `Peer not found` first-audit failures targeted the requester's own peer ID, about 8% of all first-monetized subtree audit starts, inflating the timeout metrics.

## Fix

Filter self-targets at the single ingress choke point (`queue_first_audit_event`) that both receive arms of the drainer feed through:

- A self-targeting event is never queued, never launched, and never marks its pin first-audited.
- A new `self_target_skipped` counter is reported in the periodic scheduler summary line, so a testnet run can confirm the filter is catching events (zero self-targeted outbound audit requests, non-zero skip counter) rather than the events silently not occurring.
- A `test-utils`-gated `first_audit_stats()` snapshot exposes the scheduler counters to e2e tests (they were previously created inside the drainer task and unobservable).

Audit coverage of the node's own pin is not reduced: a self-dial can never verify anything, while the payment's other payees each run the same cross-check on the same quote list and schedule their own first audit of this node (best-effort, as before).

## Evidence

Live-testnet e2e (multi-node in-process testnet, real QUIC wire, real LMDB, live drainer task):

- `first_audit_drainer_drops_self_targeting_monetized_pin`: a self event injected into the running drainer is ingested and dropped: `received: 1, self_target_skipped: 1, queued: 0, launched: 0`.
- `first_audit_drainer_launches_and_passes_remote_monetized_pin`: a remote peer's real commitment pin passes the filter and completes a passing live-wire subtree audit: `queued: 1, launched: 1, passed: 1`.
- Ablation (guard temporarily disabled, not committed): both test layers go red, and the e2e test reproduces the production bug exactly: `received: 1, queued: 1, launched: 1, timed_out: 1` — the node dialed itself and the instant `Peer not found` was recorded as an audit timeout.
- Unit tests cover the filter outcome directly (self dropped + pending untouched, remote unaffected, coalescing/eviction semantics unchanged).

An independent adversarial model review of the commit confirmed the filter is complete: the only production insertion into the pending queue is `queue_first_audit_event` (both receive arms), the cooldown re-put only recycles already-filtered events, and every other audit path (gossip-triggered, possession checks, prune confirmation) already self-excludes. Two low-severity notes from that review are addressed in this PR (live-drainer test coverage; comment wording on other-payee coverage being best-effort).

## Not included

Remote-peer `Peer not found` results are still classified as timeouts. Splitting those into a distinct unreachable outcome touches audit failure classification and is left for a follow-up.

Semver: patch, internal behavior only (one new field in an internal summary log line; the stats accessor is test-utils-gated).

### DIFF
--- a/src/replication/mod.rs
+++ b/src/replication/mod.rs
@@ -101,6 +101,30 @@ struct FirstAuditObservability {
     inflight: AtomicU64,
 }
 
+/// Test-only snapshot of the first-audit scheduler counters.
+///
+/// Lets e2e tests assert on the scheduler's decisions (e.g. that a
+/// self-targeting monetized pin was dropped and never launched) instead of
+/// scraping log lines.
+#[cfg(any(test, feature = "test-utils"))]
+#[derive(Debug, Clone, Copy)]
+pub struct FirstAuditStats {
+    /// Events ingested from the monetized-pin channel.
+    pub received: u64,
+    /// Events accepted into the pending first-audit queue.
+    pub queued: u64,
+    /// Events dropped because they targeted the local peer.
+    pub self_target_skipped: u64,
+    /// Audits launched.
+    pub launched: u64,
+    /// Launched audits that passed.
+    pub passed: u64,
+    /// Launched audits that timed out (non-response lane).
+    pub timed_out: u64,
+    /// Launched audits that ended in a confirmed failure.
+    pub failed: u64,
+}
+
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum FirstAuditTerminalOutcome {
     Passed,
@@ -162,8 +186,9 @@ enum FirstAuditQueueOutcome {
 /// is one of them, so each verified payment surfaces a self-targeting event.
 /// The node cannot audit itself over the network (there is no dialable address
 /// for the local peer, so the challenge fails instantly and is miscounted as a
-/// timeout), and the pin still receives its deterministic first audit from the
-/// payment's other payees. A self-target is therefore never queued — and hence
+/// timeout), while any other payee that verifies the same payment schedules its
+/// own deterministic first audit of this node's pin — a self-dial adds no
+/// coverage either way. A self-target is therefore never queued — and hence
 /// never launched nor marked first-audited.
 fn queue_first_audit_event(
     pending: &mut LruCache<PeerId, MonetizedPinEvent>,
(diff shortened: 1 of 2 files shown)


mickvandijke (approved):
## Merged review — PR #182

**Verdict: Approve — no blocking findings.**

### Confirmed
- **[Both]** The self-target guard is at the first-audit pending queue's single ingress helper, and both the primary `recv()` arm and batched `try_recv()` arm use it.
- **[Both]** The cooldown `pending.put()` only re-inserts entries that already passed the guard; no self-target can enter through that path.
- **[Both]** A skipped self-target is never queued, launched, or inserted into `first_audited`, and `self_target_skipped` is incremented on both receive arms.
- **[Both]** `FirstAuditStats` and its accessor remain gated behind `test`/`test-utils`; the production API and wire contracts are unchanged.
- **[GPT-5.6-sol]** Commitment pins are identity-bound (peer ID/public key are signed and included in the full commitment hash), so cross-peer pin aliasing cannot bypass the filter or its dedup behavior.

### Reconciled concerns
- Claude raised a possible shared-cooldown flake in the remote e2e test. I cleared it against the live path: DHT warmup completes before the commitment rebuild, the test-only rebuild does not gossip or trigger neighbor sync, and periodic neighbor sync is 10–20 minutes. The remote e2e passed four times during this review.
- Claude/Copilot noted that the assertions use lifetime counters. These engines are fresh per test, payment enforcement is off, warmup performs no payment verification, and `commit_and_seed` writes directly to storage, so the explicit injected event is the only counter source in the current harness. Delta assertions would be future-hardening, not a current correctness fix.
- The `Relaxed` snapshot is intentionally observational and can be transiently mixed; the polling tests wait for the relevant monotonic outcome. No production decision depends on the snapshot.

### Verification
- Exact head reviewed: `fb64e5b05224d7574dad4c28557058653863903d`
- Current base: `eb74b5640055986e41ea408053a72e6249cb0d82`
- `git diff --check`: clean
- `git merge-tree --write-tree origin/main HEAD`: clean
- Unit filter test: **1 passed**
- New live-drainer e2e tests: **2 passed**
- Remote live-drainer e2e repeated three additional times: **3/3 passed**
- GitHub CI: all checks green on macOS, Ubuntu, and Windows

_Reviewers: Claude Code `claude-opus-4-8` at xhigh + GPT-5.6-sol._

PR #183, opened 2026-07-23
fix(replication): cap first-audit pending admission at the launch budget

grumbach:
## Summary

Follow-up to #177 (stacked on its head; see merge-order note below). Sizes first-audit pending admission to what the launch budget can actually audit.

#177 bounded launches (token bucket) and expired dead entries eagerly, but admission was still unbounded relative to the budget: `pending` shared the 4096-entry commitment-cache cap while the token bucket can launch at most ~31 audits inside one effective answerability window. On the 2026-07-17 DEV-02 staging run this meant ~113 admitted nominations per node-hour against ~12 launches: over 99% of admitted work could never launch, pending sat at ~331-431k fleet-wide, and the oldest-pending age tracked the age of the deployment itself — an aging backlog whose steady state is mass expiry.

## Changes

- `pending` is capped at `FIRST_AUDIT_PENDING_CAP`, derived from the strict reserve-time launch horizon (burst at age zero plus every refill that stays answerable through max jitter + send slack; currently 31). The unit test simulates the shipped predicate instant-by-instant and must agree with the derivation. The dedup set keeps the 4096 bound.
- Overflow follows two rules, in order (review feedback):
  1. **Reservation before destructive overflow** — an arrival that would displace a different peer first gives the queue a launch opportunity; with a token free, one entry moves into the reservation and admission proceeds with no eviction, so an ingress burst cannot flush eligible work past a ready token.
  2. **Random-victim displacement** — an otherwise-admissible arrival that still finds the queue full after that one opportunity (a single jitter reservation is the guard slot; a second burst token can remain while it is occupied) displaces a uniformly random incumbent, counted as `capacity_evicted`. Per-nomination eviction probability of a specific target is 1/cap regardless of arrival order: suppressing it with 95% confidence costs ~95 distinct-peer paid nominations and is never certain (deterministic keep-newest LRU, flagged in review, allowed a certain flush with 32).
- The summary line gains `pending_cap` and `reserved` (the single reservation held outside the queue), so schedulable occupancy `pending + reserved` is directly computable in Elasticsearch.
- ADR-0004 Amendment 4 documents the admission sizing, the two overflow rules, and the eviction-cost math.

## Evidence

- 745 lib tests pass. Regressions include: ordered-flood non-determinism at the drain/reserve boundary (`pending_cap < 60 <= FIRST_AUDIT_DRAIN_BATCH`, 100-trial statistical bounds with miss odds < 1e-20), reserve-before-eviction at capacity, displaced-peer re-nomination, cap-vs-predicate simulation.
- Expected staging signals: `pending` ≤ 31 with `oldest_pending_quote_age_ms` far below the answerability window (queue drains instead of aging), overflow visible as `capacity_evicted` growth at admission time instead of `outside_answerability_window` mass expiry hours later.

## Semver

Patch: no protocol or config-surface change; two additive fields in the summary log line.

Tracking: [V2-738](https://linear.app/autonominetwork/issue/V2-738/first-audit-scheduler-size-pending-admission-to-the-launch)

### DIFF
--- a/docs/adr/ADR-0004-commitment-bound-quote-pricing.md
+++ b/docs/adr/ADR-0004-commitment-bound-quote-pricing.md
@@ -507,7 +507,7 @@ at least 99% detection within 24 hours for at least 1% missing bytes.
 Amendment 3 stopped dead entries from squatting the queue, but admission
 itself was still unbounded relative to the budget: `pending` shared the
 4096-entry commitment-cache cap while the token bucket can launch at most
-~32 audits inside one effective answerability window (150 min of
+~31 audits inside one effective answerability window (150 min of
 eligibility at one token per 5 minutes, plus a burst of 2). At the
 2026-07-17 staging arrival rate (~113 admitted nominations per node-hour
 against ~12 launches) over 99% of admitted work could never launch: the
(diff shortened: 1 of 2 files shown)


grumbach:
Update: #177 has merged, so the stack is resolved — this PR is now a normal 3-commit PR against main (admission cap, review fixes, policy-neutral test fix), rebased on top of the #177 and #182 merges.

dirvine (changes requested):
Reviewed exact head `d1fc1ddf0a4f710e16debd471cd3ec4c3916e92f` with a six-seat panel (four substantive completions; two tool-failed), then independently checked the scheduler ordering and focused tests.

**Verdict: changes requested.** The queue-sizing diagnosis is sound, but the current keep-newest cap introduces a deterministic suppression route before the scheduler can reserve work.

## Blocking — drain-before-reserve can evict a target deterministically

`FIRST_AUDIT_PENDING_CAP` is 32 (`src/replication/mod.rs:956-984`), while `FIRST_AUDIT_DRAIN_BATCH` is 64 (`src/replication/config.rs:399-405`). On an ingress wake the event loop enqueues/drains up to 64 nominations first (`src/replication/mod.rs:1621-1640`) and only then attempts reservation (`:1736-1758`). At capacity, every new distinct peer deterministically evicts the LRU peer (`:199-227`, `:619-632`).

Therefore an ordered batch of target + 32 later distinct-peer nominations removes the target before either newest/oldest lane can inspect it, even when a token and in-flight slot are available. Lane alternation cannot help because the target is already gone.

This cuts the stacked ADR's documented capacity-eviction threshold from 4,096 nominations (`docs/adr/ADR-0004-commitment-bound-quote-pricing.md:462-470`) to 32 — a 128x reduction — while Amendment 4 says every fresh pin retains an “unpredictable chance” of prompt audit (`:527-533`). The current replacement is deterministic under controlled arrival order. That matters because the ADR already assumes recyclable payment principal and sidecar-only pins without a lottery backstop (`:443-453`).

Please either give eligible work a reservation opportunity before destructive overflow, or use genuine adversary-resistant/random/weighted retention rather than deterministic keep-newest LRU. Add a production-order regression with `pending_cap < ordered_batch <= drain_batch`; the new enqueue-only test currently proves that the oldest target is evicted but does not exercise the drain/reserve boundary.

## Also fix in this head

- **Cap maths/wording:** the strict predicate rejects `age >= 150m` (`src/replication/mod.rs:920-928`). For a fresh quote, burst 2 plus refills at 5…145 minutes gives 31 usable launches, not 32; the max-jitter/slack prefilter is stricter again. The test at `:6590-6604` repeats the formula rather than simulating the predicate. Derive the cap from the same strict launch predicate/horizon.
- **Occupancy telemetry:** a reservation is held outside `pending` (`:477-487`, `:721-728`), so the scheduler may hold one reserved + 32 pending while reporting `pending=32, pending_cap=32`. Report reserved/total schedulable occupancy or narrow the claim.
- **Stack dependency:** #183 must remain after #177. The merge-order comment names stale cap SHA `cbc2f5e`; the current cap commit is this head, `d1fc1dd`.

## Verification

- `cargo test first_audit_ --lib --no-fail-fast` — 28/28 pass
- GitHub CI — all current checks green across Linux/macOS/Windows, clippy, format, docs, no-logging, audit and ADR validation
- Head revalidated immediately before review

The causal note on DEV-02/DEV-03 is fair: that comparison cannot attribute the upload p90 change. I also agree another attribution run is not necessary for this queue finding; the scheduler counters already show the admission/budget mismatch. This review is about the proposed retention policy, not the diagnosis.

dirvine (changes requested):
Re-reviewed exact head `3462edf3c4f183f47c7dc83f68a1fb46fb3f2c3a` with the six-seat follow-up panel and local focused checks.

The prior production blockers are addressed:
- overflow uses a random incumbent rather than deterministic LRU eviction;
- pending work gets a reservation opportunity before destructive overflow;
- strict launch-horizon maths now yields 31 and is checked by predicate simulation;
- telemetry exposes `pending`, `pending_cap`, and the separate reservation.

One merge blocker remains: `first_audit_suppressed_lower_flood_leaves_recency_and_counts` still asserts deterministic LRU eviction after the production policy changed to random victim selection (`src/replication/mod.rs:7284-7304`). Windows CI failed on that exact stale assertion (740 passed, 1 failed). I reproduced it locally on run 5/100. Please make the test verify retained value/count and unchanged recency directly, without requiring a particular random victim, then rerun CI.

Panel dissent, non-blocking: the ADR says random displacement occurs only “when no launch is possible”, but the scheduler permits only one outstanding jitter reservation even though burst/in-flight capacity is two. After the first reservation, a second token/slot may remain while later arrivals use random eviction. The single-reservation design is deliberate and the eviction is now adversary-resistant, so I do not classify this as a production blocker; please narrow the wording to “when no reservation slot is available” (or model multiple reservations if the stronger statement is intended).

Verified locally:
- `cargo test first_audit_ --lib --no-fail-fast` — 30/30 pass in one run
- repeated stale test — reproduced failure

Hold only for the stale test and fresh green CI. No merge performed.

dirvine (approved):
Final pass on exact head `2d1527f8d9012ee0cbfea5990fd9e7f531340d27`, using a six-reviewer panel plus independent local verification.

**Approved.** The previous blockers are resolved:
- the stale suppressed-lower test now checks MRU→LRU recency directly, without asserting a random victim;
- eviction sequences are reproducible under seeded test RNGs while production uses entropy-seeded `StdRng`;
- random-victim displacement removes the deterministic ordered-flush route;
- both ingress paths offer pending work a reservation before destructive overflow;
- the strict launch horizon derives and tests a cap of 31;
- `pending_cap` and the separate one-entry reservation are visible, and the ADR wording now describes that guard-slot limit accurately.

Verification:
- GitHub: 12/12 checks green, including Linux/macOS/Windows tests and builds, clippy, format, docs, no-logging, security audit and ADR validation.
- `cargo fmt --all -- --check` — pass.
- `cargo clippy --all-targets --all-features -- -D warnings` — pass.
- `cargo test first_audit_ --lib --no-fail-fast` — 31/31 pass.
- Formerly stale recency test — 200/200 repeated runs pass.
- `git diff --check origin/main...HEAD` — pass.

Panel consensus: five reviewers found no merge blocker. One operations reviewer dissented on observability wording: `capacity_evicted` is cumulative and can also include cooldown-race requeue displacement, not only direct ingress overflow. I agree that is worth tightening, along with a future drainer-level integration regression and softer “schedulable occupancy” wording, but the counter remains technically accurate, the behaviour is bounded/accounted, and none is a production-correctness or security blocker for this PR.

No merge performed.

PR #185, opened 2026-07-24
chore: add PR template and required CI checks (V2-719, V2-720)

jacderida:
## Linear issue
- https://linear.app/autonominetwork/issue/V2-719 (add the standard PR template)
- https://linear.app/autonominetwork/issue/V2-720 (require a linked Linear issue on every PR)

## Risk tier

## Compatibility
- Wire: none
- Storage: none
- API: none

## Semver impact
- [ ] breaking
- [ ] feature
- [x] fix

## Test evidence
`check_pr.py` was exercised locally across its cases: missing Linear reference
fails `linear`; a key in the title/body/branch passes; a body without the
template fails `template`; an rc-* base is a no-op pass; a fully-filled template
passes; and the failure modes (no/duplicate tier or semver box, empty section,
Tier 2/3 without an ADR link) each fail with a specific message. This PR's own
description is validated by the new checks.

## New dependency
none

## ADR
n/a

## Mitigation / rollback
Revert this PR; the template and checks are additive and carry no runtime effect.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,40 @@
+## Linear issue
+<!-- REQUIRED. Link the issue: an issue key like V2-123, or a linear.app URL.
+     CI blocks PRs with no linked Linear issue. -->
+
+## Risk tier
+<!-- Check exactly one. Boundary question: does this change node behavior, the wire
+     protocol, stored-data format, payments/economics, or the upgrade mechanism?
+     If yes -> T2/T3, and it rides the weekly release train.
+     If no  -> T0/T1, and it may ship via the client track.
+     Propose the tier; a human confirms it at review. -->
+- [ ] T0 — docs / tooling / CI / pure UX-output. Repo CI only.
+- [ ] T1 — client-only, no network-facing behavior change. CI + prod compat smoke.
+- [ ] T2 — node/client logic with behavioral surface, no protocol/format/economics change. Dev testnet + ADR.
+- [ ] T3 — protocol / storage format / payments / routing. T2 evidence + adversarial testing.
+
+## Compatibility
+<!-- State the impact on each axis, or "none". -->
+- Wire:
+- Storage:
+- API:
+
+## Semver impact
+<!-- Check exactly one. This is where the crate's bump level is decided, while the
+     context is fresh; the train-manifest skill maps it to a per-crate version bump. -->
+- [ ] breaking
+- [ ] feature
+- [ ] fix
+
+## Test evidence
+<!-- Per the tier: what was run, at what scale, and the result. Attach or link artifacts. -->
+
+## New dependency
+<!-- Any new external dependency? Write "none", or list them. New deps need explicit
+     human acknowledgement in review. -->
+
+## ADR
+<!-- REQUIRED for Tier 2/3 — link the ADR. Write "n/a" for Tier 0/1. -->
+
+## Mitigation / rollback
+<!-- One line: how we back this out or limit the blast radius if it misbehaves. -->
--- a/.github/scripts/check_pr.py
+++ b/.github/scripts/check_pr.py
@@ -31,7 +31,12 @@ LINEAR_TEAM_PREFIXES = ("V2", "AUTO", "REL", "INFRA", "QA")
 LINEAR_KEY = re.compile(
     r"\b(?:" + "|".join(LINEAR_TEAM_PREFIXES) + r")-[0-9]+\b", re.IGNORECASE
 )
-LINEAR_URL = re.compile(r"linear\.app/", re.IGNORECASE)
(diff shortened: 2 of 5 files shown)


dirvine (changes requested):
I found two gaps in the enforcement, reproduced against the checker on this PR head:

1. **The Linear gate does not reliably require a Linear issue.** `LINEAR_KEY` accepts any `<letters>-<digits>` token anywhere in the title, body, or branch. For example, a PR containing only `UTF-8` as its supposed issue reference passes both `linear` and `template`. Common technical strings such as `SHA-256`, `HTTP-2`, or `RFC-123` can therefore satisfy the required check accidentally. Please require a real `linear.app/.../issue/...` URL, or constrain keys to the organisation's known Linear team prefixes.

2. **“Fill every field” is not enforced.** A Compatibility section with only `Wire: none` passes while `Storage:` and `API:` are blank, and a T0/T1 PR also passes with an empty ADR section. I reproduced that combination locally and received `PR template complete`. Please require all three compatibility axes and require an explicit ADR value (`n/a` for T0/T1; a link for T2/T3).

The six PRs currently carry the same checker (SHA-256 `89aa9a45c0444cb40726a3e5d3b8cc5e402d75ed4e50341fb2fff84053098649`), so the same fix should be applied consistently across them. The template/workflow shape otherwise looks sound, and the new checks themselves are running successfully.

jacderida:
Thanks — both gaps are fixed, and the same checker (identical across all six PRs) is updated consistently.

**1. Linear gate no longer accepts arbitrary tokens.** A reference now counts only if it is a `linear.app` issue URL, or an issue key with a known Linear team prefix (`V2` / `AUTO` / `REL` / `INFRA` / `QA`). `UTF-8`, `SHA-256`, `HTTP-2`, `RFC-123` all fail now (verified). The prefix list is a documented constant to update when a team is added. Relatedly, HTML comments are now stripped before matching, so the template's own `V2-123` example in the comment no longer satisfies the check.

**2. "Fill every field" is enforced.** All three Compatibility axes (Wire/Storage/API) must each carry a value (`none` is fine), and the ADR field must be filled explicitly — `n/a` for Tier 0/1, a real link for Tier 2/3. A T0 PR with a blank ADR, or a Compatibility block with only `Wire:` filled, now fails. While fixing this I also caught a regex bug where an empty axis could "borrow" the next line and pass.

Also addressed from the Copilot review: exact template headings in the "missing section" messages (`## ADR`, `## Mitigation / rollback` rather than `Adr` / `Mitigation / Rollback`), and the workflow now also triggers on `ready_for_review`.

One point I did **not** change yet: Copilot's tamper-resistance note (switch `pull_request` → `pull_request_target` so a PR can't edit the checker to force a pass). It's valid, but `pull_request_target` runs the workflow from the *base* branch — which won't have this workflow until these PRs merge — so applying it now would stop the checks running on these very PRs. Suggest converting it in a fast-follow once this is on `main`.

dirvine (changes requested):
The compatibility and ADR enforcement is fixed, and the constrained issue-key prefixes correctly reject `UTF-8`/`SHA-256`-style tokens. One residual gap remains in the Linear URL path:

`LINEAR_URL = re.compile(r"linear\.app/", re.IGNORECASE)` accepts any page on the domain. I reproduced both `https://linear.app/changelog` and `https://linear.app/not-an-issue` passing the `linear` and `template` checks as a linked issue.

Please constrain the URL pattern to an actual issue path, for example `linear.app/<workspace>/issue/<key>` (allowing the optional issue slug if needed), and include this negative case in the test matrix. Everything else from my previous review is addressed.

jacderida:
Fixed — and thanks for catching the loose URL path.

**Linear URL is now constrained to a real issue path.** `LINEAR_URL` matches `linear.app/<workspace>/issue/<key>` (with an optional trailing slug), so `https://linear.app/changelog` and `https://linear.app/not-an-issue` are both rejected now. I also fixed the message inconsistency Copilot flagged: a reference is accepted in the title, body, or branch, and the message now says exactly that.

**Both negatives are in a committed test matrix.** Added `.github/scripts/test_check_pr.py` — a 17-case matrix that runs the real checker as a subprocess and asserts the outcome, including `linear.app/changelog`, `linear.app/not-an-issue`, `UTF-8`/`SHA-256`, the unfilled-template-comment case, and the completeness cases (compat axes, ADR by tier, tier/semver box counts). It runs in CI as a new `self-test` job, so the matrix is enforced on every PR, not just something I ran locally.

The same checker + tests are identical across all six PRs (`check_pr.py` sha256 `3f88cf70…`).

dirvine (approved):
Follow-up verified. The Linear URL matcher now requires an actual `<workspace>/issue/<key>` path, the two reported false-positive URLs are covered by regression tests, and the shared test matrix passes 17/17 locally. The required `linear-link` and `pr-template` checks are green. Approved.

PR #186, opened 2026-07-27
Reduce storage-commitment audit egress by verifying co-held leaves locally

grumbach:
## Summary

The storage-commitment audit (ADR-0002) round 2 has the responder serve the
sampled chunks' bytes back to the auditor, which is a large share of
steady-state replication traffic. When the auditor already holds a sampled
leaf, it can verify that leaf against its own copy instead of receiving the
bytes over the wire.

This change:

- Verifies a sampled leaf the auditor already holds against the auditor's own
  copy, with no chunk bytes on the wire. A sampled leaf the auditor does not
  hold is wire-challenged for the original bytes exactly as before. The verdict
  is the unchanged pass/fail with whole-slice credit.
- Makes the possession digest a domain-separated keyed BLAKE3 hash so the fresh
  per-audit nonce is bound into the digest over the whole content. The
  replication protocol id advances to v3 for a clean cutover.
- Returns no verdict (Idle) from the responsible-chunk audit when a challenged
  key's own local copy is unreadable, instead of a pass that reflects only the
  readable keys.

## Expected impact

Expected egress for an audit is `(1 - q)` of round 2's chunk traffic, where `q`
is the fraction of the sampled leaves the auditor holds. The first-audit lane,
where the auditor holds nothing, is unchanged.

## Testing

- Adds test-only observability (compiled under `test`/`test-utils` only) so an
  e2e test can confirm co-held leaves are verified locally with no byte
  challenge.
- Full lib suite, the commitment-audit PoC suites, and the real-QUIC
  subtree-audit e2e pass; `cargo fmt` and `clippy` are clean.

### DIFF
--- a/src/replication/audit.rs
+++ b/src/replication/audit.rs
@@ -48,7 +48,11 @@ pub enum AuditTickResult {
         /// The peer claiming bootstrap status.
         peer: PeerId,
     },
-    /// No eligible peers for audit this tick.
+    /// No eligible peers for audit this tick, OR the auditor could not run a
+    /// meaningful check (e.g. the responsible-chunk audit found every sampled
+    /// key's own local copy unreadable). Nothing was proven either way: no
+    /// holder credit, no trust event, no penalty. The audit simply did not
+    /// happen this tick, exactly like having no peer to audit.
     Idle,
     /// Audit skipped (not enough local keys).
     InsufficientKeys,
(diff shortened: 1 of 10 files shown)


grumbach:
Closing in favour of #181, which reaches the same goal — cutting audit egress without weakening the possession proof — by a much larger margin, and which has already absorbed the parts of this PR that stand on their own.

## What #181 took from here

`9c5f83f` on #181 carries two of this PR's three changes:

- **The domain-separated keyed BLAKE3 possession digest.** `compute_audit_digest` is byte-identical to the version here, as is the `possession.rs` doc hunk. #181 extends the same construction to every digest-based lane (responsible-chunk, post-replication possession probe, prune confirmation) rather than the possession digest alone.
- **Idle-on-unreadable.** The responsible-chunk audit yields no verdict instead of a pass when a challenged key's own local copy cannot be read.

So nothing in those two areas is lost by closing this.

## Why the local-verify optimisation does not survive #181

The remaining change here — verifying a co-held sampled leaf against the auditor's own bytes — was a clear win against the current 6 MB round-2 response: expected egress `(1 - q)` of round-2 chunk traffic, roughly 5× at a co-hold fraction of `q = 0.8`.

But #181 attacks the same bytes multiplicatively rather than proportionally. jacderida's 990-node DEV run measured round-2 at **14.49 KB vs 6.19 MB** for a v2 control cohort in the same network under the same load — a **417× reduction**, which puts total round-2 audit egress at **1.04 GB out of 3,644 GB** of replication egress, or **0.03%**.

Layering this PR on top of that saves at most `q × 1.04 GB`. Even at `q = 1.0` the ceiling is 0.03% of replication egress.

The cost side then inverts. `local_leaf_matches` has the *auditor* read its own full chunk (up to 4 MiB) out of LMDB and hash it twice — `blake3::hash(bytes)` plus `compute_audit_digest(..bytes)` — for each co-held sampled leaf, 3–5 leaves per audit. Against a 6 MB response that buys ~4 MiB of WAN egress for ~4 MiB of local disk read, and disk is far cheaper than WAN. Against #181's ~3 KB per leaf it spends ~4 MiB of disk read and ~8 MiB of hashing to save ~3 KB of wire. That is roughly a thousand to one against, on the auditor side, for no security gain: #181's round-1 verifier already enforces `bytes_hash == key` by content addressing, so the extra check local-verify performs is redundant there.

One caveat worth recording: `q` was never measured, here or in #181. It does not change the conclusion, since the saving is bounded by that 0.03% for any `q`.

## The cutover also differs, and #181's is safer

This PR advances `REPLICATION_PROTOCOL_ID` v2→v3, which partitions *all* replication during the upgrade window — fresh replication, neighbour sync, fetch, repair, commitment fetch. #181 keeps core replication on v2 and moves only the audit pair onto `SUBTREE_AUDIT_PROTOCOL_ID` / `POSSESSION_AUDIT_PROTOCOL_ID`, so a cross-version audit lands in the graced-timeout lane instead of the confirmed-failure lane and nothing else pauses.

The testnet run confirms that matters in practice: all 7 v2 hosts kept growing `paid_list.mdb` as well as `chunks.mdb`, i.e. the v2 cohort accepted paid stores from v3 peers throughout, with zero transfer failures across 736/736 and 135/135 mixed-version transfers.

## Note on CI

This PR is green and #181 is currently red, which reads backwards. #181's failures are three e2e tests still sending on the old protocol id after the audit moved to its own, plus the new PR-template check from #185 — both mechanical, neither a production defect. Being fixed there.

PR #188, opened 2026-07-28
Record the storage economics and payment protocol, and fix merkle/single-node payment divergence

grumbach:
ADR for payments as they are now + fix identified merkle and single payment divergence.

**This PR enforces the fix by default.** There is no shadow mode and no flag to
flip later. It must ship on the same train as the paired ant-client change that
pays the multiplier: **WithAutonomi/ant-client#161**. Either order is safe in the
meantime — 3x clears the old 1x requirement, and this side keys its requirement
off each receipt's own timestamp — but neither is complete without the other.

Tripling the common large-upload path is a pricing decision, so this still needs
the economic owner's sign-off before the train ships.

No ADR stated how storage is priced and paid for end to end. ADR-0004 capped
what a node may charge, ADR-0006 floored what it will accept, ADR-0005 decided
who may be paid — but the model those three amend was never written down.
ADR-0008 records it: the unit of sale, the price curve, the quoting round, both
payment shapes, what every storer re-verifies, and which of the surrounding
gates are enforced versus still observe-only.

Writing it down surfaced a defect. A chunk stored through a **merkle batch**
earned its node **one third** of what the identical chunk earned through a
single-node upload, for the same storage and the same replication. The
single-node path pays the median-priced issuer `3 x price`; the merkle path
copied the bare quoted price into the on-chain payable `CandidateNode.amount`,
so the contract's `median16(amount) x 2^depth` settled `1 x median` per leaf.
Nothing detected it: the storer recomputes its expectation from the same
un-multiplied signed prices, and `verify_prices` (which would compare on-chain
amounts against signed prices) is called by neither repo.

This side now **requires** the parity amount. The one thing enforcement must not
do is refuse money that was already spent correctly: a merkle receipt stays
spendable for a week, so the requirement keys off the receipt's own timestamp
against a compiled-in boundary rather than switching on the instant a node
upgrades. Receipts bought under the old rule keep it until they expire; the
branch honouring them becomes unreachable one week past the boundary, because
anything older is expired anyway. That is also what stops a client backdating
its way out of the increase. Full matrix under Compatibility.

It still logs `ant_node::payment::merkle_parity` per merkle store admission, but
that line no longer gates a decision — it answers how much traffic is still
arriving on pre-boundary receipts, i.e. when the 1x branch is dead code.

## Linear issue
- https://linear.app/autonominetwork/issue/V2-785/merkle-batch-payments-settle-a-chunk-at-a-third-of-the-single-node

## Risk tier

T3 without qualification: this changes what a storer accepts as payment, and
raises the effective price of the common large-upload path by 3x.

## Compatibility
- Wire: none. No message, proof or quote shape changes; the multiplier lands on the value of an existing on-chain field.
- Storage: none.
- API: additive — one new public const, `MERKLE_PARITY_ENFORCED_FROM_UNIX`.

**Admission behaviour does change**, so spelling the cutover out rather than
leaving it implied:

| Receipt stamped | Settled | Result |
|---|---|---|
| before the boundary | 1x | admitted (the rule it was bought under) |
| before the boundary | 3x | admitted |
| at/after the boundary | 3x | admitted |
| at/after the boundary | 1x | **refused**, naming the required multiplier |

- **Existing one-week receipts.** A merkle receipt stays spendable for a week.
  Requiring 3x from the moment nodes upgrade would refuse uploads already paid
  for, correctly, under the previous rule, with no refund path. Keying off the
  receipt's own timestamp means those keep working until they expire.
- **The window closes itself.** Receipts older than a week are expired anyway,
  so one week past the boundary every valid receipt is post-boundary and the 1x
  branch is unreachable. No cleanup task, and no way to backdate into the old
  price — a stamp old enough to qualify is old enough to be expired.
- **Rollout order is free.** 3x clears the 1x floor too, so upgraded client +
  old node works, and upgraded node + in-flight legacy receipt works. Neither
  half has to land first, though both should ride the same train.
- **Not protected:** a client still on old code after the boundary pays 1x on a
  fresh receipt and is refused. Unavoidable if 3x is required at all; the
  boundary sits a train later to give clients time, and the rejection is loud.

## Semver impact
- [ ] breaking
- [x] feature
- [ ] fix

## Test evidence
Unit tests, `cargo test --lib payment::` — 179 pass; full `cargo test --lib` —
757 pass. Eleven new tests in total; six cover the
per-node expectation: it reproduces the contract formula at 1x; it scales
within the division remainder across a matrix of medians and depths, and equals
the exact floor of the multiplied total; median 901 at depth 7 pins the
multiply-before-divide order, where scaling a 1x result would be one wei low; a
merkle leaf settles for the single-node 3x median at every depth 1..=8; depth 0
is zero; an empty pool is rejected.

Two of those came out of review. The contract's `total / depth` discards a
remainder when `depth` does not divide `median x 2^depth`, so the original
scaling test encoded `floor(3x/d) == 3*floor(x/d)` and only passed because its
inputs divided evenly.

`cargo clippy --all-targets --all-features -- -D warnings` clean;
`cargo fmt` clean; `RUSTDOCFLAGS=--deny=warnings cargo doc` clean.

### Enforcement change (head `10ecd58`)
- `MERKLE_PAYMENT_MULTIPLIER_ENFORCED` (shadow bool) replaced by
  `MERKLE_PARITY_ENFORCED_FROM_UNIX`, a receipt-timestamp boundary. Parity is
  required, not measured.
- Five new tests cover the cutover: 1x refused after the boundary (naming the
  multiplier), 1x honoured before it, 3x accepted on both sides, the switch is
  exactly at the boundary instant, and no unexpired receipt can still claim 1x
  one lifetime past it.
- Merkle tests pin the boundary instead of reading the production constant.
  Every receipt is stamped within a week of now, so tests left on the real
  boundary would sit in the legacy window today and silently switch rules on
  2026-08-04, changing what their assertions meant. Two were paying 1x and would
  have begun failing for the wrong reason.

### Addressed from earlier review (head `4bbf868`)
- Parity telemetry moved after the full paid-node validation loop and gated on
  `is_store_admission()`, so rejected proofs and paid-list replays cannot enter
  the rollout signal.
- Telemetry keyed by pool hash instead of chunk address: one settlement covers
  up to 256 chunks, so per-chunk lines were 256 duplicate samples of one
  economic event and let a big batch outvote a small one.
- ADR separates the shipped baseline from the decision, and corrects several
  claims: the witnessed quorum is not a fixed 5-of-7 and degrades to as low as
  1 with missing responder views; bundles carry 1..=7 quotes; non-paid quotes
  are not re-verified, so a bad signature on one is accepted; the on-curve check
  is off; the unmerged ADRs are labelled pending. It also states what parity
  does and does not equalise — aggregate client spend per padded leaf, against
  two different medians, not what any node earns, since merkle pays `depth`
  batch candidates rather than each chunk's storer — and lists the alternatives,
  including the shadow-first rollout that this head replaces.

Left deliberately unaddressed, per the PR author: the economic sign-off for
upward 3x parity.

**Gap, stated plainly:** no dev-testnet run and no adversarial pass yet, which
T3 asks for, and no live-chain test of the cutover. The unit tests cover the
matrix, but what a testnet run should still prove is a real batch upload from an
upgraded client being admitted at 3x, and an old client's 1x upload on a fresh
receipt being refused. Worth doing before this rides the train.

## New dependency
none

## ADR
- https://github.com/grumbach/ant-node/blob/payment-economics-adr/docs/adr/ADR-0008-storage-economics-and-payment-protocol.md (added by this PR)

Numbered 0008, not 0005: `ADR-0005-earned-reward-eligibility` exists on the
`adr-0005-earned-reward-eligibility` branch and is referenced by name in
shipped code (`ADR5_ENFORCE`, the audit-report type), and ADR-0007 is on the
verified-slice-audit branch. 0008 avoids colliding with either.

## Mitigation / rollback
Two levers, no data migration either way. Move
`MERKLE_PARITY_ENFORCED_FROM_UNIX` forward to postpone the requirement (also
what to do if the train slips past the current date — it must never sit earlier
than the release carrying it), or revert the commit to drop the requirement
entirely. Clients paying 3x stay admitted under both, since 3x clears the 1x
floor.

### DIFF
--- a/docs/adr/ADR-0008-storage-economics-and-payment-protocol.md
+++ b/docs/adr/ADR-0008-storage-economics-and-payment-protocol.md
@@ -1,0 +1,293 @@
+# ADR-0008: Storage economics and the quoting + payment protocol
+
+- **Status:** Proposed
+- **Date:** 2026-07-28
+- **Decision owners:** Anselme (@grumbach)
+- **Reviewers:** <pending>
+- **Supersedes:** none
+- **Superseded by:** none
+- **Related:** ADR-0002 (storage audit), ADR-0004 (commitment-bound quote
+  pricing), ADR-0006 (receiver-side revenue floor), ADR-0007 (audit proof
+  shape), ant-client ADR-0003 (earned-reward eligibility gate)
+
+## Context
+
+Every ADR so far amends one corner of the economics — ADR-0004 puts a ceiling
+on what a node may charge, ADR-0006 puts a floor under what it will accept,
+ADR-0005/ant-client ADR-0003 decide who may be paid at all — but the model
+those amendments modify was never written down. Anyone reading the ADR set
+today can see the patches and not the machine.
+
+This ADR is **descriptive**: it records the pricing and payment design as
+built, so that later ADRs can amend a stated baseline instead of an implied
+one. Where the code is behind a rollout flag, the ADR says so.
+
+Writing it down immediately surfaced one defect, which this ADR also fixes:
+the merkle batch path was paying a third of what the single-node path pays for
+an identical chunk, because the 3× settlement multiplier was never applied
+when pool commitments were built. Nobody had compared the two paths' arithmetic
+side by side before, which is the argument for having written this down.
+
+Terms: *record* (one stored chunk, the priced unit), *close group* (the 7 nodes
+closest to an address, which hold the replicas), *quote* (a node's signed
+offer), *commitment* (ADR-0002's signed Merkle root plus key count over what a
+node holds), *ANT* (the ERC-20 payment token on Arbitrum), *atto* (10⁻¹⁸ ANT).
+
+## Decision Drivers
+
+- One payment, stored indefinitely — the client transacts once and never again
+  for that data.
+- The price must be a public function anyone can recompute. No negotiation, no
+  oracle, no per-operator configuration.
+- The price must rise as the network fills, so scarcity attracts new supply,
+  and must be non-zero when empty, so storage is never free to spam.
+- Every storer must be able to verify payment independently, from the chain,
+  without trusting the client or the paying node.
+- Gas must not dominate: a large upload cannot mean one transaction per chunk.
+- Client and node must compute prices from the same code, so they can never
+  disagree about what was owed.
+
+## Considered Options
+
+1. **Per-byte pricing.** Rejected: the network's unit of work is a record —
+   replication, audit, and routing all cost the same per chunk regardless of
+   how full it is. Per-byte pricing prices the wrong thing and invites padding
+   games.
+2. **Rent / recurring payment.** Rejected: it makes permanence conditional on
+   a live payer, requires an eviction-for-non-payment path, and puts a
+   recurring on-chain cost on every stored object.
+3. **Pay every member of the close group.** Rejected on gas: seven settlements
+   per chunk. The chosen design pays one node a multiple instead, so the
+   network receives the same money for a seventh of the transactions.
+4. **Fixed network price.** Rejected: it carries no supply signal, so a
+   filling network cannot pay more to attract capacity.
+5. **Quadratic per-record price, median-of-group selection, one on-chain
+   settlement, re-verified by every storer (chosen).**
+
+## Decision
+
+We price and settle storage as follows.
+
+### 1. What is sold
+
+A **record** — one chunk of up to 4 MiB — is the unit of sale. The price is
+per record and does not depend on how many bytes the chunk actually carries.
+Payment is once; the data is then stored indefinitely. There is no rent, no
+renewal, no retrieval fee, and no emission or inflation subsidy: **client
+payments are the network's only revenue**.
+
+### 2. The price curve
+
+    price_per_record(n) = BASELINE + K × (n / D)²
+
+with `BASELINE = 0.00390625 ANT`, `K = 0.03515625 ANT`, `D = 6000 records`,
+computed in `u256` wei. `n` is the node's **committed** key count — the count
+in its signed storage commitment (ADR-0004) — not its raw on-disk count. A
+node with no live commitment can only quote the baseline.
+
+The formula lives once, in `ant-protocol`, and both sides import it, so a
+quote's price is recomputable by the client before paying and by every storer
+after (see `ant-protocol/src/payment/pricing.rs`).
+
+At $0.10/ANT and full 4 MiB chunks (256 chunks/GiB), the quoted price per GiB
+is roughly:
+
+| Committed records | ANT / chunk | Quoted $/GiB | Client pays (3×) |
+|---|---|---|---|
+| 0 (empty) | 0.0039 | $0.10 | $0.30 |
+| 6 000 | 0.0391 | $1.00 | $3.00 |
+| 12 000 | 0.1445 | $3.70 | $11.10 |
+
+Chunks smaller than 4 MiB cost the same each, so the effective $/GiB rises for
+small-chunk data.
+
+### 3. Quoting
+
+The client asks the **witnessed close group** — the 7 nodes closest to the
+address, each confirmed by a 5-of-7 quorum of its neighbours' views, so the
+client cannot be handed a fabricated group. The request carries the address,
+data size, type, and a fresh 32-byte nonce.
+
+Each responder returns, signed with ML-DSA-65: the **quote** (content address,
+timestamp, price, rewards address, committed key count, commitment pin), the
+**signed commitment** the price was derived from, and its **audit report** on
+the peers it observes near that address, bound to the request nonce so it
+cannot be replayed. A responder that already holds the chunk says so, and the
+client skips payment for that address entirely.
+
+Before paying, the client drops any quote it cannot fully resolve: wrong
+public-key-to-peer binding, bad signature, wrong content address, a price that
+is not exactly `calculate_price(committed_key_count)`, or a commitment that
+does not parse, does not belong to the quoter, or does not hash to the quote's
+pin. The audit reports feed the payee-eligibility gate, which excludes quoters
+their neighbours have not attested clean at the size they are monetizing.
+
+### 4. Paying — two shapes
+
+**Single-node (default below 64 chunks).** The 7 quotes are sorted by price;
+the client pays the **median-priced** issuer **3× its quoted price** and the
+other six nothing, in one `payForQuotes` call. Cost per chunk is 3 × median.
+The multiplier keeps the network's revenue equal to paying three of the group
+while costing one transaction's gas.
+
+**Merkle batch (64 chunks and above).** The client builds a Merkle tree over
+up to 256 addresses, derives `2^ceil(depth/2)` candidate pools from its
+midpoints, and collects 16 quotes per pool — one `payForMerkleTree` call for
+the whole batch. The contract selects a winner pool and pays `depth` of its 16
+candidates: `total = median16(amount) × 2^depth`, split evenly. The client
+applies the same 3× multiplier here, to the on-chain payable `amount` rather
+than to the signed quote, so a batch settles `3 × median16 × 2^depth`. The
+receipt is valid for one week; larger uploads split into successive batches.
+
+The two paths now price a **leaf** identically at 3 × median. **They did not
+until this ADR**: the merkle path submitted the bare quoted price as the
+payable amount, so a chunk stored through a batch earned its node one third of
+what the same chunk earned through a single-node upload. Nothing detected it,
+because the storer recomputed its expectation from the same un-multiplied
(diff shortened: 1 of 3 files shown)


dirvine (commented):
Reviewed current head `5582444c4dd00bdaa6d0c00dd8c24440de3fb633` with two independent waves of three reviewers, then verified the main findings locally.

**Verdict: changes requested before merge.** The helper arithmetic is sound and the dormant gate preserves 1× compatibility, but the rollout telemetry and descriptive ADR are not yet reliable enough for a T3 payment decision.

## Merge blockers

1. **Emit parity telemetry only after successful admission validation, and only for store admissions.**
   `src/payment/verifier.rs:2855-2860` logs before paid indices, reward addresses and amounts are validated at `:2877-2901`; it also logs for `PaidListAdmission` despite claiming one sample per Merkle store admission. A later-rejected proof can therefore enter the rollout signal. Move the call after the full paid-node loop and guard it with `context.is_store_admission()`.

2. **Correct the false exact-scaling test.**
   `merkle_expected_per_node_scales_with_the_parity_multiplier` assumes
   `floor(3*x/depth) == 3*floor(x/depth)`. That fails when division has a remainder (for example median 901, depth 7 differs by one wei). Production arithmetic correctly multiplies before division; make this test remainder-aware rather than encoding a stronger invariant.

3. **Separate shipped baseline from the proposed 3× client policy in `WithAutonomi/ant-node ADR-0008`.**
   The ADR currently describes several unmerged/planned behaviours as built/live:
   - witnessed support is nominally 5-of-7 but degrades with missing responder views, and payment accepts 1–7 quotes; it is not always seven quotes each confirmed by fixed 5-of-7;
   - audit reports/payee eligibility and the referenced ADR-0005/ADR-0007/client ADR work are not present at this head and should be labelled/linked as pending;
   - current `ant-client` main still submits bare Merkle candidate prices, so “client applies 3× — enforced” and the claimed live ≥3× Merkle guarantee are not current facts;
   - node quote-arithmetic rejection is off, non-paid SNP quotes are not fully re-verified, and current Merkle admission still accepts historical 1×;
   - “parity” here means aggregate settlement normalised per **padded tree leaf**, against the winner pool’s own median. Merkle pays `depth` batch candidates, not one storer per chunk, so this is not equality of recipient/node earnings.

Please make the ADR explicit about:
- **current baseline:** SNP uses 3× the supplied-set upper median; Merkle uses 1× its 16-candidate pool median per padded leaf before rounding;
- **normative proposal:** raise Merkle payable amounts to 3× while keeping signed candidate prices/pool hash unchanged;
- **alternatives:** retain the differential, lower SNP to 1×, raise Merkle to 3×, or defer pending economics evidence.

## Release/enforcement gates — not necessarily blockers for landing neutral shadow telemetry

The panel agreed the node half can land independently once the above is corrected. It must not be described as the live economic fix, and the gate should not be enabled, until there is:

- a linked paired client PR;
- an Anvil client → vault → node test proving 3× construction, settlement and acceptance (plus 1× / just-below-3× / redirect cases);
- old/new client × shadow/enforcing node coverage;
- a drain period of at least the one-week receipt lifetime after the last old-client 1× settlement;
- rollout telemetry based on successfully verified store admissions, preferably deduplicable by pool/settlement rather than raw high-cardinality per-xorname INFO lines;
- complete platform CI and the stated dev-testnet/adversarial evidence.

There was reviewer dissent on whether the paired client/E2E work must block this **shadow-node merge**; consensus was that it must block claiming parity is deployed or enabling enforcement, not necessarily neutral measurement code.

## Economic sign-off

Tripling the common large-upload path is a pricing-policy choice, not merely an arithmetic repair. Please record explicit approval from the economic owner/CEO (`@forthebux`) for upward 3× parity, the aggregate-per-padded-leaf invariant, and retention of the near-2× padding premium around power-of-two boundaries.

## Verification

- `cargo fmt --all -- --check` — passed
- `cargo test --lib payment:: --no-fail-fast` — **173 passed, 0 failed**
- Added focused helper tests — **5 passed**
- Most GitHub checks are green; macOS/Linux/Windows platform test jobs remain pending at review time.

No approval or merge performed.

dirvine (commented):
Re-reviewed exact head `ee09642c9bcf372246547c4b323096e4f51c0cbd` with two review waves and local verification.

**Verdict: the original three blockers are addressed, but changes are still requested because this head introduces a new hard-calendar enforcement design.**

## Original blockers — resolved

- Parity telemetry now runs after paid-index/address/amount validation and only for store-admission contexts (`src/payment/verifier.rs:2981-2995`).
- The integer-division tests now use the correct remainder-aware invariant, including the 901/depth-7 regression.
- `WithAutonomi/ant-node ADR-0008` now separates shipped baseline, pending work and the normative 3× proposal, and accurately records 1–7 quotes, degraded witness quorum, non-paid-quote limits and current enforcement state.

## New blockers / decision gates

1. **The fixed activation date is not release-derived or operationally safe.**
   `MERKLE_PARITY_ENFORCED_FROM_UNIX = 1_785_855_600` is **2026-08-04 15:00 UTC**. The comment correctly says the boundary must not precede the release carrying it, but nothing enforces that invariant. If publication or fleet rollout slips, a newly deployed binary immediately rejects fresh 1× receipts bought while no enforcing release was available. Select/freeze the date only after the paired client release and fleet schedule are committed, or provide a coordinated activation mechanism rather than an imminent date in generally releasable code.

2. **Backdating is bounded, not impossible.**
   The timestamp is client-selected. A modified client can pay after the nominal boundary while stamping `boundary - 1`, and the proof remains valid until one receipt lifetime later (approximately **2026-08-11 15:00 UTC**). Expiry then closes the route; it does not make “It cannot be gamed” true during the grace week. Please document this as an intentional bounded 1× compatibility/arbitrage window, or derive cutover from a non-client-controlled value.

3. **Rollout order does matter after the boundary.**
   The compatibility table itself says old client + enforcing node + fresh post-boundary receipt is refused. During partial node rollout, the same proof can be accepted by old nodes and refused by new ones. Replace “neither half has to land first / partial rollout cannot strand uploads” with explicit phases: client publication/adoption first, then node enforcement after the approved threshold and mixed-fleet evidence.

4. **Economic sign-off remains outstanding.**
   The ADR is Proposed, reviewers are pending, and the code now schedules hard rejection rather than shadow measurement. Explicit approval from the economic owner/CEO (`@forthebux`) must cover the 3× price direction, padded-leaf premium, hard-enforcement model, bounded backdating window and activation date before merge if merging creates a releasable artefact—or unequivocally before release otherwise.

## Warning: telemetry is keyed, not deduplicated

Including `pool_hash` makes downstream deduplication possible, but the node still emits once per successfully verified chunk: there is no pool-keyed emission cache, and every store admission reaches `log_merkle_parity`. The comment claiming “one line per batch per storer” is false. Either add bounded first-emission-per-pool state or state plainly that consumers must deduplicate and retain the expected per-chunk log-volume caveat.

## Evidence / release gates

- Local `cargo test --lib payment:: --no-fail-fast`: **179 passed**.
- Node boundary and mixed-regime unit tests pass.
- Current macOS/Linux CI is green; Windows test is still pending.
- Paired client-PR → vault → **this enforcing node head** evidence is still absent. Client #161’s Merkle E2E uses released `ant-node 0.15.0`, which only proves a 3× client remains compatible with an older 1×-floor node.

No approval or merge performed.

dirvine (commented):
Re-reviewed current head `3027de1ef917bbe3042cb62ce2f46e561ed88ab5` after the follow-up commits.

**Technical verdict: the prior code/ADR findings are addressed.** The pool-keyed telemetry cache is bounded and race-safe, is populated only after successful payment validation, and suppresses paid-list/rejected proofs. Local `cargo fmt` plus `cargo test --lib payment:: --no-fail-fast` passed (187 tests); all current node CI checks are green across Ubuntu, macOS and Windows.

I am not posting an unqualified approval yet because:

1. **The PR body is stale and contradicts the corrected ADR.** It still says either order is safe, rollout is free, and backdating is impossible. The current design is client-first; a modified client can backdate during the bounded one-week grace; old clients with fresh post-boundary 1× receipts are refused by enforcing nodes. Please update the PR body so the public rollout record matches `WithAutonomi/ant-node ADR-0008`.
2. The hard boundary remains `2026-08-04 15:00 UTC`; it must move if client publication slips.
3. Paired client → vault → this enforcing node evidence remains a release gate.
4. The 3× price, padded-leaf premium, hard enforcement date and bounded underpayment window still require `@forthebux` economic sign-off.

No merge performed.

dirvine (approved):
Engineering approval at head `3027de1ef917bbe3042cb62ce2f46e561ed88ab5`. Verified payment arithmetic, timestamp compatibility path, post-validation once-per-pool telemetry deduplication, bounded concurrency behaviour, ADR implementation consistency, focused tests and green platform CI. This is code-review approval only; it is not economic sign-off, ADR acceptance, release/activation authorisation or merge authorisation.

PR #192, opened 2026-08-04
feat(payment): enforce the ADR-0004 quote arithmetic check

grumbach:
## Linear issue
https://linear.app/autonominetwork/issue/V2-817/turn-on-the-two-quote-enforcement-gates-once-the-fleet-has

This PR is the first of the two gates that issue covers. The second
(`QUOTE_COMMITMENT_MISMATCH_TRUST_ENABLED`) stays off here, deliberately: it carries
a trust penalty rather than being reject-only, so it gets its own read of the
observe-only logs before flipping.

## Risk tier

Payment admission behavior changes: a quote whose signed price is not exactly on
the public pricing curve is now rejected instead of logged.

## Compatibility
- Wire: none. `committed_key_count` and `commitment_pin` have shipped in the quote
  since ADR-0004 and are already covered by the signature and quote hash. No new
  field, no new message, no size change.
- Storage: none.
- API: none. The gate is a private crate constant.

## Semver impact
- [ ] breaking
- [x] feature
- [ ] fix

No public API changes; this enables node-side enforcement that was previously
observe-only. Flagged as a feature rather than a fix because it changes what a
node accepts, not because anything was incorrect before.

## Test evidence

**Production canary, ant-prod-01, 2026-07-29 to 2026-08-04.** The check has been
running observe-only against every quote in every verified bundle. Over the window
it flagged nothing:

| signal | count |
| --- | --- |
| verified bundles evaluated | 70,078 |
| node instances reporting | 912 of 913 (99.9%) |
| off-curve prices observed | 0 |
| invalid binding shapes observed | 0 |
| price/count mismatches observed | 0 |

Measured by querying the `ant_node::payment::verifier` log target for the
`log_off_curve_single_node` warnings. WARN-level records from payment modules do
reach the log store (451 under `ant_node::payment::verify` on the busiest day), so
the zero is a genuine absence rather than a collection gap.

Honest traffic therefore sits exactly on the curve, and enforcing rejects nothing
that is accepted today. What the production data does *not* establish is detection
power: prod contained no off-curve quotes, so the rejection path was never
exercised there. That side is covered by unit tests.

**Unit tests.** 768 pass, 0 fail. Getting there took more than the flag flip, and
the reason is worth a reviewer's attention.

Flipping the gate initially broke **30 tests**, none of them a product problem: the
payment test fixtures build quotes whose `committed_key_count` does not match their
`price`, so they were never valid under the ADR-0004 binding rule and only passed
because the gate was off. Specifically:

- `make_fake_quote` priced every quote at a placeholder `1` while claiming
  `committed_key_count: 0`, whose curve price is `3906250000000000`.
- `make_fake_quote_at_records(n)` set `price = calculate_price(n)` but left the
  count at `0`, i.e. a quote charging as though it stored `n` keys while committing
  to storing nothing. That is exactly what the gate is meant to reject, so the
  fixture named "honest bundle" was not honest.
- `make_signed_quote` had the same split, so it now derives the binding from the
  price.
- The merkle candidate fixture priced at a placeholder `1024`, and the depth-2
  settlement constants (`6144`, `2048`) were scaled to that placeholder. Both are
  now derived from the curve so they track it instead of drifting.

Product code is unchanged apart from the const. Every edit above is test-only.

Two tests changed meaning rather than just fixtures, flagged explicitly:

- `adr0004_observe_only_does_not_reject_off_curve_quote` was gated on the const
  being `false`, so after the flip its body would never execute. It now asserts
  both branches and is renamed `adr0004_off_curve_quote_follows_rollout_gate`.
- `test_legacy_zero_price_median_rejected`: `Amount::ZERO` is not on the curve, so
  the arithmetic gate now rejects the bundle before median selection and the
  dedicated zero-price check is unreachable for single-node quotes. The gate
  subsumes it. The test accepts either rejection and the docstring says why.

**Follow-up from review: the pay-yourself fixture's settlement was wrong.**
`test_attack_merkle_pay_yourself_fabricated_pool` funded its fabricated
`OnChainPaymentInfo` at `price * 2^depth`, dropping the `/ depth` term from the
formula the verifier applies (`multiplier * median16(price) * 2^depth / depth`).
At depth 2 that is `2 * price` under the legacy rule and `6 * price` from
`MERKLE_PARITY_ENFORCED_FROM_UNIX` onward, so the fixture's `4 * price` was
neither and underpaid on the parity side.

The test stayed green because candidate closeness is checked before cached
settlement, but the fixture had stopped satisfying its stated invariant: it
claims to be funded so closeness is the only check that can reject the proof,
and an unrelated underpayment rejection sat behind it. Had closeness been
bypassed or regressed, the test would have passed for the wrong reason.

It now funds at the parity amount, derived from the curve rather than a literal.
Since the verifier only rejects underpayment, the parity amount also clears the
legacy expectation, so the fixture is correct on both sides of the boundary
without pinning it, and a unit test restates the depth-2 case as literals so a
future drift fails loudly.

## New dependency
none

## ADR
https://github.com/WithAutonomi/ant-node/blob/main/docs/adr/ADR-0004-commitment-bound-quote-pricing.md

This is the rollout gate ADR-0004 specifies, moving from its observe-only stage to
enforcement. See also
[the implementation slices](https://github.com/WithAutonomi/ant-node/blob/main/docs/adr/implementation-slices-adr-0004.md).

## Mitigation / rollback
Flip `QUOTE_ARITHMETIC_RECHECK_ENABLED` back to `false` and ship.

In source terms that is as cheap as a rollback gets: a single crate constant, no
persisted state, no wire effect, so it is a one-line revert with no migration and
nothing to undo on disk or on the network. It is worth being plain that the
*source* change is the cheap part. There is no runtime switch here, so getting
the revert into the field means a rebuild, a release and a fleet restart, on the
same timeline as any other node release. Plan for release latency, not for a
flag flip.

What limits the blast radius in the meantime is the gate's shape rather than the
speed of the rollback. It is reject-only per ADR-0004: an off-curve quote
produces no trust evidence and schedules no audit, so a misbehaving gate cannot
feed the trust engine or evict peers. The worst case is refused stores on the
affected quotes until the release lands, not accumulated trust damage that
outlives it.

## Note for reviewers

This is deliberately only the quote-arithmetic half. It does **not** address a
client shopping the closest group for the cheapest quote, which is a separate
economic question. The receiver-side price floor from #176 is **not** enabled here
and, on the same production data, should not be: honest stores already fall below
it 0.377% of the time because node prices legitimately vary about 4x across the
fleet.

### DIFF
--- a/docs/adr/ADR-0008-storage-economics-and-payment-protocol.md
+++ b/docs/adr/ADR-0008-storage-economics-and-payment-protocol.md
@@ -306,16 +306,24 @@ address, its ML-DSA-65 signature, and the on-chain settlement — which must be
 at least 3× the median **and recorded for the quote's own rewards address**, so
 a client cannot redirect the money to itself and still be admitted.
 
-Two limits of that check, both current behaviour:
+ADR-0004's `price == calculate_price(committed_key_count)` recheck **is
+enforced**, over every quote in the bundle rather than the paid one alone. A
+quote whose signed price is not exactly the curve value for its own signed
+count is rejected, as is a quote whose binding shape is invalid. The gate is
+reject-only: an off-curve quote produces no trust evidence and schedules no
+audit.
+
+Two limits of that check remain, both current behaviour:
 
 - **Non-paid quotes are not fully re-verified.** They position the median but
   their signatures and content addresses are not checked, so a bad signature on
   an unpaid quote is accepted (`test_legacy_unpaid_quote_bad_signature_accepted`
   pins this). They still influence which quote gets paid.
-- **The on-curve price check is off.** ADR-0004's
-  `price == calculate_price(committed_key_count)` recheck over every quote is
-  rollout-gated by `QUOTE_ARITHMETIC_RECHECK_ENABLED`, which is `false`; today
-  off-curve quotes are only logged.
+- **The signed count is not resolved against the peer's own commitment.** The
+  arithmetic gate forces the price to the count the quoter signed, but a quoter
+  can still sign a count it does not store. Resolving the pin and reporting the
+  contradiction is `QUOTE_COMMITMENT_MISMATCH_TRUST_ENABLED`, which is `false`;
+  today such mismatches are only logged.
 
 Merkle proofs are checked against the pool's on-chain record: every candidate's
 signature, the pool's closeness to the midpoint, the tree proofs, and each paid
@@ -349,7 +357,7 @@ and a node that was actually paid is nominated for a first audit.
 | Client resolve-before-pay quote binding | **enforced** |
 | Node re-verification of the **paid median** quote (signature, content, K-closeness, settlement, payee address) | **enforced** |
 | Node re-verification of **non-paid** quotes in the bundle | not done (bad signature on an unpaid quote is accepted) |
-| Node rejection of off-curve quotes (`QUOTE_ARITHMETIC_RECHECK_ENABLED`) | off (telemetry only) |
+| Node rejection of off-curve quotes (`QUOTE_ARITHMETIC_RECHECK_ENABLED`) | **enforced**, on every quote in the bundle |
 | Quote/commitment mismatch reported to trust | off (telemetry only) |
 | Receiver-side price floor (ADR-0006) | shadow (`ANT_PRICE_FLOOR_ENFORCE`, 50% tolerance) |
 | Payee eligibility gate (ADR-0005) | **not merged**; observe-only on its branch (`ADR5_ENFORCE`) |
--- a/docs/adr/implementation-slices-adr-0004.md
+++ b/docs/adr/implementation-slices-adr-0004.md
@@ -20,9 +20,10 @@ already specifies.
 non-negative integer `n`, by exact recomputation, on every quote in every
 payment bundle (all 7 single-node quotes and all 16 merkle candidates), in
 every `VerificationContext`. Reject-only when enforced; no trust evidence.
-Rollout-gated by `QUOTE_ARITHMETIC_RECHECK_ENABLED` (defaults to `false` —
-observe-only). Telemetry runs only after ML-DSA-65 signature verification has
-passed, so unauthenticated peers cannot poison rollout logs.
+Rollout-gated by `QUOTE_ARITHMETIC_RECHECK_ENABLED`, now `true` after a clean
+observe-only canary on ant-prod-01. Telemetry runs only after ML-DSA-65
+signature verification has passed, so unauthenticated peers cannot poison
+rollout logs.
 
 **Why first:** needs no evmlib change, no new state, no new wire types, no
 new gossip; it is the ADR's "every storer re-runs the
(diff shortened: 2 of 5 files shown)


dirvine (changes requested):
## Hermes review — changes requested

The production gate itself looks sound, and the new commit fixes the original off-curve E2E candidate. One test-fixture issue remains on current head `0924d07463b2586c0541212b39162de212a04581`.

### Blocking: parity settlement in the isolated pay-yourself fixture is wrong

`tests/e2e/merkle_payment.rs:588-592` says the fabricated proof is funded so every check except candidate closeness passes, but sets:

```rust
let per_node = candidate_price() * Amount::from(4u64);
```

The verifier's actual formula is:

```text
multiplier × median_price × 2^depth / depth
```

For depth 2 this is `2 × price` in the legacy 1× window and **`6 × price`** after `MERKLE_PARITY_ENFORCED_FROM_UNIX` (`2026-08-04 15:00 UTC`). The fixture's `4 × price` is therefore neither the documented legacy formula nor sufficient under the 3× rule.

Candidate closeness currently runs before cached settlement validation, so the expected rejection can still make CI green. But the fixture no longer satisfies its stated isolation invariant: if closeness is bypassed or regresses, post-boundary verification has an unrelated underpayment rejection available before the proof could demonstrate the attack.

Please fund the fake `OnChainPaymentInfo` at the parity-safe amount (`candidate_price() * 6` for depth 2), or explicitly pin the verifier's parity boundary in this test and derive the corresponding expected amount. Update the comment to include both the multiplier and `/ depth` term. Then require all three platform Test jobs green on the resulting exact head.

### Non-blocking follow-ups

- `docs/adr/ADR-0008-storage-economics-and-payment-protocol.md:315-318,352` still describes this gate as off/telemetry-only.
- `src/payment/verifier.rs:1592-1600` is pre-existing stale rustdoc saying the signed count/pin fields do not exist.
- Rollback is a one-line source revert with no migration, but operationally requires rebuild, release and fleet restart; the PR wording could state that more plainly.

### Evidence

- Local exact-head format check passed.
- Focused `merkle_required_multiplier` test: 1/1 passed.
- Focused `merkle_expected_per_node` tests: 5/5 passed.
- Previous head: local library suite 768/768 and strict clippy passed.
- Review panel: five substantive seats completed; repeated replacement seats timed out and are recorded as no opinion. Completed seats agreed the production arithmetic/binding implementation is sound; release/CI review identified the fixture migration gap.

dirvine (approved):
Re-reviewed current head `086cb47c36a248c4d1567750c95c56ca556e6f33` after the requested changes.

The previous blocker is resolved: the fabricated pay-yourself Merkle fixture now derives the parity settlement as `3 × price × 2^depth / depth`, giving `6 × price` at depth 2, and the literal regression test pins that formula. The candidates are also priced on-curve, so the arithmetic gate no longer masks the intended closeness rejection.

Verified locally:
- `cargo test -q payment::verifier --lib --no-fail-fast` — 124/124 passed
- `cargo test -q --lib --no-fail-fast` — 905/905 passed
- focused parity helper and pay-yourself E2E tests — passed

The complete local E2E run reached 88 passed / 6 failed / 3 ignored; all six failures were harness setup errors from macOS `Too many open files`, not failures in this PR's paths. The corresponding GitHub Test jobs pass on macOS, Ubuntu and Windows.

All 15 GitHub checks are green on this exact head. Six valid independent review seats reached the same conclusion; no material correctness or security blocker remains. One stale-checkout dissent was excluded after live GitHub and `refs/pull/192/head` confirmed it inspected the wrong SHA.

Approved. This review does not merge the PR.

PR #193, opened 2026-08-04
feat(payment): price the revenue floor against the close-group median

grumbach:
## Linear issue
https://linear.app/autonominetwork/issue/V2-696/receiver-side-price-floor-rollout-canary-plan-and-pre-canary-work-ant

## Risk tier

Changes which payments a node admits. No protocol or format change.

## Compatibility
- Wire: none. Quotes, their fields and their signatures are untouched. This is
  receiver-side admission logic reading state the node already holds.
- Storage: none.
- API: none. `PriceFloorConfig` keeps its shape; only the default tolerance moves.

## Semver impact
- [ ] breaking
- [x] feature
- [ ] fix

## What this changes

One thing: **what a node compares an incoming payment against.**

The floor used to compare the settled amount to the *receiving node's own*
commitment-bound price. It now compares against the **median committed key count
of the receiver's close group** — the neighbours' signed commitments it already
holds from gossip, plus its own.

A client pays `3x` the **median** of the quotes it collected. Comparing that to
one node's own price compares a median to a single draw from a distribution
spanning ~4x, which is why the old reference failed in both directions at once.

**Read this as "the floor can now be measured", not "the floor now works".** It
ships shadow-only and is not safe to enforce; see Residual risks.

## Test evidence

**The problem, measured.** ant-prod-01, target `ant_node::payment::price_floor`,
2026-07-29..08-04, 70,078 evaluations across 912 of 913 node instances, all shadow:

- fired on **264 honest stores (0.377%)**, against zero in ~350k staging evaluations
- systematic, not noise: rejections landed on the fullest nodes (rejecter median
  committed count 5,365 vs fleet median 3,091) and missed by a median of 9%
- enforcing at the then-default 50% would have cost 229 of 18,258 chunks a replica
  and refused 16 chunks at **every** node that evaluated them — upload failures
  after the client had already paid, with no refund path
- and it still admitted the underpayment it targets at **62.9% of receivers**

Replaying the floor formula over all 70,078 records reproduced the nodes' own 264
`below_floor=true` decisions exactly, so those figures are measured, not modelled.

**Calibration.** Simulated over the measured price distribution, with the client
paying `3x` the median of a `CLOSE_GROUP_SIZE`-quote bundle and the receiver at the
minimum permitted view (15 neighbours + own):

| tolerance | honest rejected | underpayment caught |
|-----------|-----------------|---------------------|
| 50%       | 0.000%          | 6.4%                |
| **65%**   | **0.140%**      | **89.8%**           |
| 70%       | 0.338%          | 96.0%               |
| 75%       | 1.175%          | 98.1%               |

Against the old reference's measured 0.377% / 37.1%, both axes improve — by
roughly **2.7x** on rejections, not the order of magnitude an earlier revision of
this PR claimed. That claim came from a simulation that wrongly modelled the client
as paying the median of 20 quotes; the bundle carries at most 7, and a 7-sample
median is far noisier. Sample size is itself a safety parameter: below ~11 known
neighbours this policy is *worse* than the one it replaces, hence the gate at 15.

**Unit gates.** 919 pass, 0 fail, `cfd` and `--no-default-features` clean.

Reference and anti-manipulation:
- `group_reference_price_is_the_close_group_median` — pins the reference exactly;
  this node's own extreme count does not move it
- `price_floor_prices_against_the_group_median_not_the_receiver` — the production
  false-rejection shape: a receiver far above the group median admits an honest
  at-median payment
- `price_floor_rejects_cheapest_of_group_payment` — the underpayment is rejected
- `price_floor_reference_ignores_client_supplied_quotes` — a bundle padded with
  cheap quotes does not move the floor

Availability (all must SKIP, never reject):
- `price_floor_skips_when_group_view_is_too_thin` (one short of the gate, built from
  the *dearest* counts so deleting the gate flips it to a rejection)
- `price_floor_evaluates_at_exactly_the_minimum_sample` (the other side of the gate)
- `price_floor_skips_when_this_node_has_no_commitment`
- `test_price_floor_without_any_group_view_skips` (a receiver that *would* have
  rejected under the old rule)
- `price_floor_ignores_stale_group_commitments` and
  `price_floor_excludes_only_the_stale_entries_from_a_mixed_cache`
- `price_floor_ignores_commitments_above_the_protocol_cap`

Known limits, pinned deliberately:
- `group_median_resists_three_understated_commitments` /
  `group_median_is_disarmed_by_four_understated_commitments`
- `group_median_resists_three_overstated_commitments` /
  `group_median_is_griefed_by_four_overstated_commitments`

The last four assert a **weakness** on purpose, so if a future change improves the
bound they fail and the number gets re-derived rather than quietly dropped.

Pre-existing floor tests were migrated onto the group reference rather than left to
pass vacuously — with no group view the floor skips, so an unmigrated test would
have gone green for the wrong reason.

## New dependency
none

## ADR
https://github.com/WithAutonomi/ant-node/blob/main/docs/adr/ADR-0006-receiver-side-revenue-floor.md

ADR-0006 amended in place. The plain price floor it originally proposed is not a
viable design, so a reader arriving there needs to learn that the approach was
tried and does not work — not be sent to a separate ADR alongside it. The
amendment adds an **"Attempted and failed"** section recording exactly what the
canary measured and why node size difference makes a per-receiver floor
unworkable, then replaces the Decision with close-group pricing.

## Mitigation / rollback
Default is unchanged shadow mode: the floor is computed and logged, never enforced,
until a node opts in with `ANT_PRICE_FLOOR_ENFORCE=1`. Unsetting it and restarting is
the kill switch, though in practice that means a rebuild, release and fleet restart
rather than a live toggle. Rollback of the code is a revert with no migration and no
wire effect. Floor rejections remain economic declines that never feed trust or
misbehaviour scoring.

## Notes for reviewers

**Where the reference comes from, and why that is the security property.** Built
only from this node's own routing view and its neighbours' signed, TTL-fresh
gossiped commitments. Nothing in the payment bundle contributes, so a client cannot
pad, prune or reorder quotes to move the floor.

**Residual risks, stated plainly. These are why enforcement stays off:**

- **The manipulation bound is about a quarter of the sample, not a majority.** An
  earlier revision of this PR claimed a majority was needed. That was wrong: the
  floor does not need the median to reach a liar's value, only to move past the
  tolerance headroom, and each liar shifts the order statistic one rank. Measured on
  the fixture, **4 of 15** suffices either way — overstating rejects an honest
  at-median payment (griefing: the attacker spends nothing and destroys already
  settled money), understating admits the underpayment. Both are pinned by tests.
- **One-quote liveness is in tension with any floor.** An honest client that could
  reach only one cheap quoter is indistinguishable from one that chose the cheapest
  of twenty. Shadow mode does not have this problem; enforcement does.
- **Availability of the reference is unmeasured.** The gate assumes a settled node
  normally knows most of its neighbours. The K-closest scan returns at most 20
  entries *including self* and the cache never holds self, so the ceiling is 19: the
  gate tolerates 4 unknown or stale neighbours. The `skipped=true` /
  `skip_reason` telemetry exists to measure this before anyone enforces —
  `no_commitment_cache` and `no_routing_view` are wiring bugs, `thin_sample` is the
  signal that the gate exceeds what the network supplies.
- Simulation caveats: 97% of the production evaluations come from one busy day, and
  the tables above are simulations over the measured distribution, not measurements.

**Considered and declined:** deduplicating the repeated local K-closest lookup.
`local_k_closest_peer_ids` and `validate_paid_quote_issuer_k_closest` look identical
but differ in failure behaviour — one skips, the other hard-rejects in production
builds. Collapsing them would thread a `Vec<PeerId>` through three signatures or risk
flattening that asymmetry into the payment-rejection path, to save a local
routing-table scan. Two independent reviewers agreed.

**Not in scope.** This does not change what nodes quote, so it does not address
whether fuller nodes should earn more than emptier ones. That is an emissions
question, not an admission one. It also does not cover the merkle-batch lane, which
never reaches this floor — canary results should not be read as "underpayment
stopped".


dirvine (commented):
## Hermes review — head `32e0b90`

**Verdict: no code blocker for the shadow-only merge.** Six independent review seats converged on the implementation: the group reference is built from identity-bound, TTL-fresh commitment gossip plus the receiver's live commitment; the lower-median/sample gate, skip-safe failure modes, store-admission scope, arithmetic, lock discipline and no-logging build are sound. The self-double-count concern is ruled out because commitment ingest requires routing-table membership and saorsa-core refuses to insert self.

I would correct these documentation inconsistencies before merge:

1. **`docs/adr/ADR-0006-receiver-side-revenue-floor.md:72-80`** — counts `1,700 → 6,300` span about **3.71×**, but under the pinned quadratic pricing formula the corresponding prices are `6,728,515,625,000,000 → 42,666,015,625,000,000 wei`, about **6.34×**, not ~4×/4.1×. This strengthens the stated signal-vs-noise argument, but the price-spread claim itself is wrong.
2. **`src/payment/verifier.rs:900-901`** — says a missing local commitment source falls back to baseline pricing; the new code skips with `NoOwnCommitment`.
3. **`src/payment/verifier.rs:1409-1413`** — says the settled amount is compared with the receiver's own commitment price; it is now compared with the close-group reference.
4. **`src/payment/verifier.rs:1650-1656`** — says every skipped evaluation logs `group_sample=0`; `ThinSample` deliberately logs its real non-zero sample counts.
5. **`docs/adr/ADR-0008-storage-economics-and-payment-protocol.md:354`** — the living enforcement table still says the receiver floor is at 50% tolerance; this PR changes the default to 65% and the basis to the group median.

Non-blocking enforcement watch-outs remain correctly disclosed: the fixture demonstrates 4/15 manipulation in either direction, one-quote liveness conflicts with any enforced floor, Merkle admissions remain outside this floor, and the production-derived simulation figures are not reproducible from repository evidence alone. None of these make shadow telemetry unsafe.

Verified locally on the exact head:
- `cargo test -q --lib --no-fail-fast` — **919 passed**
- `cargo test -q payment::verifier --lib --no-default-features --no-fail-fast` — **138 passed**
- `RUSTFLAGS='-D warnings' cargo check --release --no-default-features` — clean
- `cargo fmt --all -- --check` / `git diff --check` — clean
- GitHub: **15/15 checks passed**, including all OS jobs, Clippy, audit, ADR governance and no-logging

Review-team result: **6/6 completed; no code blocker.** One governance dissent preferred a separate superseding ADR, but the repository's written immutability rule applies only to `Accepted` ADRs; ADR-0006 remains `Proposed`, so amending it is permissible.

dirvine (approved):
Approved at head 1e4f578.

Re-review of the fix commit (was 32e0b90):

All five documentation corrections from the prior review are addressed in 1e4f578 (docs/ADR + verifier.rs doc-comments only, no behaviour change):
- Price spread: ADR-0006 now states both figures correctly - 6.3x end-to-end on the quadratic curve and 4.1x between the 5th/95th percentiles, no longer pairing the committed-count range with the wrong ratio.
- attach_local_commitment_source: now documents SKIP with no_own_commitment rather than baseline-pricing fallback.
- verify_legacy_median_candidate: now documents comparison against the close-group median, not the receiver's own commitment price.
- enforce_price_floor: now documents real neighbours/group_sample counts plus skip_reason, not a blanket group_sample=0.
- ADR-0008 living table: floor updated to 65% tolerance, priced against the close group's median; own-price basis marked withdrawn.

Verification:
- All 15 CI checks green on this head. The single Windows Test failure was a testnet transport/bootstrap setup flake in scenario_12_bootstrap_quorum_aggregation (failed before any test logic ran); confirmed by a successful rerun of the same job.
- Diff is docs-only plus verifier.rs doc-comments (19 changed lines, no code behaviour change).

No code blocker; previously raised items all resolved.

PR #194, opened 2026-08-05
fix(replication): scope round-2 audit openings and apply the audit wire ceiling on encode

grumbach:
## Linear issue

- https://linear.app/autonominetwork/issue/V2-869

## Risk tier

T2 rather than T3, and the boundary is worth stating since the parent PR was T3. Nothing here changes the wire format, the proof shape, stored data, economics or the upgrade mechanism. What does change is responder behaviour: a round-2 challenge that was previously served is now refused if it names a key outside the proved subtree, and an oversized audit body now fails at the sender instead of at every receiver. That is behavioural surface, hence T2 rather than T1.

## Compatibility

- Wire: none. No message shape, field or protocol id changes.
- Storage: none.
- API: additive within an already-public module. `family_of_variant` (crate-internal) now returns `Option<BodyFamily>`; a `ceiling_for` helper is added beside it.
- Docs: ADR-0009 records the two bounds this PR adds. It is still `Proposed`, so amending it is what governance asks for rather than a new ADR.

## Semver impact

- [ ] breaking
- [x] feature
- [ ] fix

Both changes are defect fixes in behaviour, but one alters a crate-internal signature (`family_of_variant` now returns an `Option`). On a `0.x` crate that is a minor bump, which is what `feature` delivers.

## Test evidence

All figures below are from the rebased tree on current `main` (`0.17.0`, after the close-group price floor and the quote-arithmetic enforcement landed).

- `cfd` clean: clippy `-D warnings` across all targets and features, fmt, `cargo doc --deny=warnings`. `--no-default-features` clippy clean too.
- 921 lib tests, including the round-2 scope check, the shared wire-ceiling selector and the undeclared-discriminant classification.
- `poc_audit_handler_live` 16/16: the production responder against real LMDB, including an out-of-subtree key refused with its bytes present, the same refused with its bytes absent (so the refusal is shown not to depend on storage contents), and a challenge mixing an authorised with an unauthorised opening refused whole.
- `poc_commitment_audit_attacks` 19/19.
- Subtree-audit e2e 5/5, real multi-node QUIC and LMDB: honest node passes, data-deleting node caught, repeated audits with no false positive, and both first-audit drainer cases. Run after the rebase, not carried over from before it.

**Both fixes were verified to fail without the change they cover.** Each production line was reverted, the test observed to break, and the line restored. Reverting the scope check produces a served `Present` opening for a key round 1 never proved; reverting the encode ceiling lets a 6.9 MB audit body encode.

## New dependency

None.

## ADR

https://github.com/WithAutonomi/ant-node/blob/main/docs/adr/ADR-0009-audit-proof-shape-and-protocol-families.md

No decision changes. ADR-0009 governs the subtree audit's proof shape, its responder work bounds and its wire ceilings, and both fixes are bounds inside what it already decided, not departures from it. It is amended here rather than superseded, for two reasons: it is still `Proposed`, so governance permits editing it, and it enumerates the responder work bounds explicitly, so leaving the round-2 scope rule out would describe a responder that no longer exists. The amendment adds that round 2 may open only the leaves round 1 proved, that the wire ceiling is selected by one shared rule on encode as well as decode, and what the new tests pin.

## Mitigation / rollback

Revert the branch. No stored data changes and no migration. The two changes are independent, so either commit can be reverted on its own. The round-1 pool size is already config-driven, so admission pressure can be relieved without a code change.

---

## What this is

Two of the three findings reviewers raised against #181 and agreed were follow-ups rather than merge blockers. Both are responder-side hardening. The third is deliberately excluded; see below.

**Round 2 authorised any key in the pinned commitment**, not just the subtree round 1 proved. A cheap round 1 over small records could therefore authorise openings against large records elsewhere, so round 2's cost was not bounded by what the caller had already paid for.

Stated precisely, because the one-line version reads worse than the fact: the keys were always ones the responder had committed to, and nothing outside the commitment was ever served. `main` has the identical whole-commitment gate today and serves full chunk bytes for it. An honest auditor cannot reach the case either, since verifying an opening needs the round-1 leaf and it has none outside the subtree.

**The audit wire ceiling applied on decode but not on encode**, so an oversized audit body would serialise and then be silently dropped by every receiver — and the sender scored for the resulting non-answer. Both directions now share one selector rather than two hand-synced matches. Separately, an undeclared discriminant used to classify as "core" and inherit the larger allowance; it now classifies as nothing, and nothing takes the strict ceiling.

## The third finding is deliberately NOT here

Reviewers also raised that round-1 work is bounded by bytes but not by rate: nothing bounds a chunk from below, so a commitment of small records runs a full subtree of lookups per audit while charging almost nothing, and a caller rotating peer identities never trips the per-peer cooldown.

I implemented a responder-wide rate bucket for that and **withdrew it before opening this PR**, because review showed it made audit suppression cheaper rather than harder. Keeping both heavy workers busy needs a sustained stream of full proofs; holding a global rate bucket empty needs roughly one small proof per second, and the caller picks the audit nonce, which picks the proved subtree. Since a refused audit becomes a graced timeout, a globally shed refusal suppresses every audit of that node for free.

Answering a spent budget with a rejection, to make suppression cost something, was also tried and reverted: the rejection grades as a timeout, which is graced, so it never penalises a deleter, and holder credit expires on its own regardless — while an intermittent flood could use it to zero an honest node's credit cheaply.

The conclusion, reached independently by several reviewers, is that a global limit cannot be added safely without deciding *who* may spend it. That needs an eligibility or fair-share admission rule, which is a design change deserving its own PR and its own review rather than being bolted on here. Tracked on the Linear issue.

## Review

This went through several review rounds with independent reviewers, before and after the rebase onto current `main`. They corrected the analysis more than once, found a concurrency race the exactness tests structurally could not see, and showed that two tests proved nothing. The final round is why the rate budget is not in this PR.

Two claims of mine were also walked back rather than left standing: that the absent-bytes test proves storage was never read (it proves the refusal does not depend on storage contents, which is weaker and is what the comment now says), and that an unclassified body would "fail loudly" at encode (only if it exceeds the strict ceiling).

A final adversarial pass was run on the rebased tree, aimed at the two ways this PR could be worse than what it replaces. Both were cleared by tracing the code rather than reading the diff:

- **Can the new refusal fire on an honest auditor?** No. It is a confirmed failure for the responder if it ever does, so this is the item that matters. One nonce is drawn, carried into round 1, held in the audit context and copied byte for byte into round 2; the responder's single-use session matches source, challenge id, commitment hash and nonce before the handler runs; both rounds resolve the same pinned commitment through retained immutable state, so rotation cannot substitute the current tree; and the plan is a pure function of tree and nonce. Every way of missing (no session, expired, restart, rotated-out pin) lands on `Transient` or `UnknownCommitment`, not on the new rejection. Reaching it needs local state corruption or a future caller that stops sampling from the round-1 proof.
- **Can the encode ceiling refuse a legitimate audit message?** No, with roughly 4x headroom on the largest and 23x on the round-2 reply: challenge under 128 B, round-1 proof about 109 KiB at the 1,024-leaf cap, slice challenge under 0.5 KiB, slice response about 22 KiB for ten openings over maximum-depth 4 MiB chunks, against a 512 KiB ceiling. Slice sizes grow with the log of the chunk size, so this does not tighten as chunks get larger.

It also found one stale comment, now fixed: `is_held` claimed to be the same predicate the round-2 responder applies. It is now a strict upper bound on it, which is the direction the pruner needs, and narrowing it to match would be unsound because the nonce is not known when the pruner runs.

### DIFF
--- a/docs/adr/ADR-0009-audit-proof-shape-and-protocol-families.md
+++ b/docs/adr/ADR-0009-audit-proof-shape-and-protocol-families.md
@@ -111,12 +111,31 @@ The work budget carries debt, so an expensive proof is admitted proportionally
 less often than a cheap one. Its refill and burst sizes are set above measured
 honest demand at the maximum supported commitment size.
 
+Round two may open only the leaves round one proved, not any key in the pinned
+commitment. Membership of the commitment is the weaker property: it would let a
+cheap round one over small records authorise openings against large records
+elsewhere, so round two's cost would not be bounded by the work the caller had
+already paid for. The authorised set is recomputed from the pinned tree and the
+audit nonce, both of which the session already fixes, so it costs one tree walk
+with no chunk reads. A challenge naming any key outside the proved subtree is
+refused whole, before any chunk is read. An honest auditor cannot reach that
+case: it samples only the leaves round one returned, and it holds no committed
+block-tree root or content length to verify an answer for anything else against.
+
 ### Wire and decode bounds
 
 Subtree family messages take a tighter wire ceiling than core replication. The
 limit is selected from the encoded body discriminant before deserializing
 attacker-controlled collections and is sized above the largest legitimate
 round-one proof at the commitment key-count cap.
+
+One selector serves both directions. Encoding measures the body it produced
+against the same family ceiling the decoder applies, so an audit body that
+outgrew that ceiling fails at the sender rather than being dropped, before
+decode and without diagnosis, by every peer that receives it. A discriminant no
+declared variant uses classifies as no family at all, and no family takes the
+strict ceiling on both paths, matching how a prefix too malformed to parse is
+already treated. Failing to classify never means decoding generously.
 
 Core messages retain the existing core ceiling. In particular,
 `AuditChallenge` and `AuditResponse` remain core messages and are not
@@ -175,7 +194,12 @@ reclassified by this ADR.
 - Routing tests assert that subtree bodies are accepted only on the subtree id
   while core and digest-audit bodies remain on the core id.
 - Size tests prove the largest legitimate subtree response fits its ceiling and
-  oversized subtree collections are rejected before allocation.
+  oversized subtree collections are rejected before allocation, on the encode
+  path as well as the decode path.
+- Live-responder tests drive one round one and then assert both halves of the
+  round-two scope rule from its result: a proved leaf is still served, and a
+  committed key outside the proved subtree is refused whether or not its bytes
+  are on disk.
 - Re-validation is required for changes to subtree key derivation, sampling,
   round-one selection, session rules, or the subtree protocol family.
 
--- a/src/replication/commitment_state.rs
+++ b/src/replication/commitment_state.rs
@@ -592,15 +592,22 @@ impl ResponderCommitmentState {
     /// could still pin a recently gossiped root and open this key's blocks in a
     /// round-2 slice challenge.
     ///
-    /// This is the SAME predicate the round-2 responder uses to decide a key is
-    /// "committed" (`handle_subtree_slice_challenge` calls `built.proof_for(key)`
-    /// on the pinned slot, which is committed iff `contains_key`), folded over
-    /// every retained slot. The pruner consults it before deleting an
-    /// out-of-range key, so "the pruner will not delete it" and "the responder
-    /// still owes an answer for it" are provably the same boolean and cannot
-    /// drift. `slots` holds at most `RETAINED_GOSSIPED_COMMITMENTS` + 1
-    /// commitments, and `contains_key` is an allocation-free binary search, so
-    /// this is a short, allocation-free read.
+    /// This is `contains_key` on the pinned slot, folded over every retained
+    /// slot, and it is a strict UPPER bound on what the round-2 responder can be
+    /// asked to open. `handle_subtree_slice_challenge` authorises an opening only
+    /// if its key is in the subtree round 1 actually proved, which is one
+    /// nonce-selected block of the pinned tree rather than all of it. The bound
+    /// holds in the direction the pruner needs: every key the responder may owe
+    /// an answer for is committed under some retained slot, so a key this returns
+    /// `false` for can never be opened. The pruner consults it before deleting an
+    /// out-of-range key, so it keeps at least what is still answerable, and over
+    /// a given nonce it keeps more. Narrowing this predicate toward the
+    /// responder's would be unsound: the nonce is not known when the pruner runs,
+    /// so every committed key is reachable by some future challenge.
+    ///
+    /// `slots` holds at most `RETAINED_GOSSIPED_COMMITMENTS` + 1 commitments, and
+    /// `contains_key` is an allocation-free binary search, so this is a short,
+    /// allocation-free read.
     #[must_use]
     pub fn is_held(&self, key: &XorName) -> bool {
         self.inner.read().slots.iter().any(|c| c.contains_key(key))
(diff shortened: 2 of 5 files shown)


dirvine (approved):
## Hermes review — scope round-2 openings + wire ceiling on encode

Head `339c13e` (3 commits on top of `main`). **Verdict: approve.**

Both follow-up findings from #181 are addressed and I've independently verified the security invariants rather than taking the diff on trust.

**Round-2 scope check (a3578ff)** — The new refusal is unreachable by an honest auditor, and I traced the path:
- `admit_slice_challenge` → `SubtreeRound1Limiter::consume_session` matches `(source, challenge_id, expected_commitment_hash, nonce)` byte-for-byte against the single-use round-1 session before the handler runs. Any miss (no session, expired, restart, rotated-out pin) lands on `NoSession`/`Transient`/`UnknownCommitment` — the graced lanes — not on the new `RejectKind::Protocol`.
- Because the nonce and tree are both session-pinned, the recomputed `subtree_plan(tree, nonce)` reproduces exactly the leaf set round 1 proved, so the `authorised` set is precisely what an honest auditor samples. It cannot name a key outside it.
- The `Protocol` refusal only fires when an auditor with a live session deliberately opens outside the proved subtree — a case only a malicious auditor constructs, and it scores the responder itself, so it adds no false-positive vector against honest nodes.
- Recompute cost is one allocation-free-of-chunk-reads tree walk, gated behind the global/per-peer audit-responder admission semaphore; `key_order` is already capped at `BYTE_SPOTCHECK_MAX` before the check.

**Wire ceiling on encode (c9f2870)** — `encode` and `decode` now share one `ceiling_for(family_of_variant(...))` selector from the same table, so a body that would be dropped pre-decode by every peer now fails at the sender. The `family_of_variant` `Option` change is well-contained: every declared variant classifies to a family, `None` (undeclared index) maps to the strict 512 KiB ceiling on both paths, and all 8 call sites are consistent (protocol.rs:69 encode, :109 decode). Core bodies keep the 10 MiB allowance — no regression there. No legitimate audit body approaches the ceiling (round-1 proof ~109 KiB at the key-count cap, slice response ~22 KiB at `MAX_SLICE_OPENINGS=10`).

**ADR-0009** amendment matches the code (round-two scope rule, shared one-selector ceiling, both halves of the scope test), and it stays `Proposed`, which is the correct governance surface for a still-proposed ADR.

**Verified locally** (detached worktree at head):
- `cargo test --lib` — protocol 45/45 (incl. `the_encoder_refuses_an_audit_body_over_the_audit_ceiling`, `an_undeclared_variant_index_classifies_as_nothing`, `max_round1_proof_fits_the_audit_family_ceiling`); storage_commitment_audit 23/23; subtree 31/31; commitment_state 24/24
- `cargo test --test poc_audit_handler_live --features test-utils` — 16/16, incl. `slice_challenge_outside_the_audited_subtree_is_refused` and `an_out_of_subtree_key_is_refused_without_consulting_storage`
- `gh pr checks` — all green on head: Build (macos/ubuntu/windows), Clippy, Documentation, Format, Security Audit, Test (×3 + no-logging), ADR validation, linear-link, pr-template, self-test

The deliberately-excluded third finding (responder-wide rate bound) is correctly deferred — the PR body's analysis that a global limit without an admission/eligibility rule makes suppression cheaper is sound, and tracking it on V2-869 is the right call.

Note: I reviewed as a single independent pass with full local verification rather than a multi-agent panel; the PR had already been through independent reviewer rounds. Review posted under the `dirvine` GitHub identity this environment is authed as.

PR #202, opened 2026-08-12
fix(replication): stop conscripting peers for local write failures

grumbach:
## Linear issue

V2-963 — https://linear.app/autonominetwork/issue/V2-963/ant-node-local-possession-and-capacity-pre-check-on-the-replication

Sibling to V2-411 (disk-space pre-check in the PUT handler), which explicitly
scoped the replication path out: *"keep the existing check inside
`try_put`/`put` as defence-in-depth (covers … replication-path writes via
`replication/mod.rs`)"*. This closes that path.

## Risk tier

Node-side replication behaviour changes: a node that already holds a key, or
cannot write, stops dialling for it. No protocol, format, or economics change.

## Compatibility

- Wire: none. `FetchResult` is a private internal enum; no message, field, or
  protocol id changes. Mixed-version fleets interoperate unchanged — an
  upgraded node simply sends fewer fetch requests.
- Storage: none. No schema, key, or on-disk format change.
- API: none. No public API change (`check_capacity` is already `pub(crate)`).

## Semver impact

- [ ] breaking
- [ ] feature
- [x] fix

## Test evidence

**Unit (`cargo test --lib --features test-utils`, 924 passed)** — three new
tests drive the real pending → promotion → dequeue → in-flight path with a live
retry-slot reservation:

- `local_write_failure_does_not_conscript_the_remaining_sources` — with three
  verified sources, `LocalWriteFailed` goes straight back to verification
  instead of walking to sources two and three.
- `local_write_failure_without_retry_metadata_is_terminal` — the no-metadata
  path stays terminal, so bootstrap-drain accounting is not stranded.
- `already_held_key_leaves_the_pipeline_terminally` — `AlreadyHeld` releases the
  retry-slot reservation exactly as a completed fetch does.

**E2E (`cargo test --test e2e --features test-utils -- --test-threads=1`,
96 passed / 0 failed / 3 pre-existing ignores)** — new file
`tests/e2e/fetch_local_write_guard.rs`, observed through the holder's
`chunks_retrieved` counter:

- `write_blocked_node_does_not_dial` — target given an impossible
  `disk_reserve`, precondition established through the public `put` path:
  **holder served 0 chunks**.
- `already_held_key_is_not_fetched_again` — **held_serves=0, control_serves=1**;
  the control proves the seam and worker are live, so the zero is not vacuous.

**Both new tests were confirmed to fail without the change.** With the two
pre-dial gates removed from `execute_single_fetch`, the same run gives
`held_serves=1` and `served_by_holder=1`.

**Local workload measurement** — 40-node testnet (grown to 43 by churn), real
QUIC, real on-chain payment against Anvil, 25 seed + 60 measured paid uploads,
continuous churn, 3 seeds per arm. With 10 of 40 nodes capacity-blocked:

| | baseline | with this change |
|---|---:|---:|
| fetch serves | 4214.0 ± 295.3 | 82.7 ± 14.0 |
| fetch-lane bytes | 1053.5 ± 73.8 MiB | 20.7 ± 3.5 MiB |
| repeat delivery (share of bytes) | 95.7% | 0% |
| **total replication egress** | **1003.17 MB/run** | **60.94 MB/run** (−93.9%) |
| verification lane | 2.46 MB | 2.17 MB (does not grow) |
| holders per key | 6.92 ± 0.37 | 6.84 ± 0.13 |

The verification lane falling rather than growing is the check that this removes
bytes instead of displacing them. Possession gains per run are unchanged
(577 vs 572), so the same data still lands.

Without capacity pressure the fetch lane is already tight — every acquisition
costs exactly one delivery across steady, pruning and restart-storm arms — so
this change is inert on a healthy node.

**Known coverage gaps**, recorded in the test module rather than papered over:

1. Ordering (possession before capacity) is not pinned by a test. Proving it
   needs a node that both holds a key and refuses writes, and `disk_reserve` is
   fixed at construction, so such a node cannot be seeded. It rests on
   inspection and the `AlreadyHeld` doc comment.
2. Nothing forces `put` to fail *after* the round trip, so reverting that one
   arm to `SourceFailed` would still pass. The pre-check fires first in every
   local failure mode this harness can construct.

Closing either needs test-only hooks in `LmdbStorage`, which felt like more
surface than the change warrants.

**Review**: `codex exec --model gpt-5.6-sol` at `xhigh` — *"Safe to ship from a
runtime correctness, security, and compatibility perspective. I found no
blocker, reservation leak, bootstrap-accounting bug, or peer-scoring
regression."* Its doc-accuracy findings are all applied in this branch.

`cargo fmt --all` and `cargo clippy --all-features --all-targets` clean; PoC
suites (19 / 16 / 3) and `--no-default-features` (924) green.

## New dependency

none

## ADR

No new decision — this applies decisions already recorded:

- [**ADR-0003** — full-node detection, penalisation, and eviction](https://github.com/WithAutonomi/ant-node/blob/main/docs/adr/ADR-0003-full-node-detection-and-eviction.md) frames how a
  node that cannot accept writes should behave toward its close group. This is
  that principle applied one layer down: a full node should not consume the
  group's upstream either.
- [**ADR-0005** — replication repair hardening under churn, load, and shutdown](https://github.com/WithAutonomi/ant-node/blob/main/docs/adr/ADR-0005-replication-repair-hardening.md)
  covers the verification-and-fetch pipeline this changes, and its recurring
  shape — "a decision made once against a snapshot, then acted on much later
  against a world that had moved" — is exactly the promotion-then-dial window
  the possession recheck closes.

Both are `Status: Proposed`. Happy to write a dedicated ADR if a reviewer reads
this as a new decision rather than an application of those two.

## Mitigation / rollback

Revert the commit. The change is two added `FetchResult` variants, two local
checks in one function, and one match arm; it touches no persisted
state, no wire format, and no peer scoring, so a revert restores previous
behaviour immediately with no migration. Blast radius is bounded to the
replication fetch path: worst case on a mistaken capacity reading is that a node
defers a fetch by one verification cycle (15 s) instead of performing it, and
the existing check inside `LmdbStorage::put` still backstops the write.

### DIFF
--- a/src/replication/mod.rs
+++ b/src/replication/mod.rs
@@ -8244,6 +8244,38 @@ enum FetchResult {
     IntegrityFailed,
     /// Source failed (network error or non-success response) — retryable.
     SourceFailed,
+    /// This node could not write the chunk locally: the capacity pre-check
+    /// refused before the dial, or `put` failed after the bytes arrived.
+    ///
+    /// Distinct from [`Self::SourceFailed`] because nothing about the source
+    /// is implicated, and because the remedy is the opposite one. A source
+    /// failure should move on to the next holder; a local write failure must
+    /// not — every remaining source would send the same chunk into the same
+    /// full disk. The key returns to verification instead, so a later cycle
+    /// retries it once local capacity allows. That is a retry path, not a
+    /// liveness guarantee: `evict_stale` still expires the entry at
+    /// `PENDING_VERIFY_MAX_AGE` measured from when it was first queued, after
+    /// which re-acquisition depends on a fresh hint.
+    LocalWriteFailed,
+    /// This node already holds the key, so there is nothing to fetch —
+    /// terminal, exactly like [`Self::Stored`], because the duty is already
+    /// discharged.
+    ///
+    /// Responsibility is rechecked at download time; possession has to be too.
+    /// The earlier gates already drop held keys out of the fetch pipeline —
+    /// `queue_admitted_hints` skips them, and the promotion-time
+    /// `fetch_allowed_keys` filter and the local-paid probe both terminate
+    /// them. (`is_relevant` is not one of them: possession there makes a key
+    /// *relevant*, which is what keeps us tracking it.) So a held key reaches
+    /// the dial only when it arrived *after* promotion — a fresh-offer push, a
+    /// client PUT, or another source landing first. The nearest-first fetch
+    /// queue is deep enough for that window to be real.
+    ///
+    /// This check must also precede the capacity pre-check below, because
+    /// `LmdbStorage::put` tests `exists` *before* it tests disk space: without
+    /// it, a full node would decline a key it already holds, which `put` would
+    /// have accepted as a duplicate.
+    AlreadyHeld,
     /// Live routing state no longer places this node in the storage-admission
     /// group for the key — terminal, exactly like [`Self::Stored`]. The duty
     /// the fetch was serving has lapsed, so no alternate source is tried and
@@ -8291,7 +8323,7 @@ fn apply_fetch_result(
     verification_retry_after: Duration,
 ) -> FetchFollowUp {
     match result {
-        FetchResult::Stored | FetchResult::NoLongerResponsible => {
+        FetchResult::Stored | FetchResult::NoLongerResponsible | FetchResult::AlreadyHeld => {
             q.complete_fetch(key);
             FetchFollowUp::Terminal
         }
@@ -8299,6 +8331,19 @@ fn apply_fetch_result(
             if let Some(next_peer) = q.retry_fetch(key) {
                 FetchFollowUp::RetryFrom(next_peer)
             } else if q.requeue_fetch_for_verification(key, verification_retry_after) {
+                FetchFollowUp::RequeuedForVerification
+            } else {
+                FetchFollowUp::Terminal
+            }
+        }
+        // No `retry_fetch`: the local write is what failed, so every other
+        // source would be asked to send the same chunk into the same full
+        // disk. Requeueing for verification is what keeps the key from being
+        // stranded — it comes back once capacity allows — and the fallthrough
+        // to `Terminal` when there is no retry metadata is what keeps
+        // bootstrap drain accounting correct, exactly as for a source failure.
+        FetchResult::LocalWriteFailed => {
+            if q.requeue_fetch_for_verification(key, verification_retry_after) {
                 FetchFollowUp::RequeuedForVerification
             } else {
                 FetchFollowUp::Terminal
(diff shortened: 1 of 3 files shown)


dirvine (approved):
**Verdict: Approve — no blockers found.**

Reviewed exact head `053fc69` against `main`. I read the full replication-path diff (`src/replication/mod.rs`, `src/replication/scheduling.rs`, the new unit tests, and `tests/e2e/fetch_local_write_guard.rs`), ran the suite locally, and checked CI on this head.

**The fix is sound.** `execute_single_fetch` now gates in the right order — responsibility → local possession (`storage.exists`) → local capacity (`check_capacity`) → dial — so a full node that already holds a key, or cannot write, never spends a holder's bandwidth. The reclassification of a post-round-trip `put` failure from `SourceFailed` to `LocalWriteFailed` is the key change: instead of `retry_fetch` walking to every remaining holder (each re-uploading the same chunk into the same full disk), the key goes straight back to verification and comes back once capacity allows. The ordering (possession before capacity) correctly mirrors `LmdbStorage::put`, which tests `exists` before disk space, so a full node still accepts a key it already holds.

**Semantics and accounting are correct.** `AlreadyHeld` shares the terminal path with `Stored`, releasing the retry-slot reservation and shrinking the bootstrap pending set; `LocalWriteFailed` requeues for verification, and the no-retry-metadata fallthrough to `Terminal` keeps bootstrap-drain accounting from stranding (both pinned by unit tests). I confirmed the only remaining `SourceFailed` returns in `execute_single_fetch` are genuine source/network/shutdown paths — every `put`-failure arm was updated.

**Evidence I verified directly:**
- `cargo test --lib --features test-utils`: **924 passed / 0 failed**.
- Both new e2e tests pass locally (`write_blocked_node_does_not_dial` served=0 with precondition proven via public `put`; `already_held_key_is_not_fetched_again` held_serves=0, control_serves=1 — a genuine positive control).
- CI on this head is green: Build/Test/Clippy/Format/Doc/Security all success. The one `pr-template` failure is a stale re-run; the latest run passes after the `## ADR` section (ADR-0003 + the existing capacity ADR) was added.
- The measurement claim (93.9% less egress) is backed by the 40-node local run, and "holders per key 6.92→6.84" indicates replication coverage is unchanged — it removes wasted bytes rather than displacing them.

**Non-blocking observations:**
1. `storage.exists(&key).unwrap_or(false)` swallows a storage-read error as "not held" — graceful, but a transient LMDB read-txn failure would cause one redundant dial that `put` then dedupes. Fine as-is; just noting the asymmetry with `check_capacity`, which surfaces its error.
2. The e2e tests observe egress via the holder's `chunks_retrieved` counter on a freshly built testnet. There's a small inherent flakiness window if unrelated traffic bumps that counter mid-observation; serial execution and a 15 s window mitigate it, but the `served == 0` assertion may occasionally need a re-run on a loaded CI box.

**Process note (honesty):** per the usual workflow I attempted two independent 3-seat review panels. Every subagent was blocked by the sandbox's 256-descriptor soft limit (`Too many open files`) and timed out, so this review reflects my own direct code inspection plus the test runs above — not an independent panel consensus. I found no blocking issues by either route.

Nice, tight change — the doc comments on the ordering and the terminal/requeue semantics are genuinely helpful.

PR #204, opened 2026-08-13
feat(quote): refuse quotes to clients that cannot settle correctly

grumbach:
## Linear issue

Closes V2-975 — https://linear.app/autonominetwork/issue/V2-975/refuse-to-quote-clients-that-cannot-settle-correctly-instead-of

## Risk tier

New wire handling on the payment-admission path.

## What this fixes

Production nodes reject a steady trickle of merkle uploads for underpayment at an exact 3x ratio. The node is right and the client is wrong: those clients settle under the pre-ADR-0008 rule and apply no multiplier.

The problem is *when* we refuse. A merkle batch pays on-chain before any storer sees a PUT, so checking the settlement rule at PUT time checks it after the money is gone, and merkle receipts are not refundable. ADR-0008 already names this as a re-open trigger: "a rise in refused batch uploads after the boundary, indicating clients that never upgraded." That trigger has fired.

## Change

Refuse at quote time instead. No quote means no pool commitment, which means no payment, so a refused client has spent nothing.

- A client **below** `MIN_SUPPORTED_SETTLEMENT_VERSION` gets `ClientUpdateRequired` with an upgrade instruction.
- A client **above** `CURRENT_SETTLEMENT_VERSION` gets `StorerUpdateRequired`. This node is the old one; the client should use a different storer and tell its user nothing.
- **Unversioned requests are still served.** A node cannot distinguish a client that settles correctly but predates the version field (ant-core 0.5.1 through 0.6.0) from one that does not, so refusing both would break clients that are behaving. A running count is logged every 1000 per path under `ant_node::quote::settlement`, which is the evidence for flipping that later.
- **The underpayment message now carries upgrade advice**, keyed on an exact multiplier shortfall. This is the only signal that reaches clients losing money today, since a client too old to settle correctly is also too old to declare a version.

## Changes since the last review

- **Merged current `main`** (22 commits, through the unresolved-verification backoff merge). Git reported no conflicts; `main`'s only edit to a file this branch also touches is a comment and test rename in `src/storage/handler.rs`, and it does not move the quote path. Lint, docs and the library suite are clean on the merged tree.
- **Renumbered the ADR to 0013.** `main` took ADR-0010 for the beta upgrade channel decision while this branch was open and has since reached ADR-0012, so the repository's ADR governance check was failing this branch on a duplicate number. Number only; the content is unchanged. Three other open pull requests (#197, #213, #215/#216) still carry numbers `main` has taken, so whoever renumbers next should start at 0014.
- **Corrected two stale references to `ant-protocol#23`**, which was reverted in `ant-protocol#24`. The live protocol pull request is `ant-protocol#25`.
- **The unversioned-quote counter now says what it counts.** It increments when the request arrives, before validation and before a quote is generated, so a refused request is counted too, but the log line reported a running total of quotes *served*. A node turning away a thousand oversized unversioned requests would have reported serving a thousand quotes, and that number is the stated input to the decision to retire the legacy path. Arrival is the right place to count, since a client whose request failed for an unrelated reason still cannot declare a version and would still break if the path closed, so the counter and the log line are reworded to match where the increment fires rather than the increment being moved. No behavioural change.
- **Rebased onto `main`** (through the chunk GET telemetry merge). The one conflict was two additive blocks landing at the same point in `src/storage/handler.rs`; both are kept. 932 lib tests pass, `cfd` clean.
- **The gate now runs over a real connection.** New e2e driver, details under Test evidence. This closes the half of the mixed-version gate that did not need the coordinated set to land, and it is the first observation of the `ant_node::quote::settlement` log target on a running node.
- **Took the `h2` fix for RUSTSEC-2026-0258.** The advisory published on 2026-08-17, after main's last CI run, so Security Audit started failing here without anything in the branch causing it; main is affected identically. Lockfile only, and the `h2` entry alone — a plain `cargo update` re-resolves socket2 and windows-sys across quinn and hyper-util, which is QUIC-stack churn a settlement change should not carry. `cargo check --locked` accepts the result.
- **ADR-0013 gained a rollout order and a revert order**, with the signal that says each step worked. Both rows stay marked not rehearsed.

## Changes since first review

- **Bounded the range at both ends.** The first revision served any version at or above the minimum, on the reasoning that the storer verifies whatever arrives anyway. That holds only for settlement changes that *raise* what is paid, which ADR-0008 happened to be; a change to the median rule or the payable field breaks it, after the client has already settled. Corrected, and the test that pinned the old behaviour is inverted rather than deleted.
- **Split the refusal in two**, so a lagging node never reports a client fault. Collapsing them would tell up-to-date users to upgrade during exactly the window when most of the fleet is the old side.
- **Added ADR-0013** (below), including a correction: it had claimed the client's downgrade path was build-enforced, which held for one of the two fallbacks only.

## Compatibility

- Wire: additive, no `CHUNK_PROTOCOL_ID` bump. Every existing message shape is handled as before. Deployment ordering matters in one direction: **nodes before clients**, since a node on the current published `ant-protocol` cannot decode versioned requests. The client PR covers the mixed fleet with a per-peer fallback, so this is a preference rather than a hard gate.
- Storage: none.
- API: none. Both new handler methods are private.

## Semver impact

- [ ] breaking
- [x] feature
- [ ] fix

No version bump is taken in this PR; the release train owns that.

## Test evidence

`cargo test --lib` — **932 passed, 0 failed**. `cfd` (fmt + clippy with `-D clippy::panic -D clippy::unwrap_used -D clippy::expect_used -D warnings`) clean.

Gate behaviour (`storage::handler`):
- `v2_merkle_quote_is_served_at_the_current_settlement_version` — the gate is invisible to clients that can pay.
- `v2_merkle_quote_is_refused_below_the_minimum_settlement_version` — asserts the refusal type, both versions, and that the rendered message contains `ant update`.
- `v2_single_node_quote_is_refused_below_the_minimum_settlement_version` — so a refused merkle client cannot fall back to per-chunk quotes and burn money that way instead.
- `a_newer_settlement_version_is_refused_as_this_nodes_fault` — the corrected upper bound, asserting `StorerUpdateRequired` specifically.
- `unversioned_requests_are_still_served` — the decision that would break correctly-paying clients if it regressed.

Message behaviour (`payment::verifier`):
- `merkle_legacy_1x_settlement_rejected_after_the_parity_boundary` extended to assert the upgrade advice on an exact 1x settlement.
- `a_partial_shortfall_is_not_blamed_on_an_outdated_client` — a one-wei shortfall rejects without upgrade advice, so the advice keeps meaning something.

### The gate over a real connection

`settlement_version_gate_observed_on_live_network` (`tests/e2e/settlement_version_gate.rs`) spawns a devnet from **this branch** and puts all four cases on real QUIC between two nodes, on both quote paths:

| Request | Result |
|---|---|
| Unversioned — the shape every released client sends | Served |
| Declaring `CURRENT` | Served |
| Declaring below `MIN` | `ClientUpdateRequired`, no quote issued |
| Declaring above `CURRENT` | `StorerUpdateRequired`, no quote issued |

That promotes **old client against an upgraded node** and the **structured refusal against a real peer**, both previously unit-level only, to observed. Run with `RUST_LOG=ant_node::quote::settlement=warn` it also prints the refusal lines as a running node emits them:

```text
WARN ant_node::quote::settlement: Refusing single_node quote: client settlement version 0
  is below the minimum 1. No quote issued, so the client has not been charged.
WARN ant_node::quote::settlement: Refusing merkle quote: client settles under version 2,
  newer than this node's 1. This node needs upgrading; the client has not been charged.
```

### Mixed-version validation

No longer outstanding for the direction that could be tested. `ant-client`'s merkle E2E spawns a 35-node testnet from the **published** `ant-node`, so it is a live new-client-against-old-fleet run: every node logs a decode failure for each versioned probe and answers only the unversioned retry.

It passed functionally and **failed on cost**, which is what makes it worth having run: the suite went from a 24–38 minute baseline on `main` to exceeding the 60-minute CI cap with 4 of 7 tests done. A peer that cannot decode never answers, so the client waited a full quote timeout before falling back, on every request rather than once per peer. Two bounds on the client PR fix it. Full write-up in ADR-0013.

**Still outstanding:** the reverse direction, old client against an upgraded node, and a genuinely mixed fleet. Both need a testnet built from this branch's `ant-node`, which does not exist until the coordinated set lands.

### CI

**All 15 checks green** on the current head, `Test (windows-latest)` included.

That job failed on an earlier head, in testnet **setup** ("Failed to create dual-stack network nodes") rather than on an assertion, and the same job fails on `main` in both of its recent failing runs. It passes here on re-run, which settles it as the known hosted-runner flake rather than anything in this change.

## New dependency

None added. One transitive lockfile bump, `h2` 0.4.15 to 0.4.16, for the advisory above. `ant-protocol` is temporarily repointed at the review branch for WithAutonomi/ant-protocol#23, and reverts to a published version pin once that merges and the release train publishes it. **That repin is a merge-order dependency, not something this PR can take**: publishing is the train's step, and per repo policy a feature PR does not bump or name crate versions.

## ADR

**ADR-0013: Settlement version and pre-payment compatibility** (Proposed) — added on this branch in response to review.

https://github.com/grumbach/ant-node/blob/settlement-version-quote-gate/docs/adr/ADR-0013-settlement-version-and-pre-payment-compatibility.md

Records the inclusive `MIN..=CURRENT` range and why the upper bound is load-bearing, the two refusal directions and why they must stay distinct during a client-first rollout, the unversioned retry as a bounded downgrade path with a shared compile-time cutover guard, the rule that a refusal must not depend on which peers answered first, and the residual that merkle storers are not exactly the peers a client quotes.

Builds on ADR-0008, whose re-open trigger is what fired here.

## Release readiness: NOT production ready

Merging puts this in the next release, so the bar is fleet-ready rather than code-complete. CI being green is evidence for the code gate only. **Do not merge while any row below is open.**

| Gate | Status | What closes it |
|---|---|---|
| Code / CI | **Proven.** 15/15 green, 931 lib tests, three Codex xhigh rounds adjudicated | — |
| Dependency | **Open**, and the only code-level blocker left. `Cargo.toml` pins a mutable personal-fork branch | ant-protocol #23 merged, then repin to upstream. The registry repin waits on the train, since publishing needs a tag and a version bump a feature PR cannot take |
| Mixed-version | **Partial.** New-client/old-fleet proven by the client suite. Old-client/new-node, new-client/new-node and both refusal directions now run over a real connection on a devnet built from this branch | A fleet running both node versions at once, which needs a deployment rather than a devnet |
| Deployment ordering | **Planned, not rehearsed.** ADR-0013 now carries the order, the three steps and the signal that says each one worked | Node deployment preceding the client release, observed |
| Observability | **Open**, narrowed. The refusal lines are now seen emitting from a running node. The unversioned-quote counter is the signal that retires the legacy path, and at one line per 1000 requests it cannot fire below production volumes, let alone have been read there | Confirm that line appears and is queryable on a production node |
| NAT / canary | **Open.** No canary. Relayed and NAT'd peers are the slow paths this change adds work to | A canary run with the counter and quote latency observed |
| Rollback | **Planned, not rehearsed.** ADR-0013 states the revert order — clients first, then nodes, deliberately not the reverse of the rollout — and what it costs | Rehearsing it |
| Fleet safety | **Open.** The corroboration quorum and client-wide latch have never met a real fleet | Canary evidence that no honest upload is halted |

### What is genuinely low risk today

The gate is **inert on arrival**. `MIN_SUPPORTED_SETTLEMENT_VERSION` and `CURRENT_SETTLEMENT_VERSION` are both the first declarable version, so no client in existence can be refused for being too old and no node can be refused for being behind. The refusal machinery, the corroboration quorum and the latch cannot fire until a future settlement bump.

That is a reason the *risk* is low. It is not evidence the gates are closed, and it does not make the set fleet-ready.

## Mitigation / rollback

Revert. `MIN_SUPPORTED_SETTLEMENT_VERSION` equals the first version any client can declare, so no client in existence can trip the lower gate, and no node has a lower `CURRENT`, so none can trip the upper one. The change is inert on arrival and becomes load-bearing at the next settlement bump. The underpayment message change is text only.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -862,8 +862,8 @@ dependencies = [
 
 [[package]]
 name = "ant-protocol"
-version = "2.3.2"
-source = "git+https://github.com/grumbach/ant-protocol?branch=reapply/pr-23-settlement-version#f2bf6e42547ef06bf6ab46acbc4941a5d3a1014d"
+version = "2.3.3"
+source = "git+https://github.com/grumbach/ant-protocol?branch=reapply/pr-23-settlement-version#a22897d7a3048c41f8c419975274151c9ac9a843"
 dependencies = [
  "blake3",
  "bytes",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -862,8 +862,8 @@ dependencies = [
 
 [[package]]
 name = "ant-protocol"
-version = "2.4.0"
-source = "git+https://github.com/grumbach/ant-protocol?branch=settlement-version-quote-gate#439ed83d70ab3bbd14c1031c2e319f793141ea5a"
+version = "2.3.2"
+source = "git+https://github.com/grumbach/ant-protocol?branch=settlement-version-quote-gate#83160b044bdd3f035b9a78ba214d9510ab3ca753"
 dependencies = [
  "blake3",
  "bytes",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -863,7 +863,7 @@ dependencies = [
 [[package]]
 name = "ant-protocol"
 version = "2.4.0"
-source = "git+https://github.com/grumbach/ant-protocol?branch=settlement-version-quote-gate#19845c8c20e1a3505cfbfc446b7e2c26bf5b0726"
+source = "git+https://github.com/grumbach/ant-protocol?branch=settlement-version-quote-gate#620bb9971ba5669bdb23a38568f8d3a597987db1"
 dependencies = [
  "blake3",
  "bytes",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -862,9 +862,8 @@ dependencies = [
 
 [[package]]
 name = "ant-protocol"
-version = "2.3.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "20a8d4262f20934f579f1f37e7855718d75156e24b3441eabf089e3049d4c831"
+version = "2.4.0"
+source = "git+https://github.com/grumbach/ant-protocol?branch=settlement-version-quote-gate#19845c8c20e1a3505cfbfc446b7e2c26bf5b0726"
 dependencies = [
  "blake3",
  "bytes",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -863,7 +863,7 @@ dependencies = [
 [[package]]
 name = "ant-protocol"
 version = "2.3.2"
-source = "git+https://github.com/grumbach/ant-protocol?branch=settlement-version-quote-gate#83160b044bdd3f035b9a78ba214d9510ab3ca753"
+source = "git+https://github.com/grumbach/ant-protocol?branch=settlement-version-quote-gate#e97901694cba5cad84da0b99ee8af02a791df2cf"
 dependencies = [
  "blake3",
  "bytes",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -863,7 +863,7 @@ dependencies = [
 [[package]]
 name = "ant-protocol"
 version = "2.3.2"
-source = "git+https://github.com/grumbach/ant-protocol?branch=settlement-version-quote-gate#e97901694cba5cad84da0b99ee8af02a791df2cf"
+source = "git+https://github.com/grumbach/ant-protocol?branch=reapply/pr-23-settlement-version#f2bf6e42547ef06bf6ab46acbc4941a5d3a1014d"
 dependencies = [
  "blake3",
  "bytes",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -863,7 +863,7 @@ dependencies = [
 [[package]]
 name = "ant-protocol"
 version = "2.4.0"
-source = "git+https://github.com/grumbach/ant-protocol?branch=settlement-version-quote-gate#620bb9971ba5669bdb23a38568f8d3a597987db1"
+source = "git+https://github.com/grumbach/ant-protocol?branch=settlement-version-quote-gate#439ed83d70ab3bbd14c1031c2e319f793141ea5a"
 dependencies = [
  "blake3",
  "bytes",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ mimalloc = "0.1"
 # Branch pin while the settlement-version wire types are in review
 # (WithAutonomi/ant-protocol#25). Swap back to a published `ant-protocol`
 # version pin once that PR merges and the release train publishes it.
-ant-protocol = { git = "https://github.com/grumbach/ant-protocol", branch = "settlement-version-quote-gate" }
+ant-protocol = { git = "https://github.com/grumbach/ant-protocol", branch = "reapply/pr-23-settlement-version" }
 
 # Core (provides EVERYTHING: networking, DHT, security, trust, storage)
 saorsa-core = "0.27.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,8 +40,8 @@ mimalloc = "0.1"
 # (the rc-2026.4.2 branch) so Cargo can unify the wire types here
 # with ant-protocol's re-exports.
 # Branch pin while the settlement-version wire types are in review
-# (WithAutonomi/ant-protocol#23). Swap back to `ant-protocol = "2.4.0"` once
-# that PR merges and 2.4.0 is published.
+# (WithAutonomi/ant-protocol#23). Swap back to a published `ant-protocol`
+# version pin once that PR merges and the release train publishes it.
 ant-protocol = { git = "https://github.com/grumbach/ant-protocol", branch = "settlement-version-quote-gate" }
 
 # Core (provides EVERYTHING: networking, DHT, security, trust, storage)
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ mimalloc = "0.1"
 # (the rc-2026.4.2 branch) so Cargo can unify the wire types here
 # with ant-protocol's re-exports.
 # Branch pin while the settlement-version wire types are in review
-# (WithAutonomi/ant-protocol#23). Swap back to a published `ant-protocol`
+# (WithAutonomi/ant-protocol#25). Swap back to a published `ant-protocol`
 # version pin once that PR merges and the release train publishes it.
 ant-protocol = { git = "https://github.com/grumbach/ant-protocol", branch = "settlement-version-quote-gate" }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,10 @@ mimalloc = "0.1"
 # Until then, the git pin tracks the matching saorsa-core lineage
 # (the rc-2026.4.2 branch) so Cargo can unify the wire types here
 # with ant-protocol's re-exports.
-ant-protocol = "2.3.3"
+# Branch pin while the settlement-version wire types are in review
+# (WithAutonomi/ant-protocol#23). Swap back to `ant-protocol = "2.4.0"` once
+# that PR merges and 2.4.0 is published.
+ant-protocol = { git = "https://github.com/grumbach/ant-protocol", branch = "settlement-version-quote-gate" }
 
 # Core (provides EVERYTHING: networking, DHT, security, trust, storage)
 saorsa-core = "0.27.1"
--- a/docs/adr/ADR-0010-settlement-version-and-pre-payment-compatibility.md
+++ b/docs/adr/ADR-0010-settlement-version-and-pre-payment-compatibility.md
@@ -103,7 +103,7 @@ Flipping that to a refusal is a **follow-up**, gated on that count decaying, and
 
 - **It does not fix the current burn.** A client too old to settle correctly is also too old to declare a version, so the gate cannot see the population causing today's rejections. Only the reworded error reaches them. Driving client adoption remains the cheaper and faster remedy for the incident that prompted this.
 - **Merkle storers are not exactly the quoted peers.** The gate covers the 16 candidates a client quotes, but the chunk is stored by each chunk's close group, which may include a peer that never quoted. A storer outside the candidate set can still refuse at PUT time. This narrows the exposure substantially without closing it, and only option 3 closes it fully.
-- The fallback costs one extra timeout per silent peer during rollout. Requests run concurrently, so the worst case is 2x the quote timeout overall.
+- The fallback costs one probe per silent peer, bounded by `VERSIONED_QUOTE_PROBE_CEILING` and paid once per peer rather than once per request. Before those two bounds it was one full quote timeout on **every** request, which took the merkle E2E suite from ~24 minutes past the 60-minute CI cap. See the validation section.
 - Clients declaring a version are refused by nodes that have not upgraded. Harmless today because no node has a lower `CURRENT`, but it makes node rollout a prerequisite for the next settlement bump.
 
 ### Neutral / Operational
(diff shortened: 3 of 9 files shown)


dirvine (changes requested):
Reviewed exact head `2f7a05ff6e7575e732c6b3ae09c9f3ada66beb97` with the coordinated protocol/client PRs.

The local gate implementation is straightforward and the focused settlement tests pass 11/11. I am requesting changes on the coordinated behaviour rather than the handler mechanics:

- `handle_quote_v2` / `handle_merkle_candidate_quote_v2` inherit `settlement_version_is_supported(version >= MIN)`. An older node consequently quotes to an arbitrarily newer settlement version it cannot understand. That forfeits the pre-payment compatibility guarantee; actual PUT-time verification is too late once payment has settled.
- Legacy requests remain unconditionally served (`src/storage/handler.rs:288-296`). Combined with the client PR retrying V2 after any `Network`/`Timeout`, a transient failure or dropped V2 response can downgrade an upgraded peer to the ungated legacy path. This becomes payment-burning again when `MIN_SUPPORTED_SETTLEMENT_VERSION` is next raised.

Please land the superseding ADR and specify a downgrade-resistant rollout/capability policy, then add a mixed-fleet test covering legacy node, upgraded node, structured refusal, lost refusal/timeout, and send failure before merging.

CI is not fully green: Windows tests failed because two testnets could not initialise the dual-stack transport (94 passed, 2 failed), apparently unrelated but still unresolved. The manifest also points at a contributor branch and must be changed to the published `ant-protocol = "2.4.0"` before merge.

dirvine (changes requested):
Follow-up review of exact head `26b34e35f94c08c4ff513c4ad9ff1aa06bb67861` across the coordinated protocol/node/client set.

The node-side code blocker from my previous review is resolved: `settlement_gate()` now maps `ClientTooOld` to terminal `ClientUpdateRequired` and `NodeTooOld` to skippable `StorerUpdateRequired`. Local focused settlement tests passed 11/11, and completed CI checks are green.

The coordinated merge gates are not complete yet:

- `Cargo.toml` still uses the mutable personal-fork branch pin. Merge/publish ant-protocol 2.4.0 first, then repin this PR to the registry release and regenerate the lockfile.
- ADR-0010 remains Proposed and explicitly records the mixed-version dev-testnet validation as outstanding. For this T3 pre-payment gate, that run should cover legacy/upgraded nodes, structured refusal, lost response/timeout and send failure before this PR merges.
- Three platform test jobs are still running at review time.
- The PR body still says future settlement versions are served and cites the removed `a_newer_settlement_version_is_not_treated_as_an_error` test; the implementation now deliberately refuses them. Please update the body before sign-off.

No additional handler defect found, but the stated release/validation gates remain open.

dirvine (changes requested):
Re-reviewed exact head `1a8ff66408cae9915605ef39627dafac4528644f`. The prior code findings are resolved: compatibility is bounded in both directions, ADR-0010 now describes both downgrade paths accurately, and the uneven-depth stale-client diagnostic uses the correct bare expectation. Focused local settlement tests and the new depth-7 regression pass; all 15 GitHub checks are green.\n\nOne merge blocker remains: `Cargo.toml` still uses the mutable personal-fork branch `github.com/grumbach/ant-protocol#settlement-version-quote-gate`. Please merge/release ant-protocol #23, repin this PR to the published registry version, and regenerate `Cargo.lock`. Also refresh the PR body: its Windows-failure note is stale because the current Windows unit/build jobs pass.

dirvine (changes requested):
Re-reviewed exact head `a4aa14e0c1133b677347033a7549a8daea509946`. The incremental changes are ADR-only and introduce no executable regression. The revised release-gate and probe-ceiling analysis is materially accurate. Local formatting and focused settlement tests pass 11/11; all 15 GitHub checks are green. No node-code blocker remains.

One merge blocker remains unchanged: `Cargo.toml` still resolves `ant-protocol` through the mutable personal-fork branch `grumbach/ant-protocol#settlement-version-quote-gate`. Protocol #23 is still open and the latest registry release remains 2.3.2, so a stable repin is not yet possible. Please merge/publish the protocol release, replace this with the registry version and regenerate `Cargo.lock`; I will then perform the final sign-off.

Fleet validation, observability, canary and rollback items in ADR-0010 are production-release gates rather than additional code findings on this head.

dirvine (approved):
APPROVE — reviewed exact head `58a9500b6553de927df9c15084027ee89f0efd2a`.

No material blockers found. Settlement compatibility is checked before quote/payment work, both V2 quote paths enforce the gate, and legacy handling is preserved. CI is green. Local verification: formatting and all 11 focused settlement tests passed.

Merge/release sequencing still requires ant-protocol#25 to be merged and published first. Release version bumps remain with the release owner.

jacderida:
## Testnet evidence — ADR-0013 mixed-fleet + refusal-flip run (V2-1151, 2026-09-04)

Validated at `e6222d2` (current HEAD of this PR), alongside WithAutonomi/ant-protocol#25 @ `a22897d` and WithAutonomi/ant-client#171 @ `69e8c93`.

**5.5h run, 337 nodes at T0 rising to 458.** Four cohorts sharing one set of bootstraps: 225 branch nodes (this PR) + 105 released 0.17.2 nodes + 3 uploaders/2 downloaders per client cohort, then a "future" cohort built against a protocol fork with `MIN = CURRENT = 2` started in three stages (1, then 30, then 90 nodes) to fire the refusal path. Cohorts identified by **binary sha256**, not version string — all report 0.17.2.

### The refusal emission is correctly scoped

`Refusing` lines on `ant_node::quote::settlement`, whole run:

| Cohort | `MIN` | Refusing lines |
| -- | -- | -- |
| DEV-01 (this PR's build, 225 nodes) | 1 | **0** |
| DEV-01-REL (released 0.17.2, 105 nodes) | n/a | **0** |
| FUT-0 (1 node) | 2 | 310 |
| FUT-A (30 nodes, 2 hosts) | 2 | 1260 |
| FUT-B (90 nodes, 6 hosts) | 2 | 1140 |

Zero refusals fleet-wide before the first flip, and afterwards refusals appear **only** on nodes whose `MIN` actually excludes the client. The message is right too:

```
Refusing single_node quote: client settlement version 1 is below the minimum 2.
No quote issued, so the client has not been charged.
```

### The unversioned path stays served

Released clients declare no settlement version and were served throughout by branch nodes — **227/227, 33/33, 51/51, 44/44 uploads = 100%** across Phase 1 and all three flip windows, with `refusal_seen=0` on every attempt. The adoption counter fired as designed on the branch cohort (279 docs):

```
Received 1000 single_node quote requests from clients that declare no settlement
version. These clients cannot be told to upgrade before they pay.
```

### Cost to the mixed fleet is bounded

Branch-client upload latency over the 3h Phase 1 soak, against the released cohort as control, per uploader: **+21.7s** (300 MB), **+9.8s** (500 MB), **+15.0s** (1200 MB). Trends across the window −2.4% / −3.7% / +11.9% — bounded, roughly constant, no timeout spiral, including against relayed/NAT'd released peers (~30% NAT). Downloads held at **100% in every phase** on both cohorts across joins of 1, 30 and 90 nodes.

### ⚠️ One thing to know when querying these logs

The `ant_node::quote::settlement` target also carries the non-refusing INFO adoption counter. Scoping a query to `tag.target.keyword` **alone** reported 279 "refusals" on a cohort that issued none — the message must be filtered too. This caught us on two consecutive runs.

Full criteria table, measurement method and the raw evidence: [V2-1151](https://linear.app/autonominetwork/issue/V2-1151/settlement-version-quote-gate-adr-0013-run-4-corroborated-refusal). 11 of 12 acceptance criteria passed; the twelfth (paid-then-refused) had an empty population for reasons explained on WithAutonomi/ant-client#171.

jacderida:
**Recommendation: merge.** Scheduled for Monday 2026-09-07, after ant-protocol#25; full decision record on Linear V2-975.

The testnet evidence (comments above; runs V2-1109 and V2-1151) closes the ADR-0013 release gates: the node-side gate behaved perfectly across both measured runs — zero refusal lines anywhere pre-flip on four deployments, refusals only ever on the future-cohort nodes, the unversioned path served throughout (the released control cohort ran at 100% in every window, and the adoption counter fired on both runs), and no PUT-time payment failure ever named an uploader wallet. 1,032 refused uploads spent zero atto, confirmed on-chain.

Mechanics for this repo: merge-up onto main (16 commits behind), and **before merging, repoint the `ant-protocol` git pin from the `grumbach` fork branch to the commit merged by ant-protocol#25** — main must not depend on a personal-fork branch — bumping the lock with `cargo update -p ant-protocol --precise <rev>` (a plain update keeps the stale rev for git branch pins). Rollout order at release stays as ADR-0013 states: nodes before clients.

PR #207, opened 2026-08-18
fix(replication): stop a write-blocked node probing its close group

grumbach:
## Linear issue

V2-1009 — https://linear.app/autonominetwork/issue/V2-1009/ant-node-gate-the-replication-verification-cycle-on-local-write

Follow-up to V2-963 (#202), which added the capacity pre-check to the fetch path,
and to the testnet run recorded on
[V2-987](https://linear.app/autonominetwork/issue/V2-987/testnet-run-replication-fetch-guard-with-artificially-full-nodes-v2),
which measured the cost this closes. That run's failing criterion is not a
regression from #202: it measured a trend with no baseline arm, and the local A/B
that does have both arms shows the guard slightly *below* main on the same lane.

## Risk tier

Node-side replication behaviour: once a node's available space is below its
reserve, keys it has *already authorized* stop running the source-discovery round
and stop being promoted for fetch. Unauthorized keys still run their quorum round,
paid-list convergence and the terminal checks still run, and a failed space query
or other write error is not gated. No protocol, format, or economics change.

## Compatibility

- Wire: none. The gate decides whether this node *sends* a `VerificationRequest` it
  was about to send anyway. No message, field, or protocol id changes, and no
  peer-side heuristic consumes the absence of unsolicited verification requests.
- Storage: none. `ReplicationConfig` is neither serialised nor persisted; persisted
  TOML is `NodeConfig`. No schema, key, or on-disk format change.
- API: none in a default build, and checked three times. The backoff is a `pub(crate)` constant,
  deliberately **not** a `ReplicationConfig` field — that struct is publicly
  re-exported and is not `#[non_exhaustive]`, so adding a field would break
  downstream exhaustive construction. An earlier revision added one and it was
  removed. A later review then caught that the constant and the new
  `in_flight_verification_age` accessor were still `pub` inside public modules,
  which is additive rather than breaking but is still public surface; both are
  now `pub(crate)`. The new public items — the two engine test hooks, the probe watch and its query —
  are all `#[cfg(any(test, feature = "test-utils"))]`, so they are additive under
  `test-utils` and absent from a default build. Four items, not one.

## Semver impact

- [ ] breaking
- [ ] feature
- [x] fix

## Test evidence

**Simplicity.** The production change is one capacity read per cycle, two gates
that call the existing `defer_pending`, and a three-way verdict so a failed space
query is not mistaken for a full disk. The verdict's mapping sits in its own free
function so it can be proved without depending on how a given platform answers a
space query for an unusual path.
`scheduling.rs` carries no production change; the only edit there is one test.
An earlier revision also clamped the deferral to
half the entry's remaining life, protected the deferred key from fairness
eviction, and carried the verdict through `FetchResult`; all three were cut on
review, for the reasons recorded in ADR-0011.

**Unit (`cargo test --lib`, 934 passed / 0 failed)** — the existing
`apply_fetch_result` cases still hold with `LocalWriteFailed` back to its
upstream shape, plus five new tests. Each is mutation-checked, and each states
in its own rustdoc what it does *not* show.

- `capacity_verdict_follows_the_reserve_and_notices_recovery` — the verdict
  tracks the reserve in both directions. The third assertion is the one that
  matters operationally: once space is freed the very next read has to see it,
  because a node stuck on a remembered refusal would stop fetching for good. What
  it proves precisely is that `Full` is not negatively cached: the condition is
  cleared by dropping the reserve, which is the same comparison from the other
  side, not by freeing actual space. Killed by inverting the
  comparison, by caching the refusal in the passing-result slot, and by memoising
  it separately.
- `a_failed_space_query_reads_as_unknown_not_full` — a failed `statvfs` must be
  classified as `Unknown` rather than as a full disk, plus the reserve boundary
  itself. Killed by folding the error arm into `Full`, by inverting the
  comparison, and by slipping the boundary to `<=`. This is the round-7 finding,
  pinned at the classification. That the gate then leaves `Unknown` alone is a
  separate line inside the cycle and stays inspection-only.

  The three-way mapping moved into a small free function, `verdict_from_available_space`,
  for this. The first version of the test forced a real query failure by pointing
  the storage at a directory that does not exist; that is not portable, and CI
  proved it — the test passed on macOS and Linux and failed on Windows, which
  resolves a missing subdirectory to its volume and answers successfully. Injecting
  the query result is both simpler and provable everywhere.
- `capacity_deferral_does_not_extend_entry_lifetime` — deferral moves the next
  look, not the entry's birth, so a blocked key still ages out on schedule. This
  is why no eviction protection is carried: protecting deferred keys would
  preserve, against a bounded queue, the one duty a full node cannot discharge.
  Killed both by exempting deferred keys from eviction and by refreshing
  `created_at` on deferral.
- `capacity_blocked_retry_is_an_order_above_the_request_timeout_and_below_the_entry_lifetime`
  — the constant has to be an order above the retry it replaces, or the repeat
  cost stays the same order, and below `PENDING_VERIFY_MAX_AGE`, or every
  deferral silently becomes an eviction. Killed by collapsing the constant back
  to 15 s.
- `capacity_verdict_refuses_exactly_when_check_capacity_does` — the verdict this
  gate reads and the pre-check the dial reads have to refuse on the same
  condition. A verdict stricter than the pre-check is the harmful direction: a
  node the pre-check would let write stops looking for keys it could have
  stored. Built on a store that has deleted more than one chunk's worth of pages
  rather than on a bare full disk, because that is the state the two are about
  to part company over. Killed by applying #210's predicate to the pre-check:
  the other 933 tests still pass and only this one fails.

**Ordering against #210.** #210 (`fix/capacity-check-lmdb-reuse`) rewrites the
same pre-check this gate reads, making it two-part — below the reserve *and* out
of reusable pages inside the store — because LMDB returns a deleted record's
pages to its own free list and never to the filesystem. The two changes edit
neighbouring code in `src/storage/lmdb.rs`, so they conflict textually whichever
lands first, but the conflict that matters is not textual. With #210's predicate
in and this verdict left as it stands, a heavily pruned node reads as `Full`
here while its own dial pre-check admits the write, so it stands its keys down
for five minutes at a time instead of looking for chunks it could store. That is
under-replication, and it is the risk raised against #202 and this change
together. The test above fails in exactly that state, so whichever merges second
has to carry the second half of the predicate into the verdict rather than
resolve the conflict by hand and move on.

**What the unit tests do not do**, since it is the obvious question to ask of
them: none executes either gate. The gates live inside `run_verification_cycle`,
which needs storage, a p2p node and queues, so the e2e below is what proves them,
per gate, by mutation. The unit tests pin the pieces the gates are built from —
the verdict, the policy constant, and the deferral's interaction with eviction.
Each says so in its own rustdoc.

One thing stays unpinned by any test: which of `capacity_deferred_probe` and
`capacity_deferred_promote` a given gate increments. Reaching it means returning
cycle statistics out of a 400-line async function, which is a larger change than
the label is worth — a swap would mislabel one diagnostic line on the dev testnet,
not misbehave on the network. It is called out here rather than papered over.

**E2E (`cargo test --test e2e --features test-utils -- --test-threads=1`)** — no
new test and no new testnet. The probe assertions are folded into #202's existing
write-blocked test in `tests/e2e/fetch_local_write_guard.rs`, renamed
`write_blocked_node_neither_probes_nor_dials` to say what it now covers. All
three phases run on the one 12-node network that test already built, because a
second one is real port pressure on a CI runner.

- Phase 1 (unchanged, from #202): the holder serves `0` chunks, so no dial.
- Phase 2: `blocked_probes=0` against `control_probes=7`, with the control also
  completing its acquisition (`control_stored=true`) so a cycle that never
  reached the probe cannot read as a pass.
- Phase 3: the promotion gate on the network-authorization branch, which was
  inspection-only until this revision. The chunk is hosted widely enough to clear
  the four-of-seven quorum threshold and deliberately left out of the target's
  paid list, so the key takes that branch; it is then held at `quorum_delay=299.9s`
  rather than promoted, with `quorum_authorized=true`.

  That second value is what pins the gate *behind* authorization rather than in
  front of it. A delay on its own cannot tell those apart, which a review caught:
  a gate moved ahead of the quorum round holds the key for the same five minutes
  while never learning it was paid for, and that is the failure mode that strands
  keys, keeps bootstrap drain pending and leaves audits disabled. Moving the gate
  ahead of the round reproduces it exactly — `quorum_delay=299.9s` still, but
  `quorum_authorized=false` — and the assertion fails. Gate 1's ordering behind
  the terminal checks remains inspection-only; building a node that both holds a
  key and refuses writes is not constructible here, as the module header records.

**Each gate is pinned independently.** Disabling the promotion gate alone leaves
phase 2 passing (`blocked_probes=0`) and fails phase 3 with `quorum_delay=None` —
the key was promoted, refused at the dial, and requeued on the 15 s schedule.
Disabling both fails phase 2 first with `blocked_probes=7`. Neither gate can be
removed without a test catching it. Cutting the capacity-specific `FetchResult`
backoff is what made phase 3 observable at all, since a refused fetch now requeues
on a visibly different schedule from a gated one.

Probes are counted **per key**, not per peer. A write-blocked node legitimately
keeps sending verification requests for keys it has not yet authorized — that is
`PaidForList` convergence, which this change deliberately leaves alone — so a
per-peer count would have asserted something untrue and been flaky against
background traffic.

Phase 2 asserts both the zero probe count and the long delay; the probe count is
what discriminates the gate, the delay only confirms the key was held rather than
dropped. An earlier revision asserted on the delay *alone* and was vacuous — with
the gates removed the key was promoted, refused at the dial, and requeued on the
same long backoff, so the delay read ~299.8 s either way. The mutation run showed
that, not the code review.

Phase 3 can assert on the delay as its discriminator only because the
capacity-specific `FetchResult` backoff was cut: a refused fetch now requeues on
the ordinary 15 s schedule, so 5 minutes versus 15 seconds separates "held before
promotion" from "promoted, then refused".

**What this does not do**, which earlier drafts of this PR and of ADR-0011 both
claimed it did: it does not silence a full node. A key the node has not yet authorized still runs
its quorum round, because that is how `PaidForList` converges, and its close group
still pays a presence lookup for it. What goes away is the *repeating* cost on
keys it has already authorized and cannot store — the loop that scaled with the
owed set. Traffic falls to a convergence floor, not to zero.

A key deferred inside the last five minutes of its entry's life expires without a
further look and returns on the next neighbour-sync hint. That is the trade for
not carrying a clamp whose tail looks can themselves become ungated quorum
rounds, working against the point of the change.

**Open gate: the T2 dev testnet run.** The tier's evidence bar is "Dev testnet +
ADR". The ADR is in this PR; the run is not, and I cannot run one. The natural
shape is a repeat of the V2-987 plan with this change in, because that plan
already builds the condition under test — 25% of the fleet artificially
storage-full — and it would settle the criterion that failed there rather than
leaving it as a judgement call.

Stated as the two-sided criterion asked for in review, so whoever runs it does
not have to reconstruct what a pass looks like:

- **While full.** The full cohort's per-hour delta of
  `verification_request_tx_bytes` settles instead of climbing 1.06 → 1.77 →
  2.53 GB across the window. The counter is cumulative, so the criterion is its
  growth rate, not its value. The floor is not zero: a key the node has not yet
  authorized still runs its quorum round, which is how `PaidForList` converges,
  so what should disappear is the repeating cost on keys it has already
  authorized.
- **After capacity returns.** Freed nodes resume acquisition inside the plan's
  60-minute recovery allowance, with no key silently dropped. A deferred key is
  scheduled for its next look `CAPACITY_BLOCKED_RETRY` (5 minutes) ahead — the
  requested delay rather than a bound, since cycle polling and a backlog can
  push it later — against a `PENDING_VERIFY_MAX_AGE` of 30 minutes. Nothing
  caches the refusal: the verdict re-reads capacity every cycle and only passing
  results are cached. Both halves are pinned by unit tests; measuring the
  recovery on a fleet is what the run adds.
- **Unchanged either way.** `key_absent` audit rates and download success.

Treating that run as a merge gate is the reason this PR is marked ready for
review rather than ready to merge. The other gate raised in review — rebasing
onto current `main` so the branch resolves `h2 0.4.16` rather than `0.4.15` — is
closed: the branch sits on `661e68a`, which carries #209, and Security Audit is
green along with the other fourteen checks.

**A second flake, in this PR's own test, fixed rather than tolerated.** Phases 2
and 3 asserted that their explicit `enqueue_pending_verify_for_test` call was
what *admitted* the key. Seeding the chunk on other nodes is what makes it
advertisable, so a replica hint can reach the target's queue first, and
`add_pending_verify` then merges the hint into the entry already there and
reports `AlreadyPresent`. Phase 3 lost that race once on a Windows runner —
seeding eleven nodes is what widens the window — and it failed at the setup step
rather than at anything a gate decides, with phases 1 and 2 passing in the same
run (`served_by_holder=0`, `blocked_probes=0` against `control_probes=7`). The
phases now require the key to *be* pending verification, newly admitted or
already there, and still stop hard if it is neither, so a missing queue entry
cannot be read as a gate decision.

**Known harness flake, not from this change.** The e2e suite intermittently fails
at testnet *construction* with `Failed to create dual-stack network nodes: Failed
to create transport` — a socket bind, before any replication code runs. It has hit
all three CI platforms and six distinct tests, always at setup and never on an
assertion, including on 13 August on an unrelated branch that passed on its next
run. It also reproduced locally after many back-to-back suite runs, and both
affected tests passed immediately on isolated rerun, so it is not CI-specific;
port exhaustion under repeated testnet construction is the plausible cause rather
than a proven one. If a test job is red, check for that signature before reading
it as a defect.

## New dependency

none

## ADR

New: [**ADR-0011** — capacity-gated source discovery in the replication verification
cycle](https://github.com/WithAutonomi/ant-node/blob/fix/capacity-aware-verification-cycle/docs/adr/ADR-0011-capacity-gated-source-discovery.md),
`Status: Proposed`. It records the decision, the option that was implemented first
and rejected in review (gating on hint provenance), the clamp, and the trade-offs.

Related but not restating this decision: ADR-0003 decides how *other* nodes treat a
full peer; this decides what the full peer itself stops doing. ADR-0005 owns the
verification/fetch pipeline being changed.

## Mitigation / rollback

Revert the commit. The change is two branches in `run_verification_cycle`, both
behind a capacity verdict that is `Full` only on a node whose available space is
below its configured reserve, so a node that can write never reaches the changed
paths. What else the diff adds is inert by construction: a retry constant, a
storage verdict that reuses the existing capacity check and its cache, per-gate
counters on an existing log line, and test-only probe recording. `scheduling.rs`
has no production change.

### DIFF
--- a/docs/adr/ADR-0011-capacity-gated-source-discovery.md
+++ b/docs/adr/ADR-0011-capacity-gated-source-discovery.md
@@ -174,6 +174,16 @@ set.
   one of those is not suppressed. The default 500 MiB reserve normally makes the node
   refuse chunks well before literal ENOSPC, which is why this predicate covers the
   observed condition — but it is narrower than the name suggests.
+- The verdict restates the pre-check's predicate rather than sharing its code, so the two
+  can drift apart. That is not hypothetical. ant-node #210 makes the pre-check two-part —
+  below the reserve *and* out of reusable pages inside the store, since LMDB returns a
+  deleted record's pages to its own free list and never to the filesystem. Left
+  unreconciled, a heavily pruned node would read as `Full` here while its own dial
+  pre-check admits the write, standing its keys down for five minutes at a time instead
+  of looking for chunks it could store. That is under-replication, which is worse than
+  the probe traffic this gate exists to remove, so the coupling is held by a test rather
+  than by a comment: `capacity_verdict_refuses_exactly_when_check_capacity_does` fails in
+  exactly that state.
 - The second gate is now covered by a test rather than by inspection. Hosting the chunk
   on every other node clears the four-of-seven quorum threshold, and leaving it out of
   the target's paid list forces the network branch. Cutting the capacity-specific
@@ -196,6 +206,13 @@ set.
 - Unit: the existing `apply_fetch_result` cases still hold with `LocalWriteFailed` back
   to its upstream shape — no alternate source is conscripted, and the no-metadata path
   stays terminal.
+- Unit: `capacity_verdict_refuses_exactly_when_check_capacity_does` holds the gate's
+  verdict and the dial's pre-check to one predicate. It is built on a store that has
+  deleted more than one chunk's worth of pages rather than on a bare full disk, because
+  that is the state the two predicates are about to part company over; on a bare full
+  disk they still agree, so a test built on one would pass straight through the
+  divergence. Applying #210's predicate to the pre-check fails this test and nothing
+  else in the 934-test suite.
 - E2E: `write_blocked_node_neither_probes_nor_dials` runs all three phases on one
   network — no dial, then no probe — seeding two nodes identically through the local
   paid-list path so they differ only in whether storage accepts writes. Probes are
@@ -228,7 +245,9 @@ set.
   the plan's 60-minute allowance. Flat *near zero* is the wrong expectation either way:
   unauthorized keys still run their quorum round by design.
 - Review trigger: if `PENDING_VERIFY_MAX_AGE`, the neighbour-sync cadence, or
-  `MAX_PENDING_VERIFY` change, revisit the 5-minute constant.
+  `MAX_PENDING_VERIFY` change, revisit the 5-minute constant. If what
+  `LmdbStorage::check_capacity` refuses on changes, carry the change into
+  `capacity_verdict` in the same commit.
 
 ## Notes for AI-assisted work
 
--- a/docs/adr/ADR-0011-capacity-gated-source-discovery.md
+++ b/docs/adr/ADR-0011-capacity-gated-source-discovery.md
@@ -1,0 +1,237 @@
+# ADR-0011: Capacity-gated source discovery in the replication verification cycle
+
+- **Status:** Proposed
+- **Date:** 2026-08-18
+- **Decision owners:** Anselme
+- **Reviewers:** <pending>
+- **Supersedes:** none
+- **Superseded by:** none
+- **Related:** ADR-0003 (full-node detection and eviction — decides how *other* nodes treat a full peer; this decides what the full peer itself stops doing); ADR-0005 (replication repair hardening — owns the verification/fetch pipeline this changes). V2-1009 is the issue, V2-963 / PR #202 added the pre-check this builds on, V2-987 is the testnet run that measured the cost.
+
+## Context
+
+PR #202 stopped a node that cannot write from *pulling* chunks: `execute_single_fetch`
+runs `check_capacity()` before it dials, and a local write failure is classified as
+`LocalWriteFailed` so no alternate holder is conscripted. That removed 93.9% of
+replication egress on a local testnet with a quarter of the nodes capacity-blocked.
+
+It did not remove the round that *precedes* the dial. A node that cannot write still
+holds none of the keys it owes, so none of them terminate. Each one is selected by the
+verification cycle, probed against its close group for a holder, promoted, refused by
+the capacity pre-check, and requeued at `VERIFICATION_REQUEST_TIMEOUT`. At 15 seconds
+that is up to 120 close-group probes per entry lifetime, for every key the node owes —
+nominally, since the 15 seconds is a requested delay and cycle polling, round duration
+and bounded per-cycle selection all stretch it. The owed set grows with the network
+rather than with anything the node can do about it.
+
+The testnet run on V2-987 measured it at scale: 195 services with 50 (25.6%) held
+storage-full produced **99.6% of every `verification_request_tx_bytes` on the network**
+from that quarter of the fleet. Freeing half the full VMs collapsed their
+verification-request traffic 87% within 28 minutes while the untouched half moved
++9.5%, which is what attributes the traffic to the storage-full condition rather than
+to the fetch guard.
+
+The absolute volume in that run was small. The reason to act is the scaling: the
+per-node cost is `owed_keys x close_group x 32 B` per round, `MAX_PENDING_VERIFY` is
+131,072 keys, and every probe costs a close-group neighbour an LMDB presence lookup on
+its serial replication message path.
+
+## Decision Drivers
+
+- A node that cannot write should stop *repeating* work that cannot succeed. Re-probing
+  for a key it has already authorized and still cannot store buys nothing, and its
+  neighbours pay a presence lookup for each attempt.
+- Suppressing work must not suppress duties that do not need a local write. A full node
+  must keep learning which keys it owes were paid for, and must keep retiring keys it
+  already holds or is no longer admitted for, or it stalls its own bootstrap drain and
+  keeps its audits disabled.
+- Recovery should stay prompt. Capacity comes back — an operator resizes a volume, a
+  prune pass frees space — and the node should notice while the duty is still live. That
+  is the objective, not a guarantee: see the trade-offs below for when it is missed.
+- No wire or storage change, and no public API change in a default build. This rides the
+  release train into a mixed fleet, and an upgraded node must stay wire-compatible with
+  old peers — which it does, since no peer-side behaviour depends on receiving the
+  requests it stops sending.
+
+## Considered Options
+
+1. **Gate on hint provenance at the top of the cycle.** Drop every key whose hint
+   carried a possession claim before the cycle does anything else.
+2. **Gate the two steps that exist to enable a fetch**, after authorization and after
+   the terminal checks: the presence probe that discovers holders, and the promotion
+   that queues the download.
+3. **Leave the cycle alone and only lengthen the requeue delay** after a refused fetch.
+4. **Do nothing.** The measured volume was ~6 KB/s per full node.
+
+## Decision
+
+We will take option 2: gate source discovery, not the cycle.
+
+`run_verification_cycle` reads the storage capacity verdict once and uses it in exactly
+two places — before the local paid-list presence probe, and before promotion on the
+network-verified path. Everything else runs unchanged on a full node: the local
+paid-list fast path, `PaidForList` convergence through the quorum round, and the
+terminal checks that retire a key the node already holds or is no longer admitted for.
+
+Held keys are deferred by `CAPACITY_BLOCKED_RETRY` (5 minutes) through the existing
+`defer_pending`, and deliberately nothing more: an earlier revision also clamped the deferral to
+half the entry's remaining life, gave the deferred key fairness-eviction protection, and
+carried a capacity verdict through `FetchResult` so a refused fetch took the same long
+backoff. All three were cut on review.
+
+The clamp existed so an aged entry could not be retired without one last look. It cost a
+logarithmic burst of looks near expiry, and — where the paid-list insert had failed —
+those looks become *ungated* quorum rounds, working against the thing this ADR is for.
+Without it, a key deferred inside the last five minutes of its life expires without a
+final look and is re-acquired from the next neighbour-sync hint, which is what the
+30-minute TTL already implies.
+
+The eviction protection preserved a duty the node cannot discharge, letting a full node
+hoard pending entries against the 131,072-entry ceiling. Losing one to fairness
+reclamation is the better outcome, and it returns on the next hint.
+
+The `FetchResult` propagation duplicated the gate. After a refused fetch the key returns
+to pending and the gate at the head of the next cycle defers it for minutes anyway, so
+the machinery bought one 15-second round on a race the gate already makes rare.
+
+Only a **full** disk drives the gate. `check_capacity()` answers `Err` both for
+insufficient space and for a space query that itself failed, and neither the second nor a
+failed resize, write transaction or commit is a standing condition. A three-way
+`CapacityVerdict` separates `Full` from `Unknown`, and the gate reads only `Full`. The
+fetch path is left alone entirely: a refused fetch returns `LocalWriteFailed` and requeues
+on the ordinary schedule, because the gate at the head of the next cycle re-decides.
+
+Option 1 was implemented first and rejected in review: `HintPipeline` records only
+whether an advertiser claimed possession, so gating on it skips `PaidForList`
+convergence for keys that arrive as replica hints, strands keys that should have
+retired, and can be flipped by any routing-table peer adding a false possession claim to
+a paid-only key. Option 3 leaves the dominant cost in place, because the probe is paid
+before the fetch is ever attempted. Option 4 ignores that the cost scales with the owed
+set.
+
+## Consequences
+
+### Positive
+
+- The loop this targets goes away: a key the node has **already authorized** and cannot
+  store is no longer re-probed on every cycle it comes back on, for the rest of its
+  entry's life. That is the loop that scales with the owed set. (The 15 s request timeout
+  is the requested eligibility, not the achieved cadence — poll latency, round duration
+  and bounded per-cycle selection all stretch it.)
+
+  **How much of V2-987's 99.6% that removes is a hypothesis, not a measurement.**
+  `verification_request_tx_bytes` aggregates both branches — the gated local-paid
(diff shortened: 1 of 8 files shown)


dirvine (commented):
Reviewed exact head `fe9802c5495d744ad78410d766c71206ac8ec351`.

No code-level blocker found. The two gates are in the right places: locally authorised keys are deferred before source discovery, while newly verified keys still complete paid-list authorisation before download promotion is deferred. `Full` backs off for five minutes; a capacity-query failure remains `Unknown` and preserves the existing retry path. Deferred entries keep their original lifetime, so the change does not create immortal pending work.

Local verification:
- `cargo test --lib` — 933 passed
- focused 12-node E2E `write_blocked_node_neither_probes_nor_dials` — pass: `blocked_probes=0`, control `=7`, blocked delay ~300 s, and the network-verified key was authorised before deferral

Two merge gates remain:
1. Rebase/update onto current `main`. This head still resolves `h2 0.4.15`, so Security Audit is red; current `main` carries #209 / `h2 0.4.16`. I locally applied this commit onto current `main` cleanly and retained `0.4.16`.
2. Record the Tier-2 live-test evidence in the PR: verification-request traffic should stop for already-authorised keys while full and resume after capacity returns.

The independent review panel timed out twice before producing substantive findings, so this is my direct review rather than a panel consensus.

dirvine (approved):
APPROVE — reviewed exact head `b455064a7e9540b99f743cf85968776c25da30e7`.

No material blockers found. Full nodes defer both source probing and fetch promotion, while terminal possession/admission checks and quorum accountability continue to run. Unknown capacity remains fail-open for transient stat failures, recovery is detected promptly, and deferral does not extend entry lifetime. CI is green. Local verification: formatting and all 23 focused capacity tests passed.

jacderida:
## Testnet evidence — PASS (V2-1055, DEV-01 run 538)

Tested at scale on a 195-service testnet against **V2-987 as the reference arm** — the same plan (200 nodes, 25% artificially storage-full, continuous workload, capacity release partway) run previously on the binary *without* this change. That makes this a controlled before/after rather than an absolute check.

**Binary:** `jacderida/ant-node` `testnet/pr207-pr210-combined` @ `239aa94` — #207 merged with #210 onto `main`, including the reconciliation this PR's body prescribes (`capacity_verdict()` consults `reusable_bytes()` when the disk half reads `Full`).

### The headline: the verification lane no longer climbs

V2-987 recorded the full cohort's `verification_request_tx_bytes` per-hour delta climbing **1.06 → 1.77 → 2.53 GB** across the 3 h measurement window — bytes displaced into the verification lane rather than removed. That was the criterion it failed.

Same plan, same cohort size, with this change:

| Full cohort (50 nodes) | hour 1 | hour 2 | hour 3 | final/first |
| -- | -- | -- | -- | -- |
| **This PR** | **0.540 MB** | **0.488 MB** | **0.542 MB** | **100.4%** |
| Reference arm (no #207) | 1.06 GB | 1.77 GB | 2.53 GB | 237% |

Flat, and **~1960× smaller in absolute terms**. `verification_response_tx_bytes` behaves the same: 0.784 / 0.762 / 0.714 MB (91.1%).

The floor is non-zero (~0.5 MB/h) exactly as expected — keys the node has not yet authorized still run their quorum round, and the gate does not touch that path.

**Control validation:** the *healthy* tier's fetch lane ran at 16.41 GB/h against the reference arm's 16.65 GB/h (−1.5%). Same workload, same topology — so the full-cohort collapse is attributable to the binary, not to a quieter network.

### Recovery when capacity returns

Filler removed on 5 of the 10 full VMs with **no restart** (`ActiveEnterTimestamp` on every freed service is still T0), so this exercises the live re-read path — deferred keys had aged out at the 30-minute cap, so reacquisition had to arrive via fresh hints. The 5 still-full VMs are an untouched control. Matched 60-minute windows either side:

| | freed (5 VMs) | still-full control (5 VMs) |
| -- | -- | -- |
| `fetch_response_rx_bytes` | 0 → **48.2 GB** | 0 → **0** |
| `fetch_request_tx_count` | 0 → **17,438 dials** | 0 → **0** |
| `verification_request_tx_bytes` | 0.341 → 4.044 MB | 0.213 → 0.184 MB |
| store on disk | **9.4 – 12.8 GB** | 67 – 95 MiB |

Store growth was visible within **~7 minutes** (the acceptance window was 60). Note the freed cohort's verification traffic *rises* — that is the gate working in the intended direction: a newly-writable node resumes the cycle it had been standing down.

### Regression checks (the #202 guard must still hold)

- **Full cohort pulls nothing:** `fetch_response_rx_bytes` = **0.000 MB** and `fetch_request_tx_count` = **0 dials** in all three hours. Not merely small — exactly zero across 50 nodes for 3 hours.
- **No post-download store failures:** WARN `Failed to store fetched record` = **0** on the full cohort.
- **Single-delivery economics:** fleet fetch requests 6,460 / 6,217 / 6,074 against responses 6,452 / 6,198 / 6,102 — 1:1, no repeat-delivery multiplier.
- **Durability unchanged:** downloads **2,284 / 2,284 = 100%**; phase B alone was 1,156 uploads + 1,375 downloads with zero failures. `responsible_audit_fail_key_absent` fleet-wide 4 / 7 / 3 — *not* rising — while `responsible_audit_pass` rose 35 / 67 / 57.
- **Stability:** **0 restarts, 0 failed units, 0 panics** across 39 VMs / 195 services (checked over SSH via `NRestarts`/`--state=failed`/journald, since panics never reach the log forwarder). 79 ERRORs in 3 h over 31 hosts, all `Audit failure` classes — the reference arm had 77 in 4 h of the same.

### Setup

39 OVH 1-AZ VMs (RBX-A/SBG5/DE1, b3-8, 100 GB volumes) × 5 services = 195; `nat_percent: 30`, `full_percent: 35`, `full_headroom_mb: 100` → 10 full VMs = 50 nodes (25.6%), capacity-blocked from T0. `log_level: info`, so this PR's per-gate counters (`capacity_deferred_probe`/`_promote`, DEBUG) were not fleet-visible — the acceptance case rides the cumulative `ant_node::replication::traffic` INFO summaries. Binary identity was pinned two ways: in-band `tail.commit` showed one commit fleet-wide, and an md5 of `/usr/local/bin/ant-node` was checked on all 39 VMs every 15 minutes — zero drift.

Full write-up: [V2-1055](https://linear.app/autonominetwork/issue/V2-1055).

jacderida:
Heads-up: #210 has been merged to `main` (`c35535c`), and I've pushed a merge of `main` into this branch as `96556d3` so it can land while you're away tomorrow — using the maintainer-edit permission on the PR.

**What the merge commit does**, following the "Ordering against #210" paragraph in this PR's description:

- Resolves the textual conflict in `src/storage/lmdb.rs` (both PRs insert after `check_capacity`; #210 removed `check_disk_space`, which is now dropped here too since nothing calls it).
- Carries #210's second predicate half into the verdict: `capacity_verdict()` now consults `reusable_bytes()` when the disk half reads `Full`, returning `Writable` if the store holds at least one chunk's worth of reusable pages (uncached, matching `check_capacity`), `Full` otherwise, `Unknown` if the page query fails. So the gate and the dial pre-check refuse on the same condition, and a pruned node below its reserve keeps discovering holders for keys it can still store. `verdict_from_available_space` is unchanged in behaviour; its rustdoc now says it maps the *disk half*.
- No other files were touched by hand.

**Verification** (locally, on this exact tree): `cargo test --lib` 941/941, including `capacity_verdict_refuses_exactly_when_check_capacity_does` passing on the divergent pruned-store state it was built for; `write_blocked_node_neither_probes_nor_dials` e2e passes; fmt and clippy (`--all-features --all-targets -D warnings`) clean.

**Testnet**: the resulting tree is byte-identical to the V2-1055 build (`jacderida/ant-node` `239aa94`), which repeats the V2-987 plan with the two-sided criterion from your review — so the T2 run validates exactly what's being merged. Plan: https://linear.app/autonominetwork/issue/V2-1055

If you'd have resolved it differently, it's a single merge commit on top of your `b455064` — easy to redo.

PR #208, opened 2026-08-18
feat: restrict the beta upgrade channel to -beta pre-releases

jacderida:
## Linear issue

V2-1010 and V2-1011 — https://linear.app/autonominetwork/issue/V2-1010/ant-node-restrict-the-beta-upgrade-channel-to-beta-pre-releases

## Risk tier

Tier proposed as T2 because it changes the upgrade mechanism — which builds a node will install.
No protocol, storage or economics surface is touched.

## Compatibility

- Wire: none.
- Storage: none.
- API: no signature changes. `UpgradeMonitor::version_matches_channel` is still `pub` with the same
  signature, but on the `Beta` channel it now returns `false` for pre-releases whose first identifier
  is not exactly `beta` (previously it returned `true` for every version). `UpgradeConfig` and the
  `--upgrade-channel` / `ANT_UPGRADE_CHANNEL` surface are unchanged.

## Semver impact

- [ ] breaking
- [x] feature
- [ ] fix

## Test evidence

**Dev testnet — DEV-01, 2026-08-18: PASS.** Full results on
[V2-1012](https://linear.app/autonominetwork/issue/V2-1012/testnet-run-beta-channel-isolation-fake-beta-pulled-real-rc-ignored).

107 services (100 WAN nodes across 20 DO VMs in lon1/nyc1, plus 6 bootstraps and genesis), built
from this branch, running `--upgrade-channel beta` explicitly rather than a patched default.
Fake prerelease `v0.17.11-beta.1` published 14:20:30Z; start version `0.17.1`.

| Check | Result |
|---|---|
| Services that detected and scheduled the upgrade | 107 / 107 |
| VMs with binary at `0.17.11-beta.1` | 27 / 27 |
| Services running after upgrade | 107 / 107 |
| Processes executing the current binary inode | 107 / 107 |
| ERROR/WARN on `ant_node::upgrade*` | 0 |
| Services left pending | 0 |

Applies spread evenly across the compressed 2h staged window (15:25:02Z–17:23:44Z; per-15-min
buckets 4 / 14 / 12 / 14 / 13 / 9 / 18 / 17 / 6). Co-located services sharing
`/usr/local/bin/ant-node` handled the shared binary correctly — first writer replaces it, the rest
log `Binary already upgraded ... skipping replacement`; no contention or partial writes.

Network health held throughout: **956/956 uploads and 1130/1130 downloads succeeded**, zero
failures, with 356 operations landing inside the rollout window while the fleet was version-mixed.
Chunk stores stayed intact across the restart.

This run also exercises the release-workflow half of the change: `v0.17.11-beta.1` was built,
signed and published as a prerelease by the workflow *as modified in this PR*, and every node then
selected, downloaded and signature-verified it. That confirms end-to-end that a `v*-beta.N` tag
triggers the workflow, is flagged a prerelease, and ships a complete artifact set whose names and
`.sig` pairing satisfy `select_upgrade_from_releases`.

Scope note: this run evidences beta uptake and network health. Rejection of `-rc.*` is covered by
the unit tests below rather than by the run.

**Unit — `cargo test --lib upgrade::monitor`: 25 passed, 0 failed.** Four cases added:

| Case | Expectation |
|---|---|
| `test_stable_channel_filters_rc` | stable rejects `-rc.N` |
| `test_beta_channel_rejects_rc` | beta rejects `-rc.1`, `-alpha.1`, `-betax.1`; accepts bare `-beta` |
| `test_select_upgrade_never_picks_rc` | over `0.16.0` / `0.17.0-beta.1` / `0.17.0-rc.1`: stable → `0.16.0`, beta → `0.17.0-beta.1` |
| `test_select_upgrade_beta_ship_and_promote_same_day` | from `0.16.0-beta.1`, with `0.16.0` and `0.17.0-beta.1` both published, beta → `0.17.0-beta.1` |

Existing coverage for stable-rejects-beta and beta-accepts-beta was kept and extended.

**CI — green.** Clippy, Format, Documentation, Builds and Tests on all three platforms, ADR
validation, `linear-link` and `pr-template` all pass. Security Audit fails on RUSTSEC-2026-0258,
which also fails on `main` at the base commit `3cda0dd`; this PR changes no dependencies.

## New dependency

none

## ADR

https://github.com/jacderida/ant-node/blob/feat/beta-release-channel/docs/adr/ADR-0010-beta-upgrade-channel-semantics.md

(ADR-0010, Status: Proposed — acceptance is a human decision.)

## Mitigation / rollback

Revert `cb40bca` to restore `Beta => true`; the other commits are workflow, docs and tooling and are
independently revertible. Blast radius is limited to nodes explicitly opted into the beta channel —
the `stable` default is untouched, so the production fleet behaves identically before and after.

---

## Summary

- **Beta channel semantics (V2-1010).** The beta channel accepted every pre-release, so beta nodes
  would install `vX.Y.Z-rc.N` as soon as the Tuesday cut published it — before the release gates
  returned a verdict — and because semver ranks `-rc` above `-beta`, a node soaking a beta build
  would abandon it for the next cut's rc. Beta now accepts a version only when its pre-release
  component is empty or its first identifier is exactly `beta`. The rule previously existed in two
  places that could drift; it is now one free function used by both.

- **Release workflow (V2-1011).** Audited against `v*-beta.N` and found already correct — the `v*`
  glob matches, the version regex accepts it, `beta` was already in the prerelease test, and
  crates.io publish is skipped for any prerelease. Two implicit things made explicit: `make_latest`
  is now passed rather than relying on GitHub's implicit rule, and the release body's auto-upgrade
  claim is conditional, since only beta nodes install `-beta.N` and nothing installs `-rc.N`.

- **Test-prep commands (V2-1011).** Both `prepare-for-*-test` commands built a fake `-rc.1` release
  and flipped the compiled-in default channel to Beta — a combination this change makes inert. They
  now drive a `-beta.1` release with nodes opting in via `--upgrade-channel beta`, which is what a
  real community beta node does. These files were previously untracked; `.gitignore` is narrowed to
  `.claude/*` with an exception for `.claude/commands/` so they are reviewable.

### DIFF
--- a/.claude/commands/prepare-for-auto-upgrade-test.md
+++ b/.claude/commands/prepare-for-auto-upgrade-test.md
@@ -1,0 +1,123 @@
+# Prepare for Auto Upgrade Test
+
+@Cargo.toml
+@src/config.rs
+@src/upgrade/monitor.rs
+@src/node.rs
+@.github/workflows/release.yml
+
+I want you to help me prepare for a test of the automatic-upgrades process.
+
+## Overview
+
+The process has 4 phases:
+
+* Derive test parameters from the current version
+* Clear previous test artifacts
+* Prepare the current branch for an auto-upgrade test
+* Create a fake beta release branch and trigger the release
+
+The fake release uses a `-beta.N` suffix, not `-rc.N`. The beta upgrade channel accepts a version
+only when its pre-release component is empty or its first identifier is exactly `beta`; release
+candidates are deliberately rejected on every channel, so an `-rc.1` fake release would never be
+picked up by anything.
+
+# Phase 1: Derive Test Parameters
+
+We need to derive all test parameters from the current `Cargo.toml` version.
+
+Follow these steps:
+- Read the current version from `Cargo.toml` (the `version` field under `[package]`).
+- Create a `BETA_VERSION` parameter by adding 10 to the current PATCH component and appending a
+  `-beta.1` suffix. For example, if the current version is `0.3.2`, the beta version should be
+  `0.3.12-beta.1`.
+- Create a `TAG_NAME` parameter by prepending `v` to the `BETA_VERSION`. For example:
+  `v0.3.12-beta.1`.
+- Create a `BRANCH_NAME` parameter by taking the version with the bumped patch (without the
+  `-beta.1` suffix) and prepending `beta-`. For example: `beta-0.3.12`.
+
+Print the values you have obtained for these parameters and prompt me to verify them before
+proceeding to the next phase.
+
+# Phase 2: Clear Previous Test Artifacts
+
+It's possible there was a previous test run that has not been cleaned up. We need to remove any
+artifacts from any previous test.
+
+Follow these steps for the current `BRANCH_NAME` / `TAG_NAME`:
+- If a `BRANCH_NAME` branch exists locally, delete it.
+- If a `BRANCH_NAME` branch exists on the `upstream` remote, delete it.
+- If a `TAG_NAME` tag exists locally, delete it.
+- If a `TAG_NAME` tag exists on the `upstream` remote, delete it.
+- If a GitHub release exists for `TAG_NAME` on the upstream repository (`WithAutonomi/ant-node`),
+  delete it using `gh release delete TAG_NAME --repo WithAutonomi/ant-node --yes`.
+
+Earlier runs of this command used release-candidate names, so also clean up the legacy equivalents:
+- The `rc-<bumped version>` branch, locally and on `upstream` (e.g. `rc-0.3.12`).
+- The `v<bumped version>-rc.1` tag, locally and on `upstream` (e.g. `v0.3.12-rc.1`).
+- The GitHub release for that rc tag, if one exists.
+
+For each step, check if the artifact exists before attempting deletion. Report what was cleaned up.
+
+# Phase 3: Preparing the Current Branch
+
+For the test, the current branch needs a temporary commit that compresses the upgrade timing so
+that test nodes detect and apply the fake beta release quickly.
+
+Note that we do **not** change the default upgrade channel. Cohort nodes opt into beta explicitly
+at launch with `--upgrade-channel beta` (or `ANT_UPGRADE_CHANNEL=beta`), exactly as a real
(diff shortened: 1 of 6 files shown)


dirvine (commented):
## Hermes Agent review — head `2ebe8adc`

**Verdict: no blockers.** Clean, well-scoped change. Single free function now owns the channel rule and all three call sites (`version_matches_channel`, `process_release`, `select_upgrade_from_releases`) route through it — the drift point the PR set out to close is actually closed.

### Verified
- **Logic:** `version_matches_channel` correctly accepts finals on both channels and `-beta.*` (exact first identifier, so `betax.1` is rejected) on beta only; `-rc.*`/`-alpha.*` rejected everywhere. Matches ADR-0010.
- **Local tests:** `cargo test --lib upgrade::monitor` on the exact PR head → **25 passed, 0 failed**, including the 4 new cases.
- **CI:** green across Clippy, Format, Docs, Build and Test on all three platforms, plus ADR validation, linear-link, pr-template, self-test.

### Non-blocking notes
- **Security Audit fails on `RUSTSEC-2026-0258` (h2 unbounded empty DATA frames).** `Cargo.lock` is unchanged between `main` and this head, so this is inherited, not introduced. Rebase onto an updated `main` once that advisory is fixed; don't treat it as a blocker for this PR.
- **`make_latest` change** is effectively no-op hardening — GitHub already refuses to mark a prerelease as "Latest" — but making intent explicit is fine and matches the release-body conditional.
- **Trade-off to weigh:** with `-rc.*` rejected on every channel, internal rc soaking now has no channel home; ADR-0010 records this honestly and defers option 3 (an `rc` channel) as an additive change. That's a real capability reduction worth a conscious sign-off, not a code defect.

No inline comments needed. I have not approved or merged — acceptance/merge remain a human decision.

Reviewed by Hermes Agent.

dirvine (approved):
Approved at head `2ebe8adc5ef2e84e8772d8788ab350ded933530b`. The beta-channel filtering, shared call path, tests, release workflow and ADR were reviewed; no blockers found. The failing `h2` security audit is inherited from the unchanged lockfile and was not introduced by this PR. Reviewed by Hermes Agent.

PR #210, opened 2026-08-20
fix(storage): treat a node as full only when disk and store are both out of room

grumbach:
## Linear issue
https://linear.app/autonominetwork/issue/V2-1034/capacity-check-ignores-reusable-lmdb-pages-so-pruned-nodes-refuse

## Risk tier

This is a bug fix restoring the intended capacity check, not a new design: the two-part predicate ("full" = no disk **and** no reusable page) was always what the guard was supposed to express, and the `MapFull` half was already implemented and simply unreachable. No wire message, stored format, payment path or upgrade mechanism changes, and behaviour above the disk reserve is byte-for-byte unchanged.

Reviewer note: the only behavioural delta is on a node already under its reserve, which today refuses every write. Flag it if you read that as T2.

## Compatibility
- Wire: none. No message type, field or protocol identifier changes.
- Storage: none. No change to the LMDB layout, key or value encoding. The map ceiling is a runtime setting, not part of the file format, and LMDB clamps any request below the space in use, so a store written by this build opens unchanged on the previous one.
- API: none externally. `check_capacity` keeps its signature and its `Insufficient disk space …` error text; only the condition under which it fires changes.

## Semver impact
- [ ] breaking
- [ ] feature
- [x] fix

## Test evidence
`cargo test --lib`: **936 passed, 0 failed.** `cfd` (fmt + clippy + doc) clean.

Seven tests added or rewritten, each pinned to a claim rather than to the implementation:

| Test | Proves |
|---|---|
| `below_reserve_put_reuses_freed_pages` | The regression itself. Below the reserve, a write refused before any delete succeeds after one, and `data.mdb` does not grow doing it. |
| `below_reserve_refused_put_does_not_grow_the_file` | Refused writes leave the file and the map ceiling untouched, so the reserve is preserved. Asserts the specific error, not just any failure. |
| `large_refusal_does_not_block_a_smaller_put` | A refused maximum-sized value does not lock out a smaller one. |
| `full_store_below_reserve_can_still_delete` | A pinned store can still prune, no allowance is left in the ceiling afterwards, and pinning still refuses an oversized write. |
| `check_capacity_tracks_reusable_space_not_just_disk` | The pre-check refuses with no reusable space and reopens after pruning, with `statvfs` unchanged throughout. |
| `leaving_no_growth_restores_head_room` | Freeing disk lifts the pin and re-grows the map. |
| `above_reserve_behaviour_is_unchanged` | No pinning and no behaviour change on a healthy node. |

`test_put_rejected_on_insufficient_capacity_before_verification` (handler) still proves a genuinely full node short-circuits ahead of payment verification, which is the saving the earlier pre-check work introduced.

Adversarially reviewed over six rounds. Findings addressed rather than argued: a store-wide `MapFull` verdict that let one large chunk lock out small writes; a permanent map slack that was really ordinary put capacity and multiplied by nodes per volume; transition races between the mode flag and the map size; a torn read in the reusable estimate that could under-report; and leak paths for both the raised ceiling and the delete allowance under error, panic and cancellation.

**Not yet done:** no testnet run. Worth exercising on a deliberately filled host before it rides a train, because the behaviour only differs once a volume is under its reserve.

## New dependency
none

## ADR
n/a — bug fix, not an architectural decision.

The reasoning that would have gone in one is in the commit message and in the code comments at the decision points: why the allocator answers "does this write fit" instead of a page-count estimate, why `non_free_pages_size()` must not be called, why a delete needs a budgeted allowance, and why an async lock cannot order two resizes.

## Mitigation / rollback
Revert the commit. The change is confined to `src/storage/lmdb.rs` and one handler test, adds no state that outlives the process, and writes nothing new to disk. Behaviour above the disk reserve is unchanged, so the blast radius is limited to nodes already under their reserve, which today refuse every write regardless.

### DIFF
--- a/src/storage/handler.rs
+++ b/src/storage/handler.rs
@@ -1180,18 +1180,24 @@ mod tests {
         }
     }
 
-    /// V2-411: a disk-full node must reject a PUT with the disk-space error
-    /// *before* running payment verification.
+    /// A node that is genuinely full must reject a PUT with the disk-space
+    /// error *before* running payment verification (`V2-411`).
+    ///
+    /// "Full" now means both halves of the predicate: the volume is below the
+    /// reserve **and** the store has no reusable space. A freshly created store
+    /// has no freed pages, so both hold and the pre-check short-circuits, as it
+    /// always did. The companion cases in `storage::lmdb::tests` cover the half
+    /// that changed, where pruning has left reusable pages and the node must be
+    /// admitted rather than refused on `statvfs` alone.
     ///
     /// The chunk is intentionally **not** cache-inserted, so if the handler
-    /// reached `verify_payment` it would return `PaymentRequired`/`PaymentFailed`
-    /// (an uncached chunk with no proof). Observing the `StorageFailed` disk
-    /// error instead proves the disk pre-check short-circuited ahead of
-    /// verification — there is no on-chain path to reach.
+    /// reached `verify_payment` it would return `PaymentFailed` (an uncached
+    /// chunk with no proof). Observing the `StorageFailed` disk error instead
+    /// proves the pre-check short-circuited ahead of verification.
     #[tokio::test]
-    async fn test_put_rejected_on_insufficient_disk_before_verification() {
-        // u64::MAX reserve guarantees `available < reserve`, so the cached
-        // disk-space check always fails.
+    async fn test_put_rejected_on_insufficient_capacity_before_verification() {
+        // u64::MAX reserve guarantees `available < reserve`, and a fresh store
+        // has no reusable pages, so the node is full on both halves.
         let (protocol, _temp) = create_test_protocol_with_reserve(u64::MAX).await;
 
         let content = b"chunk for a disk-full node";
--- a/src/storage/lmdb.rs
+++ b/src/storage/lmdb.rs
@@ -7,12 +7,13 @@
 //! {root}/chunks.mdb/   -- LMDB environment directory
 //! ```
 
-use crate::ant_protocol::XorName;
+use crate::ant_protocol::{XorName, MAX_CHUNK_SIZE};
 use crate::error::{Error, Result};
 use crate::logging::{debug, info, trace, warn};
 use heed::types::Bytes;
 use heed::{Database, Env, EnvOpenOptions, MdbError};
 use std::path::{Path, PathBuf};
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::Arc;
 use std::time::Instant;
 use tokio::task::spawn_blocking;
@@ -70,6 +71,20 @@ const WINDOWS_MAP_HEADROOM: u64 = 32 * GIB;
 /// Between checks the cached result is trusted.  Disk space changes slowly
 /// relative to chunk-write throughput, so a multi-second window is safe.
 const DISK_CHECK_INTERVAL_SECS: u64 = 5;
+
+/// Slack granted to a *delete* that cannot copy-on-write inside the pinned map.
+///
+/// A delete is itself a write: LMDB copies the B-tree path before it frees the
+/// leaf pages. On a store with no free page at all, a map pinned exactly to the
+/// file size leaves a delete nowhere to go, so the node could not prune its way
+/// back to health.
+///
+/// The slack is granted **only** on the delete retry path and taken away again
+/// immediately, so an ordinary store can never allocate from it. Leaving it
+/// permanently in the ceiling would hand every node a little more of the very
+/// reserve this mode exists to protect, which on a shared volume multiplies by
+/// the number of nodes.
+const DELETE_COW_SLACK: u64 = 256 * 1024;
 
 /// Configuration for LMDB storage.
 #[derive(Debug, Clone)]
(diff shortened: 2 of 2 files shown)


dirvine (approved):
APPROVE — reviewed exact head `0a8540418ce518c033fc0211f5f4cb7188019230`.

No material blockers found. Map sizing accounts for live data and reusable pages; runtime resize and all database operations are coordinated under `env_lock`; cancellation paths re-check growth mode; deletion retains bounded COW slack while full; restart restores the intended high-water behaviour. CI is green. Local verification: formatting and all 23 focused LMDB tests passed.

Integration note with #207: preserve #210's two-part capacity predicate when reconciling `capacity_verdict`; the cross-PR tripwire test should remain green.

jacderida:
## Testnet evidence — PASS, with one observability finding

Tested at scale on a 195-service testnet (V2-1055, DEV-01 run 538), combined with #207 on branch `testnet/pr207-pr210-combined` @ `239aa94` — both merged onto `main` with the reconciliation #207 prescribes, so `capacity_verdict()` and `check_capacity()` refuse on the same two-part condition.

Setup: 39 OVH VMs × 5 services; 10 VMs (**50 nodes, 25.6%**) artificially storage-full at `full_headroom_mb: 100`, i.e. ~99 MiB free against the 500 MiB reserve, capacity-blocked from T0. Continuous 10/50 MB uploads and downloads for ~5.5 h, then filler removed on 5 of the 10 with the other 5 held as control.

### The two-part predicate is live and correct

Directly observable on the real decision path. Every rejection on the full cohort carries it verbatim:

```
Rejecting PUT before payment verification: storage error: Insufficient disk space:
0.09 GiB available, 0.49 GiB reserve required, and only 8192 B reusable inside the
local store. Free disk space or increase the partition to continue.
```

The `and only 8192 B reusable inside the local store` clause is this PR's change being consulted. 8192 B = two pages < one chunk, so these near-empty-store nodes correctly read as full — matching pre-#210 semantics for this cohort, which is the required behaviour. Sustained across the whole run: **71,280 rejections** during the 3 h measurement window across 10/10 hosts and all 5 service numbers (~24k/h), plus 12,103 more on the control cohort in the hour after the others were freed. No flapping.

### Clean no-growth exit

On the freed VMs after the volume went back above the reserve, over the following hour:

| pattern | count |
| -- | -- |
| `MapFull` | 0 |
| `Cannot delete: the local store is full` | 0 |
| `Failed to restore LMDB map` | 0 |
| `Failed to pin LMDB map` | 0 |
| `Failed to grant delete slack` | 0 |

All five literals were confirmed present in the pinned source first, so these are genuine behavioural absences rather than renamed strings. Store files grew again immediately — **9.4 – 12.8 GB** within the hour, against **67 – 95 MiB** on the untouched still-full control — with **no restart** (`ActiveEnterTimestamp` still T0), so `sync_growth_mode` unpinned and `try_resize` re-grew the map on the live process.

### No regression above the reserve

This PR claims byte-for-byte unchanged behaviour on healthy nodes; the 145 healthy services support that. Fetch lane 16.41 / 15.79 / 15.41 GB/h — within 1.5% of the V2-987 reference arm's 16.65 GB/h for the same hour. Fleet fetch requests-to-responses 1:1. Downloads **2,284 / 2,284 = 100%**. **0 restarts, 0 failed units, 0 panics** across all 39 VMs / 195 services. No new WARN/ERROR patterns from the resize/pin/delete-allowance machinery.

### Finding: the `pinned LMDB map` INFO line is unreachable for a node that starts full

The run plan expected every full-cohort node to log `Disk below reserve: pinned LMDB map …` during warm-up. It fired **0 times fleet-wide** over the entire run, while those same nodes were demonstrably capacity-blocked throughout. This is structural, not a defect in this PR:

- The line is emitted by `pin_map_to_high_water()` (`src/storage/lmdb.rs:1031`), reached only via `sync_growth_mode()` (`lmdb.rs:936`), whose only callers are `LmdbStore::put` (`lmdb.rs:379`) and `LmdbStore::delete` (`lmdb.rs:597`).
- But `src/storage/handler.rs:546` runs the cheap disk pre-check **before** payment verification and returns early — its comment notes "the store path keeps its own check as defence-in-depth", i.e. the store layer sits deliberately downstream of that gate.
- So a node that is below the reserve with a **near-empty store** has every PUT and fresh replication offer rejected upstream, and the pin path is never entered. The pin is reachable only by a node that crosses below the reserve *while writing*, or by a pruned node whose reusable pages let `check_capacity` pass — and that pruned case isn't constructible in a 6-hour fleet run, since the artificially-full cohort starts with no deleted pages.

Two things follow, both worth your call rather than mine:

1. **For evidence purposes this is fine** — the predicate itself is observable on the decision path (above), which is arguably better evidence than the pin line, since it shows the check being consulted where it matters. Combined with the seven unit tests and the tripwire test, I'd treat the fleet evidence for this PR as sufficient.
2. **You may want an INFO line on the no-growth *entry* decision** regardless of which layer refuses, or the pinning machinery is invisible in production logs for the most common way a node gets full (gradually, then staying full with a non-empty store — where it *would* pin, but only once a write reaches the store layer). Not a blocker; just noting that the current instrumentation can't distinguish "pinned and refusing" from "precheck-refused before ever pinning" from the outside.

Full write-up: [V2-1055](https://linear.app/autonominetwork/issue/V2-1055).

PR #215, opened 2026-08-25
feat(replication): stop penalising a peer for not holding a close-group chunk

grumbach:
## Linear issue

V2-1033

## Risk tier

## Compatibility

- Wire: none. No message shape, field, or protocol version changes.
- Storage: none.
- API: additive. `replication::config` gains `penalise_unheld_close_group_chunk`,
  `set_close_group_storage_penalty_suspended`, `close_group_storage_penalty_suspended`,
  `apply_close_group_storage_penalty_policy`, and the release constant. Nothing is removed
  or changed in shape.

Behavioural change, which is the point of the PR: peers running this build stop applying
one trust penalty. Mixed-version fleets are fine in both directions — an older peer keeps
applying it, which is exactly why this ships a release ahead of the migration rather than
alongside it.

## Semver impact

- [ ] breaking
- [x] feature
- [ ] fix

## Test evidence

```
cargo test --lib                       942 passed, 0 failed
cargo test --test e2e replication::     37 passed, 0 failed  (245 s)
cargo clippy --all-targets --all-features   0 warnings
cargo clippy -- -D warnings             clean
cargo fmt --all                         clean
RUSTDOCFLAGS=--deny=warnings cargo doc  clean
```

`possession_check_penalises_absent_peer_only_and_obeys_the_release_switch` asserts both
directions on one harness: with the penalty explicitly on, an absent peer is penalised and
a present one is not; with the switch in the position this build ships, the same probe
leaves the absent peer's trust untouched.

Reviewed by codex at xhigh, which enumerated and classified every `ApplicationFailure`
emission in `src/replication`. It found one lane I had missed, a sole-source replica hint
whose sender then denies possession, and a default that pointed the wrong way. Both fixed
here. Its remaining finding is recorded in the ADR rather than fixed; see below.

## New dependency

none

## ADR

https://github.com/grumbach/ant-node/blob/audit/suspend-unheld-chunk-penalty/docs/adr/ADR-0012-file-based-chunk-store-and-lmdb-retirement.md

## Mitigation / rollback

Revert the commit, or set `ANT_SUSPEND_UNHELD_CHUNK_PENALTY=0` on any node to restore the
penalty without a redeploy. Both are immediate and neither touches stored data. A node
running the previous build is unaffected and keeps penalising, so a partial rollout has no
correctness consequence in either direction.

---

## What this is, and why it ships alone

The chunk store can only grow. LMDB returns a deleted page to its own free list and never
to the filesystem, so deleting chunks frees no disk: last week the fleet deleted 2.29
million chunks and got back zero bytes. Moving to a store that does return space means a
node short of disk has to give some chunks up while it moves the rest across.

It cannot avoid being seen doing that, and it cannot stop the consequence, because the
penalty is the **auditor's** decision. So the auditors stop first, one release ahead, and
the migration follows in the next one.

## What is withheld, and what is not

Withheld — every way a node short of disk gets accused of not having a chunk:

- responsible-chunk audit failures
- fresh-replication possession checks (ADR-0003)
- prune audits
- a sole-source replica hint whose sender then answers `Absent`
- fetch responses where a peer that answered `Present` returns `NotFound` or `Error`

**Not** withheld:

- the commitment-bound subtree audit, where a peer published a **signed** claim to hold
  specific keys and could not answer for them
- a sole-source replica hint the close group rejects outright, which is a claim about a key
  that does not exist rather than about the sender's own storage
- bootstrap-claim abuse, malformed messages, corrupt fetch payloads, bad hints

Keeping the commitment lane is load-bearing, not a nicety. The migration works by having a
node **reduce** its signed commitment to what it can actually hold, so its peers hold it to
the smaller claim. Suspending that enforcement would make the reduction meaningless.

Audits of both kinds keep running and keep recording. Only the trust event is withheld, and
that record is how we decide when to switch the penalty back on.

## Why a build constant and not a config field

A node writes its effective configuration back to disk. Shipping this as an ordinary
setting would bake this release's value into every operator's file, and the release that
flips it back would change nothing for anyone who had ever started a node.
`ANT_SUSPEND_UNHELD_CHUNK_PENALTY` exists for a canary and warns loudly when it disagrees
with the build.

## The cost, stated plainly

Between this release and the one that restores the penalty, a peer that publishes no
commitment at all is credited as a legacy node, can answer `Present`, and can then fail to
serve without a trust cost. It pays for an identity and some traffic. One identity removes
one of seven replicas; controlling all seven positions removes the chunk's availability.

The commitment-bound audit is untouched, so this only works for a peer publishing no
commitment, which is itself visible in the record. The mitigation is not a code change: it
is not letting the third release slip. That release is deliberately a release rather than a
compiled-in expiry, so the date can move on evidence, and this is the risk that decides how
far it may move.

### DIFF
--- a/docs/adr/ADR-0012-file-based-chunk-store-and-lmdb-retirement.md
+++ b/docs/adr/ADR-0012-file-based-chunk-store-and-lmdb-retirement.md
@@ -1,0 +1,398 @@
+# ADR-0012: One File Per Chunk, and Retiring LMDB Without Losing Data
+
+- **Status:** Proposed
+- **Date:** 2026-08-25
+- **Decision owners:** Anselme Gaeremynck
+- **Reviewers:** David Irvine, Chris O'Neil, Mick van der Most van Spijk
+- **Supersedes:** none
+- **Superseded by:** none
+- **Related:** ADR-0002 (gossip-triggered subtree audit), ADR-0003 (possession checks),
+  ADR-0004 (commitment-bound quote pricing), ADR-0007 (Windows LMDB map headroom cap,
+  retired by this decision)
+
+## Context
+
+The node stores chunks in LMDB. LMDB returns a deleted page to its own free list and never
+to the filesystem, so **deleting chunks does not free disk**. In one week the fleet deleted
+2.29 million chunks and got back zero bytes. Operators read that as a bug and are tempted
+to wipe node directories to reclaim space, which costs the network real replicas.
+
+There is no partial way out. Compaction needs free space equal to the live data, which is
+exactly the condition a full node does not meet, and it does not get us off LMDB anyway.
+Punching holes in `data.mdb` is Linux-only, needs LMDB internals to identify free pages,
+and reads back as zeros. Disk comes back exactly once: when `chunks.mdb` is removed whole.
+
+    peak disk during migration = allocated chunks.mdb (unchanged) + files written so far
+
+So a local migration is possible if and only if `free >= live payload`. On production
+volumes today (55 volumes at 492 GiB, free median 25.8 GiB, p10 9.8 GiB, about 12 nodes
+per volume, about 38 GiB of LMDB per node of which about 24 GiB is live) migrating one
+node costs 24 GiB and returns 38 GiB. One at a time the host gains about 14 GiB per node
+and the queue accelerates. All twelve at once need 288 GiB and all twelve stall.
+
+The chunk workload is the easiest possible case for a filesystem: content-addressed,
+immutable, write once, read many, delete whole, and **4 MiB**, confirmed by the team
+rather than assumed. That size is what makes one file per chunk the right shape; see the
+Storj and borgbackup note under Validation for what would change the answer.
+
+## Decision Drivers
+
+- Deleting a chunk must return its blocks to the filesystem, on a full disk, with no free
+  space required and no compaction to schedule.
+- No chunk may lose its last replica, including during a fleet rollback, a skipped
+  upgrade, or a crash halfway through the migration.
+- Mass audit failures are as damaging as data loss. Nothing here may cause them.
+- It has to work for every operator, not for our fleet. Most node operators are not us and
+  cannot be told to attach a second volume.
+- No opt-in. Whatever we ship is what every node does by default.
+
+## Considered Options
+
+1. **Stay on LMDB and compact.** Needs free space equal to the live data, which is the
+   condition we are trying to escape, and leaves us on LMDB.
+2. **Append-only packs** (borg segments, Storj hashstore). Reintroduces compaction, a free
+   list, and a cross-file index. That is LMDB's disease with a different allocator.
+3. **Fixed-size slots** (Sia `hostd`, Swarm sharky). Cheaper than log packing, and Sia's
+   sector size is exactly our 4 MiB. But a freed slot returns space to the *store*, never
+   to the *filesystem*: the volume file never shrinks. It is the right design once a node
+   has a declared capacity, and the wrong one while our whole complaint is that disks stay
+   full as chunk counts drop.
+4. **One file per chunk, sharded on the address prefix.** Broken for us, see below.
+5. **One file per chunk, sharded on the address suffix.** Chosen.
+
+## Decision
+
+### The store
+
+One immutable file per chunk:
+
+```text
+{root}/chunks/layout.json          versioned layout marker
+{root}/chunks/<xy>/<64-hex>        xy = the LAST two hex characters of the address
+```
+
+**Suffix, never prefix.** A node holds keys it is among the `CLOSE_GROUP_SIZE` closest to,
+so its holdings share roughly `log2(N / 7)` leading bits with its own node ID, and that
+shared prefix grows as the network grows. Distinct directories a single node would actually
+use, sharding on the first hex characters:
+
+| nodes | shared bits | 2 hex | 3 hex | 4 hex |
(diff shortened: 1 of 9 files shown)


dirvine (approved):
Reviewed at exact head `0edc96f152dcce8f006181be04dff9386079c8df`.

The release switch is applied before audit activity, defaults to the intended R1 state, has an immediate env rollback, and routes the seven non-commitment close-group “cannot serve” penalty paths consistently. Signed commitment/subtree failures, malformed responses, corrupt fetch payloads, bootstrap abuse, bad hints and payment commitment mismatches remain enforced. Mixed-version behaviour and the temporary availability risk are stated plainly in ADR-0012.

Verified locally: focused switch/unit tests, the possession-check e2e (1 passed), default and no-default-feature builds, and clippy. Current CI is fully green across Linux, macOS and Windows.

Non-blocking follow-ups: restore the e2e test switch to the release default, add switch regression coverage for the remaining audit/fetch lanes, document the canary env var for operators, and separate responder storage-read/integrity errors from generic `FetchResponse::Error` before R3 restores penalties.

APPROVE — no material blockers found.

dirvine (changes requested):
Follow-up after the final adversarial review: **REQUEST_CHANGES** at `0edc96f152dcce8f006181be04dff9386079c8df`.

`FetchResponse::Error` at `src/replication/mod.rs:8850-8867` should not be routed through `penalise_unheld_close_group_chunk`. Its sole producer (`handle_fetch_request`, around `mod.rs:6983-6992`) maps `LmdbStorage::get() -> Err` to this variant. `LmdbStorage::get` returns `Err` for LMDB transaction/read failures, blocking-task failures, and hash-verification corruption (`src/storage/lmdb.rs:507-547`). A chunk the peer no longer holds maps separately to `Ok(None) -> FetchResponse::NotFound`.

This also explains ADR-0012’s stated “one helper, six call sites”: the implementation currently has seven, with `fetch_error` being the extra site. The inline claim that a disk-short node “answers exactly this way” is not supported by the producer semantics.

Please restore a direct `ApplicationFailure` for `FetchResponse::Error`; keep `FetchResponse::NotFound` behind the temporary switch. A focused regression test for both classifications would be useful.

The rest of the review remains positive: CI and focused tests are green, call-site coverage is otherwise consistent, and the rollout/configuration path matches the plan.

dirvine (approved):
Re-reviewed at exact head `ac68e5915c03150ff6931dd9f08b59424740fac1`.

The requested blocker is resolved: responder reads now map `Ok(None)` to `NotFound` and `Err` to `Error`; `NotFound` remains behind the temporary unheld-chunk suspension, while `Error` reports `ApplicationFailure` directly and cannot be suppressed by that switch. ADR-0012 now states the distinction explicitly and the implementation has the documented six switched call sites.

Verified locally: both new producer/classification boundary tests pass, `cargo check --lib --no-default-features` passes, and format/diff checks are clean. Two independent three-reviewer groups unanimously found no material blocker; broader replication tests and clippy also passed in independent seats. All 15 GitHub checks are green.

APPROVE. Non-blocking follow-ups: directly test the `charge_fetch_fault` trust-event side effect under suspension, and add the singleton-hint denied-possession lane to the helper documentation enumeration.

jacderida:
## Testnet run result — 990 nodes, 2026-08-28

**Verdict: nothing in this run blocks merging.** This branch rode along on the V2-1101 testnet run for #216 (tested together as `jacderida:testnet/pr216-with-main` @ `49bc578`). Rewriting an earlier comment of mine that reported measurements without answering the question that matters.

### The hazard this change introduces, and what the run says about it

The risk is that the narrowing goes too far and penalties that *should* still fire stop firing. The run has positive evidence against that, not merely an absence of alarms:

- **Commitment-bound audits still penalise.** The `ant_node::replication::storage_commitment_audit` lane ran throughout at **13.84 events/node-hour**, and `Possession check: <addr> unreachable for <addr>; penalising at audit severity` was present across the whole run.
- **The penalty rate fell slightly, it did not collapse.** `penalising at audit severity` measured **0.36 → 0.29 per node-hour (−18%)** across the swap, on the same 660 nodes. That is the shape you'd want from removing spurious penalties; a broken narrowing would have driven it to zero.
- **The path was live everywhere.** The "does NOT penalise" line appeared on **990/990** nodes, with no `ANT_SUSPEND_UNHELD_CHUNK_PENALTY` override warning anywhere — nothing silently fell back to the old behaviour.
- **Under conditions rich in exactly this case.** The fleet was running a storage-format migration that relocates chunks on every node, and 330 of the 990 nodes were destroyed mid-run with the re-replication that follows. "Node doesn't hold a close-group chunk" was abundant, and the change did not misbehave: zero panics, zero audit-related errors attributable to it.

### What this run does NOT establish

Neither of these is a merge blocker, but they are the edges of what was tested:

1. **No adversarial case.** Nothing deliberately withheld chunks to confirm a genuinely misbehaving node still gets caught through the commitment-bound path. That is the one check I'd want before trusting the narrowing completely, and it belongs in CI rather than on a testnet.
2. **Seven hours is short for reputation effects.** Under-penalising would show up as slow drift in reputation/eviction behaviour over days, not hours. Nothing in a 7h run can speak to that.

An earlier version of this comment also raised an audit-failure baseline problem. **That was my conflation and is not relevant to this PR** — that gate measures audit *timeouts* (a network/performance property), whereas this branch narrows *penalties*. The baseline was unusable because my run's baseline phase was only 15.3 minutes on a freshly started network, before first audits fire; it says nothing about this change either way.

Migration-side results are on #216. Tracked in V2-1101.

PR #216, opened 2026-08-25
feat(storage): replace the LMDB chunk store with one file per chunk, and migrate onto it

grumbach:
> **Stacked on #215.** That PR stops peers penalising a node for not holding a
> close-group chunk, and has to be out in the wild a release ahead of this one. Review
> #215 first; this branch contains its commits.

## Linear issue

Closes V2-1033

## Risk tier

## Compatibility

- Wire: none. No message shape, field, or protocol version changes.
- Storage: **changes**. Chunks move from `{root}/chunks.mdb` to `{root}/chunks/<xy>/<64-hex>`.
  A node that runs this and then downgrades finds its chunks in a directory the old build
  ignores. While the legacy store still exists every new chunk is written to it as well, so
  a downgrade at that stage loses nothing. **After a node has removed its legacy store there
  is no downgrade.** This release does remove it, once every gate below is satisfied, because
  removing it is the only thing that returns disk.
- API: additive. `ChunkStore`/`ChunkStoreConfig` are added; `LmdbStorage`/`LmdbStorageConfig`
  are still exported. `StorageStats`, `MIB` and `GIB` move from `storage::lmdb` to `storage`.

## Semver impact

- [x] breaking
- [x] feature
- [ ] fix

## Test evidence

```
cargo test --lib                            1048 passed, 0 failed
  storage::file_store                         33 passed
  storage::chunk_store                        45 passed
  storage::migration                          21 passed
cargo test --test poc_shutdown_lmdb_drain      1 passed
cargo test --test poc_audit_handler_live      16 passed
cargo test --test poc_commitment_audit_attacks 19 passed
cargo test --test poc_bootstrap_stall          3 passed
cargo test --test poc_d1_bounded_queues        7 passed
cargo test --test e2e                         96 passed, 0 failed, 3 ignored (515 s)
cargo build --no-default-features           clean
cargo clippy --all-targets --all-features    0 warnings
cargo clippy -- -D warnings                  clean
cargo fmt --all                              clean
RUSTDOCFLAGS=--deny=warnings cargo doc       clean
```

One caveat on the lint evidence: this machine runs clippy 1.95 and CI runs 1.98. A lint was
renamed between them, so a local `-D warnings` pass is not by itself proof CI will pass.

### What CI proves on every commit

Four harnesses, run on Linux, macOS and Windows. The three that touch durability run again
on ext4, XFS and btrfs loopback volumes, because how a filesystem handles a rename, a lock,
a deletion and the space behind it is its own business and btrfs in particular accounts for
it differently.

| Harness | What it settles | Mutation that turns it red |
|---|---|---|
| `migration_reclaims_disk` | Free space sampled from the filesystem before, at the two-copy peak, and after the environment is gone | Unlink the environment while holding it open: names gone, every block still allocated |
| `migration_crash_safety` | A child killed at a failpoint *inside* a publish, and another killed with the environment renamed aside and marked and nothing yet deleted | Publish non-atomically; drop the environment write; stop believing the retirement mark |
| `migration_shared_volume` | Two drivers on one volume, driving `migration::run`; the lock held through retirement as well as copying | Remove `take_volume_lock` from either branch of the driver |
| `storage_scale` | 100,000 chunks: scan time, index bytes per chunk, bytes read, directory entries per chunk | Read any part of a chunk during the scan; `stat` each entry; write a sidecar per chunk |

Two of those deserve saying out loud, because the obvious version of each test does not
catch them. A `stat` behind every filename costs about three times a bare walk and stays
well inside any flat ceiling loose enough not to flake, so the scan is measured against the
machine instead: the same directory walked twice in the same process, once reading names and
once calling `metadata` on each entry, and the scan has to land on the names-only side of
the two. Adding the `stat` takes it from 88 ms to 296 ms against a 123 ms midpoint. And the
retirement crash is the most destructive moment in the migration, because by the time the
mark is written the node has already told the network it serves those chunks from the file
store: the next start has to finish the deletion, never reopen the directory.

Measured rather than asserted, and printed in the log so drift shows before it trips a gate:
the startup scan of 100,000 chunks takes **102 ms** (1,019 ns/key), the in-memory index costs
**52 bytes per chunk**, opening the store reads **125 bytes** whatever the chunks contain, and
`put` writes exactly one directory entry per chunk.

Two things these deliberately do **not** prove, and are not offered as proving. Killing a
process keeps the kernel page cache, so the loopback jobs would stay green with every flush
removed from the publish path: real power-loss behaviour is still a fleet gate. And 100,000
keys is what a hosted runner can do in reasonable time, so where the curve stops being linear
at 1M and 10M is still a question about a machine, not about the code.

Each harness step fails if it runs no tests, and a separate check proves a shipped binary
carries no failpoint: the environment variable name is in the binary with `--features
test-utils` and absent without it.

What the tests actually prove, rather than merely exercise: publish is exactly-once under
sixteen concurrent writers of one address; the index rebuilds from the filesystem with a
stable order across restarts; a file in the wrong shard, an uppercase name and a non-regular
file are all refused; an interrupted write is swept; a corrupt file is removed and served
from the legacy copy instead, with the key re-queued; the copier is resumable and cannot
resurrect a pruned chunk; committing narrows the commitment without narrowing what is
served; a read in flight blocks retirement and retirement completes as soon as that read
finishes, **and the chunk stays readable throughout**; the shipped configuration retires with
nothing set by hand; a fully built node holding a legacy store is actually migrating it; retirement is refused for each gate independently; verification repairs a file
that rotted and refuses to retire when it cannot; a node with no view of the network gives up
nothing; the commitment-delivery counter resets on rotation; suffix sharding reaches all 256
directories for a clustered key set where prefix sharding collapses to one.

Adversarial review: twenty-eight rounds of codex at xhigh, plus three independent
reviewers. The last round found nothing at any severity and concluded the branch is ready
to begin a gated rollout. The findings that mattered were all on the destructive path, and
all of them were cases where a node could end up with no copy of a chunk it was supposed to
have:

The last four rounds found the same defect three more times, at successively smaller
scales: a filesystem answer that is neither yes nor no, folded into one of them, and acted
on later as though it had been established.

- Reading the retirement mark answered yes or no and folded every other outcome into no. A
  retired environment whose mark could not be read looked live, went back under its own
  name, and its keys re-entered a commitment they had already left. It is three states now,
  and the two questions callers ask are asked separately: deleting needs a mark that was
  read, opening needs one known to be absent, and neither takes "cannot tell" for a yes.
- Making it three states was only half of it. Two callers asked whether removal was
  permitted and let every other answer fall through to the opposite action, so "cannot tell"
  still reached the opening path.
- The mark is written with `create_new`, and a failure saying something is already at that
  name was taken as proof the mark was there. What is at that name might be anything. Every
  other part of this code insists the name is not the evidence; this was the one place
  taking it.
- The classification gated only the path where the node had lost its handle, so on the
  ordinary path it was never asked: a node holding its store open went through every gate,
  renamed the directory aside and deleted it. It is now the first question asked, before
  anything else.
- Start-up probed the mark twice. The answer can change between two probes, and a second
  answer of "cannot tell" after a first of "retired" dropped through to opening the very
  directory the first answer said not to open.

Two of the same shape in the LMDB store: a delete whose growth could not be measured was
charged nothing, so a copy-on-write delete could spend disk the growth budget never saw;
and sizing the map read every metadata failure as an empty database, which on a node with a
large one produces a map too small to open it.

And from the earlier rounds:

- Retirement could delete the legacy store while a read had thrown away a corrupt file and
  not yet reached the copy that would replace it. Reads and retirement are now a shared and
  exclusive guard rather than a race, verified by removing the guard and watching the test
  fail.
- A copy whose directory flush failed was still reported as successful on the next attempt,
  because a name already on disk returned early without flushing. Every successful return now
  flushes, and the same applies to creating a shard directory and to moving the legacy store
  aside before deleting it.
- A good copy offered to repair a damaged chunk was acknowledged and discarded, because both
  the protocol handler and the file store answered from names alone. Both compare the length
  first now.
- The possession threshold was computed from the peers that qualified rather than the whole
  group, so the last two holders could each conclude the other had it.
- The migration was spawned only on one branch of startup, so a node whose replication engine
  failed to start ran on forever with two stores. A test now builds a real node with a legacy
  store and asserts it is migrating it; deleting the spawn turns it red.
- Off Unix a rename cannot be shown to be durable, so a power loss could bring the legacy
  directory back with its contents already deleted and the node would fail to start on it.
  The removal is now recorded durably before anything moves, with the same create-and-flush
  that publishes a chunk. A start that finds that record opens the environment to decide:
  one that opens is intact and is kept, with the migration starting again from copying; only
  one that cannot be opened is treated as the remains of an interrupted removal.
- A name on disk is not proof of the bytes under it, and both the protocol handler and the
  file store answered "already have it" from names. A good copy offered to repair a damaged
  chunk was thanked and discarded, and nothing offered another. Both read and compare now,
  and repair from the offer.
- The deletion walker followed a symlink at the top level. An operator who points the chunk
  store at another volume leaves a link there, and retirement would have deleted the
  contents of a directory that is not this node's to delete. A linked environment is copied
  out of and never retired.
- A verification proof could go stale. The pass reads every chunk and its result is reused
  for half an hour, because retirement is usually deferred by a gate unrelated to the files.
  A kept chunk that stopped being readable in that window was invisible: ordinary requests
  were still served from the legacy copy. The store counts the times a chunk stops being
  servable, a proof records that count, and retirement refuses one the store has outrun.

Several of those were introduced by earlier rounds' own fixes, which is the argument for
running the loop rather than stopping at the first clean round. Holding a legacy handle across
every read closed a race but meant retirement would never see the environment unreferenced
on a busy node; taking the resulting guard on reads alone let writes and deletes starve the
drain instead; holding it through the deletion would have stalled every chunk request on the
node for as long as `remove_dir_all` took. The guard is shared for reads, writes and deletes,
exclusive for retirement, and released once the handle is out and the directory is renamed
away.

Four findings are accepted rather than fixed, each with its reasoning, in the ADR.

## New dependency

none

## ADR

https://github.com/grumbach/ant-node/blob/storage/file-chunk-store-and-lmdb-retirement/docs/adr/ADR-0013-file-based-chunk-store-and-lmdb-retirement.md

## What the review changed, and what it left

The last five rounds found no blockers. What they did find, repeatedly, was one shape of
defect: **a fact established at one moment being acted on at another.** Copying, verifying
and retiring are hours apart by design, and the physical work runs on blocking threads that
outlive the futures that started them, so a caller that goes away does not undo what it
started. The answer that worked was to make each belief carry its own expiry rather than to
re-check and hope the check lands close enough to the act: the retired directory carries its
own mark, a verification proof carries the store's health count and is refused once the store
outruns it, a write announces itself before it starts and is cleared by the worker rather
than the caller, and a delete waits out whatever is already writing its key.

Two decisions were taken against a reviewer's first suggestion and are worth stating:

- **The suspect state is not persisted across restart.** A restart re-advertises a chunk
  until something reads it, and the first read settles it. Persisting means another file to
  keep in sync with its own failure modes. The pre-retirement pass reads every chunk before
  anything is deleted, which closes the window during the migration itself.
- **There is no per-key storage actor.** It was proposed as the structural end to the defect
  class above. Its key property, that the worker owns the bookkeeping rather than the caller,
  is applied where the defects actually land. The final round agreed the rewrite is
  unnecessary.

## Mitigation / rollback

`storage.migration.enabled = false` stops the copier at any point; the node keeps reading
both stores and simply never frees the old one's disk. `ANT_MIGRATION_RETIRE_LEGACY=0` leaves
the copier running but holds the removal off, per node, without a separate build.

Before a node removes its legacy store, reverting is a higher-semver build of the previous
code, and every chunk written meanwhile is in the legacy store too. **After** a node has
removed it, there is no rollback for that node. That is what the staged rollout, the 72-hour
shed hold, the wave stagger and the four-hour retention delay are all for: at any point the
fleet is a mix of nodes that have removed it and nodes that have not, and the ones that have
not are the rollback.

---

## The problem

LMDB returns a deleted page to its own free list and never to the filesystem. Last week the
fleet deleted 2.29 million chunks and got back zero bytes. Operators read that as a bug and
are tempted to wipe node directories, which costs the network real replicas. Compaction does
not help: it needs free space equal to the live data, which is exactly what a full node does
not have, and it leaves us on LMDB. Disk comes back exactly once, when `chunks.mdb` goes.

## Why the last two hex characters and not the first

A node holds keys it is among the closest to, so its holdings share roughly
`log2(N / close_group_size)` leading bits with its own node ID, and that prefix **grows** as
the network grows.

| nodes | shared bits | 2 hex | 3 hex | 4 hex |
|---:|---:|---:|---:|---:|
| 1,000 | 7.2 | 1.8 | 29 | 459 |
| 10,000 | 10.5 | 1 | 2.9 | 46 |
| 100,000 | 13.8 | 1 | 1 | 4.6 |
| 1,000,000 | 17.1 | 1 | 1 | 1 |

At today's fleet a two-hex prefix is already about two directories. Prefix sharding does not
degrade, it fails, and it fails later for the nodes that grow into it. The address is a
BLAKE3 output and close-group membership constrains only its leading bits, so the trailing
byte is uniform by construction at any size. IPFS shipped the same fix for a different reason
and their `_README` still says so.

## Why no index database

A persistent index cannot remove reconciliation. Commit the index first and a crash leaves a
phantom key; rename the file first and a crash leaves an unindexed file. Repairing either
means reading the filesystem anyway, so the filesystem may as well be the authority, and then
nothing can drift. Ceph FileStore's tracker #17177 is what the other choice looks like.

## How a full node is handled

A node that fits its payload copies everything and then removes the old store. It is never
unable to serve.

A node that does not fit copies closest-first, commits to what it can hold while still
serving everything it ever committed to, and only then gives the rest up. Nothing is given up
until three things hold, in this order:

1. the node is not near the front of the group for that chunk;
2. its close group has **received** the reduced commitment, proven by those peers answering a
   neighbour sync that carried it, because until then they audit it against the set it used
   to hold;
3. all but one of the chunk's close group has answered a **cryptographic possession
   challenge** over a nonce it has never seen, and is itself currently publishing a
   commitment.

Point 3 is the pruner's own evidence. The cheap `VerificationRequest` is deliberately not
used: it carries a self-reported `present` flag, and a node that silently lost a chunk still
answers yes.

Close groups migrate in **waves** derived from a hash of each node's ID, about two of seven
at a time. If every holder went at once, none could prove to the others that a copy survived
and the group would deadlock. A host-wide lock separately serialises nodes sharing a volume,
which is a different question: one machine's disk rather than one chunk's replicas.

Before the old store goes, every chunk both stores hold is re-hashed and rewritten from the
legacy copy if they disagree, because a filename is not proof the bytes behind it are good.

## What this costs, plainly

- No rollback once a node has removed its legacy store. The staged rollout is the only
  control. `ANT_MIGRATION_RETIRE_LEGACY=0` holds a node before the removal if one is needed
  to keep both copies.
- Rollback is no longer lossless *before* removal either, for a chunk whose legacy copy
  could not be written. The environment is pinned at its current size for the whole bridge,
  so a write it cannot satisfy from its own free list is refused and the chunk lives in
  files alone. That is deliberate: two stores measuring one disk can otherwise spend the
  same free space twice and fill the volume this exists to free. A node that must keep an
  old-binary rollback available should have retirement held off.
- A node whose close group is **also** short of disk will not get proofs, will not free its
  disk, and will tell its operator to add storage. That is the right answer under "no data
  loss", but the migration will not complete unattended everywhere. How large that population
  is, is the main thing to measure once this is out.
- Narrowing the commitment lowers the quoted price, which is quadratic in the committed key
  count, so a node that has just proved it is short of disk quotes cheaper and then refuses
  the store on capacity. A wasted round trip, not a mispayment. The fix belongs to the quote
  path.
- Inodes are not measured, only bytes.
- The paid list is still LMDB, which is why `heed` cannot be dropped yet.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -223,7 +223,7 @@ jobs:
         run: |
           set -euo pipefail
           cargo build --bin ant-node
-          found=$(strings -a target/debug/ant-node | grep -c 'ANT_HALT_BEFORE_PUBLISH' || true)
+          found=$(strings -a target/debug/ant-node | grep -c 'ANT_HALT_' || true)
           # With --features test-utils this count is 1, which is what makes a count of 0
           # here evidence rather than an accident of how the binary was stripped.
           if [ "$found" != "0" ]; then
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,10 +65,16 @@ jobs:
       - name: Several nodes migrating on one disk
         run: cargo test --test migration_shared_volume
       - name: Startup scan, index memory and inode cost at scale
-        run: cargo test --test storage_scale -- --nocapture
+        run: cargo test --test storage_scale --features test-utils -- --nocapture
 
+  # Runs the storage tests against real ext4, XFS and btrfs rather than whatever the
+  # runner provides. Deliberately NOT named durability: killing a process and reopening
+  # the same mounted filesystem keeps the page cache, so this exercises each filesystem's
+  # syscall, locking, rename and delete behaviour, not its behaviour under power loss.
+  # That still needs block-device fault injection or a real machine, and remains a fleet
+  # gate.
   filesystems:
-    name: Durability on ${{ matrix.fs }}
+    name: Storage on ${{ matrix.fs }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
@@ -87,17 +93,21 @@ jobs:
           # on whatever the runner happens to give us. ext4 is what most of the fleet is
           # on; XFS and btrfs are the two the design reasons about separately, btrfs
           # because it has been observed reordering writes around a rename.
+          # 3 GiB is ample: these tests use tens of MiB. The scale harness, which is
+          # the one that needs room, is not in this job.
           truncate -s 3G /tmp/${{ matrix.fs }}.img
           mkfs.${{ matrix.fs }} -q /tmp/${{ matrix.fs }}.img
           sudo mkdir -p /mnt/antfs
           sudo mount -o loop /tmp/${{ matrix.fs }}.img /mnt/antfs
           sudo chown "$USER" /mnt/antfs
           df -hT /mnt/antfs
-      - name: Prove the migration returns disk on ${{ matrix.fs }}
+      # TMPDIR is what `TempDir::new` uses, so this is what puts the test data on the
+      # mounted filesystem rather than on the runner's root.
+      - name: The migration returns disk on ${{ matrix.fs }}
         env:
           TMPDIR: /mnt/antfs
         run: cargo test --test migration_reclaims_disk --features test-utils
-      - name: Kill a node mid-migration on ${{ matrix.fs }}
+      - name: A node killed mid-write on ${{ matrix.fs }} loses nothing
         env:
           TMPDIR: /mnt/antfs
         run: cargo test --test migration_crash_safety --features test-utils -- --test-threads=1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -224,8 +224,10 @@ jobs:
           set -euo pipefail
           cargo build --bin ant-node
           found=$(strings -a target/debug/ant-node | grep -c 'ANT_HALT_' || true)
-          # With --features test-utils this count is 1, which is what makes a count of 0
-          # here evidence rather than an accident of how the binary was stripped.
+          # With --features test-utils this count is not zero, which is what makes a zero
+          # here evidence rather than an accident of how the binary was stripped. The exact
+          # number is one per failpoint and is deliberately not asserted, so that adding a
+          # failpoint does not fail this check.
           if [ "$found" != "0" ]; then
             echo "the publish failpoint is compiled into a default-feature build"
             exit 1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,6 +154,25 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - name: Build release (no logging)
         run: cargo build --release --no-default-features
+      # The crash harness drives the store through a failpoint that parks the process
+      # forever on an environment variable. It is compiled only under `test-utils`, which
+      # is not a default feature and is not passed by the release workflow, so a shipped
+      # binary does not contain it. This proves that rather than trusting it: the variable
+      # name is a string literal, so it survives into the binary whenever the code that
+      # reads it is compiled, and its absence is the absence of the failpoint.
+      - name: A shipped binary carries no failpoint
+        if: runner.os == 'Linux'
+        run: |
+          set -euo pipefail
+          cargo build --bin ant-node
+          found=$(strings -a target/debug/ant-node | grep -c 'ANT_HALT_BEFORE_PUBLISH' || true)
+          # With --features test-utils this count is 1, which is what makes a count of 0
+          # here evidence rather than an accident of how the binary was stripped.
+          if [ "$found" != "0" ]; then
+            echo "the publish failpoint is compiled into a default-feature build"
+            exit 1
+          fi
+          echo "no failpoint in a default-feature build"
 
   test-no-logging:
     name: Test (no logging)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,25 @@ jobs:
           version: ${{ env.FOUNDRY_VERSION }}
       - name: Run unit tests
         run: cargo test --lib --features test-utils
+      # Before the e2e suite, deliberately. These are fast and deterministic, and the e2e
+      # suite flakes on hosted runners for transport reasons that have nothing to do with
+      # storage. A failing step aborts the job, so anything sequenced after a flaky one
+      # never reports, which is how these ran on no platform at all for a whole run.
+      - name: Prove the migration returns disk to the filesystem
+        run: cargo test --test migration_reclaims_disk --features test-utils -- --test-threads=1
+      - name: Kill a node mid-migration and check what survived
+        run: cargo test --test migration_crash_safety --features test-utils -- --test-threads=1
+      - name: Several nodes migrating on one disk
+        run: cargo test --test migration_shared_volume
+      # Linux only. This one plants a hundred thousand files to measure what a restart
+      # costs, and the answer it is after is a fleet answer, where every node is Linux.
+      # The scan itself reads names and nothing else, which is not a platform-specific
+      # path, and opening a store is covered on all three by the unit tests. Planting that
+      # many files on the Windows runner would cost minutes of every run to re-measure
+      # something no node will ever do there.
+      - name: Startup scan, index memory and inode cost at scale
+        if: runner.os == 'Linux'
+        run: cargo test --test storage_scale --features test-utils -- --nocapture --test-threads=1
       - name: Run e2e tests
         run: cargo test --test e2e --features test-utils -- --test-threads=1
       - name: Run v12 storage-bound audit attack PoCs
@@ -60,14 +79,6 @@ jobs:
         run: cargo test --test poc_bootstrap_stall --features test-utils
       - name: Shutdown waits for writes whose caller has gone
         run: cargo test --test poc_shutdown_lmdb_drain --features test-utils
-      - name: Prove the migration returns disk to the filesystem
-        run: cargo test --test migration_reclaims_disk --features test-utils -- --test-threads=1
-      - name: Kill a node mid-migration and check what survived
-        run: cargo test --test migration_crash_safety --features test-utils -- --test-threads=1
-      - name: Several nodes migrating on one disk
-        run: cargo test --test migration_shared_volume
-      - name: Startup scan, index memory and inode cost at scale
-        run: cargo test --test storage_scale --features test-utils -- --nocapture --test-threads=1
 
   # Runs the storage tests against real ext4, XFS and btrfs rather than whatever the
   # runner provides. Deliberately NOT named durability: killing a process and reopening
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,14 +58,16 @@ jobs:
         run: cargo test --test poc_audit_handler_live --features test-utils
       - name: Run bootstrap-stall PoC regression marker
         run: cargo test --test poc_bootstrap_stall --features test-utils
+      - name: Shutdown waits for writes whose caller has gone
+        run: cargo test --test poc_shutdown_lmdb_drain --features test-utils
       - name: Prove the migration returns disk to the filesystem
-        run: cargo test --test migration_reclaims_disk --features test-utils
+        run: cargo test --test migration_reclaims_disk --features test-utils -- --test-threads=1
       - name: Kill a node mid-migration and check what survived
         run: cargo test --test migration_crash_safety --features test-utils -- --test-threads=1
       - name: Several nodes migrating on one disk
         run: cargo test --test migration_shared_volume
       - name: Startup scan, index memory and inode cost at scale
-        run: cargo test --test storage_scale --features test-utils -- --nocapture
+        run: cargo test --test storage_scale --features test-utils -- --nocapture --test-threads=1
 
   # Runs the storage tests against real ext4, XFS and btrfs rather than whatever the
   # runner provides. Deliberately NOT named durability: killing a process and reopening
@@ -106,7 +108,7 @@ jobs:
       - name: The migration returns disk on ${{ matrix.fs }}
         env:
           TMPDIR: /mnt/antfs
-        run: cargo test --test migration_reclaims_disk --features test-utils
+        run: cargo test --test migration_reclaims_disk --features test-utils -- --test-threads=1
       - name: A node killed mid-write on ${{ matrix.fs }} loses nothing
         env:
           TMPDIR: /mnt/antfs
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,11 +55,35 @@ jobs:
       # storage. A failing step aborts the job, so anything sequenced after a flaky one
       # never reports, which is how these ran on no platform at all for a whole run.
       - name: Prove the migration returns disk to the filesystem
-        run: cargo test --test migration_reclaims_disk --features test-utils -- --test-threads=1
+        shell: bash
+        run: |
+          set -euo pipefail
+          cargo test --test migration_reclaims_disk --features test-utils -- --nocapture --test-threads=1 2>&1 | tee /tmp/reclaims_disk.log
+          # A target whose required features are not passed is skipped with a
+          # warning and a zero exit, so a harness can stop running without anyone
+          # noticing. This is what makes that loud.
+          grep -qE 'test result: ok\. [1-9]' /tmp/reclaims_disk.log \
+            || { echo 'reclaims_disk ran no tests'; exit 1; }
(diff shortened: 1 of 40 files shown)


dirvine (approved):
Reviewed exact head `aaadeb242b51895089de1ce634919376b31d20cc` (new scope against stacked PR #215).

I found no merge-blocking correctness issue. The migration is fail-conservative around copying, verification and destructive retirement, and the production defaults retain the full gate set.

Local validation passed:
- 1,053 library tests
- crash-safety harness (5 passed, 3 helper tests ignored)
- disk-reclaim harness (2 passed)
- shared-volume harness (5 passed)
- 100k-chunk scale harness (2 passed)
- `cargo check --lib --no-default-features`
- clippy with `-D warnings` and rustfmt check

Non-blocking watch-out: the per-volume lock is process/host heuristic rather than a safety gate; isolated container `/tmp` namespaces can defeat serialisation. The capacity floor and retirement proofs remain conservative, but container operators sharing a data volume may see slower or stalled migration. Making the lock directory operator-configurable would be a useful follow-up.

This approval is for the code merge, not a live rollout go-ahead. ADR-0013’s fleet gates (baseline, canary, mid-rollout checkpoint and post-retirement observation) still need staged evidence before broad deployment.

dirvine (approved):
Re-reviewed exact head `b0933ee3ae312f354c02489cc935bb54b9e25592`, incremental from the previously approved `aaadeb242b51895089de1ce634919376b31d20cc`.

**APPROVE — no merge-blocking findings.**

The prior shared-`/tmp` volume-lock caveat is resolved by `211ff96`: the shipped multi-node systemd deployments now give every node on a host a shared, node-user-owned `/var/lib/ant/migration` lock directory while preserving per-node data isolation.

The incremental review also covered retirement-mark classification/durability, stale verification invalidation, reader/writer retirement exclusion, pending-write accounting, symlink-safe deletion, capacity accounting, commitment/charge correctness, the ADR-0014 renumber/governance check, Windows behaviour, and crash/cleanup recovery. Three independent focused reviewers returned APPROVE; no dissent.

Validation at this head:
- 1,062 library tests passed
- crash-safety harness: 5 passed, 3 helper tests ignored
- all GitHub checks green, including Linux/macOS/Windows, ext4/XFS/btrfs, clippy, format, docs, audit and ADR governance

As before, this approves the code merge; broad live rollout remains subject to ADR-0014’s staged fleet evidence and human rollout sign-off.

jacderida:
## Testnet run result — 990 nodes, 2026-08-28

Ran this branch on a 990-node testnet to exercise the destructive path (V2-1101). **The migration works: it completed end-to-end on live nodes under continuous upload/download load, at production node density, and returned disk.** Full criteria table in the Linear issue; below is what matters for the PR.

Build was `jacderida:testnet/pr216-with-main` @ `49bc578` (this PR + current `main`, reporting `ant-node 0.17.2`). Fleet deployed on the 0.17.2 release, filled to ~376 MiB/node of `chunks.mdb` (~90–190 chunks at ~4 MiB), then swapped onto the branch in place — the fleet was 660 nodes at 15/VM plus 330 at 5/VM.

### What was verified

- **990/990** nodes emitted `migration_event="start"`. Zero `not_started`, zero "Storage migration is disabled".
- **Every** volume's lock holder reached `Committed` — exactly one per VM, 44/44 and 66/66, within **~16 minutes** of the swap. The `/tmp/ant-migration-<devid>.lock` serialisation was flawless: zero VMs with more than one holder, zero "migrate without serialising" warnings, zero 6h-backstop trips.
- **49 nodes across all 44 large VMs** went `complete` → `space_returned`, **1:1** with no orphans either way. `chunks.mdb` deleted on all 49, **zero** `.retired` remnants left behind. **19.64 GiB** returned, mean **410 MiB**/node.
- **Zero** `needs_an_operator`, **zero** "cannot be repaired", **zero** panics.
- Restarting 10 lock holders in `Committed` preserved every marker: `first_start_unix` and `committed_at_unix` unchanged, no "implausible first-start", and the resume line correctly read **`Storage migration starting: 0 chunk(s) still only in the legacy environment`** rather than restarting from the original backlog. Chunk counts never regressed.
- Uploads were **100%** successful throughout (1244/1244 in the measurement window). File store ended holding **1,075,457** chunks.

### Three things worth your attention

**1. `no_rollback_copy` fired on 100% of nodes, and bridging is one-directional much sooner than the design implies.**

Within six minutes of the swap, all 660 nodes had logged it, from `chunk_store.rs:522`:

> Could not also write `<addr>` to the legacy environment: storage error: Insufficient disk space: 0.49 GiB reserve required and no reusable page

Measured at that same moment, those volumes held **86–92 GiB free**. A 0.49 GiB reserve cannot bind with 84 GiB free, so this is LMDB map-size pinning ("no reusable page"), not disk exhaustion — which is the benign explanation, and no chunk is at risk since the file store has it and reads serve the union.

But the consequence is stronger than "a rollback would miss some chunks". I hit it concretely: I needed to reset one arm's migration (drop the marker and the copied `chunks/`) and re-run it, and **that would have been real data loss** — 1,190 chunks on that arm existed in the file store *only*. So within minutes of the swap, under ordinary write load, the file store holds chunks that exist nowhere else on the node. Bridging is presented as reversible and in practice it isn't, on any node that is taking writes. Worth saying so explicitly in ADR-0014, or making the dual-write failure louder than a warning.

**2. The shed path has a narrow band — being *badly* short of disk gets you no relief at all.**

I could not get a single successful shed out of the run. Squeezing lock holders to 60%, then 15%, then 5% shortfall all produced the same refusal:

> This node is 438 chunk(s) short of disk, and at least 4 of them are chunks it is near the front of the group for (for example 38010922…). It will not…

The guard behaves exactly as specified — decisive, names the offending keys, deletes nothing, which is the right failure mode. But `SHEDDABLE_TAIL_RANKS = 2` of a 7-node group means only ~29% of a node's keys are ever sheddable, and `keys_this_node_must_not_give_up` requires **every** key being given up to be tail-rank. So a shed only happens when the shortfall is under ~29%: a node slightly short of disk sheds, a node badly short refuses and keeps both stores indefinitely, with "add disk or wait for replication" as the only exit. If that inversion is intended it should be stated in the ADR; if not, the selection may want to shed only what it *may* shed rather than refusing the whole set.

(Caveat: those nodes were ~30 min old, so an immature routing table may contribute to the conservative rank verdict. I can't separate that from the guard's inherent strictness with the data I have.)

**3. Per-volume pace, quantified — the number the issue asked for.**

The driver holds the volume lock from first copy through retirement, including the ≥4h delay. Observed directly: on every multi-node VM, exactly one node migrated and the rest sat in `Bridging` for the whole run. So a node cannot even be *considered* for shedding until every node ahead of it on its volume has copied, committed, waited ≥4h and retired — **≥48h on a 12-node production volume before the first shed decision is reachable**, before the 72h hold and 24h waves are counted at all. This is shipped behaviour and the run tested it as shipped; flagging it for your call on whether the pace is acceptable.

### One measurement note, in case it saves you time later

The acceptance criteria asked for per-VM `df` delta to match Σ`freed_bytes` within ±10%. That is not measurable on a fleet under write load: across 44 VMs the `df` movement was **−3.9 TiB** while retirement returned **19.6 GiB** — the signal is **0.5%** of the noise. Disk return is verifiable instead by `chunks.mdb` absence, zero remnants, and the nodes' own `freed_bytes`, which is what I used.

Also: the branch build reports the same `0.17.2` as the release it replaced, so "did the swap land" has to be checked by **sha256**, not version string.

---

Nothing here argues against the change — the destructive path did the right thing every time it ran. Items 1 and 2 are design questions rather than defects. Tracked in V2-1101; happy to re-run with a per-node-payload squeeze at full 1/3-of-fleet scale if you want the possession-gate-under-contention question answered properly, which this run did not settle.

jacderida (approved):

PR #218, opened 2026-08-28
refactor(storage)!: remove the LMDB chunk store, and never refuse a start over what it left behind

grumbach:
> **Stacked on #223.** That PR is merged, but into the release branch `rc-2026.9.2`, not into
> `main`, so until that branch is back-merged this PR's diff carries #223's single commit
> (`264014a`) alongside its own. It drops out on its own once the back-merge lands.
>
> The branch is based on `264014a` rather than on `rc-2026.9.2` deliberately: `264014a`'s parent
> is exactly `main`, so this carries no version bump, and `264014a` becomes a common ancestor of
> both sides of that back-merge whichever order the two land in. **That only holds if this PR is
> merged with a merge commit.** A squash or a rebase-merge drops `264014a` from the ancestry and
> the back-merge then re-derives its base from `31fcbae`, re-opening every #223 conflict.
>
> **Reviewing the store.** `file_store.rs` folded into `chunk_store.rs`, and `git diff -M` does
> *not* pick that up as a rename, because `chunk_store.rs` already existed — it was the facade
> over the two stores, and its contents were replaced rather than moved. So the raw diff shows
> ~3.7k deleted and ~3.9k added where the real change is 575 lines. What to read instead:
>
> ```
> git diff 264014a:src/storage/file_store.rs HEAD:src/storage/chunk_store.rs   # +408 -167
> ```
>
> That is the store's actual change. Everything else under `src/storage/` is a deletion.

## Linear issue

Closes V2-1033

## Risk tier

## Compatibility

- Wire: none. No message shape, field, or protocol version change. The migration signal #223
  put in the user agent is untouched and still sent.
- Storage: **changes.** `chunks.mdb` can no longer be read at all. **Nothing here refuses a
  start** — an earlier draft refused on an unmigrated store and that design was rejected: a node
  that refuses serves nothing, and it cannot be held on the previous release, because
  `build_upgrade_monitor` is unconditional and `UpgradeConfig` has no field that disables it. A
  leftover is removed **only** when it carries the `RETIRED` mark the previous release wrote
  inside it, or is empty; both are re-checked on the deleting thread immediately before it
  unlinks. Anything else is kept and named once under `migration_event = "legacy_store_left"`.
  Links are neither followed nor unlinked. Kept is not the same as available: there is no LMDB
  reader in this build, so a node that keeps a directory keeps its bytes and can serve none of
  them.
- Storage, second axis — **values over the 4 MB ceiling are not chunks, and none is preserved.**
  Every released ingress enforces `MAX_CHUNK_SIZE` before storing anything (the protocol handler
  on a paid store; replication on receive and on fetch), so no over-ceiling value ever entered
  this network as a chunk. The one way one could reach a disk was the bridge's own `put`, which
  wrote to LMDB first — that store has no size ceiling — and only then offered the same bytes to
  the file store, which refused them; the key was then recorded legacy-only and the copier's size
  arm deleted it. **This PR removes the bridge, so it removes that hole**: one store, whose `put`
  refuses over the ceiling *before* it writes, and no `LmdbStorage::put` left to take an unbounded
  value at all. No sidecar is built and nothing is preserved — there is no valid chunk there to be
  the only copy of. A legacy directory containing such a value is kept or removed on its mark like
  any other; this build never looks inside one.
- API: **breaking.** Removed from `ant_node::storage`, in full because the list matters to
  anyone compiling against this crate: `LmdbStorage`, `LmdbStorageConfig`, `MigrationConfig`,
  `MigrationPhase`, `MigrationState`, the `storage::migration` module, `FileStore`,
  `FileStoreConfig`, `StoreLayout`, `VerifyReport` and `LEGACY_ENV_DIR`. The `FileStore` family
  and `LEGACY_ENV_DIR` are easy to miss because the type survives under another name; they were
  public and they are not any more. `ChunkStore` and `ChunkStoreConfig` keep their
  names and their methods; they now name the file store directly rather than a facade over two
  stores. `storage.migration` and `storage.db_size_gb` are gone from the config file. **An
  operator's existing config still loads with both keys present**, because nothing declares
  `deny_unknown_fields`, and there is a test holding that true.

## Semver impact

- [x] breaking
- [ ] feature
- [ ] fix

## Test evidence

```
cargo test --lib --features test-utils                    1025 passed, 0 failed
RUSTFLAGS=-D warnings cargo test --no-default-features --lib
                                                          1025 passed, 0 failed
cargo test --test e2e --features test-utils               97 passed, 0 failed, 3 ignored (503 s)
                                                          and, on the last run, 96/1: one
                                                          transport flake, `Failed to create
                                                          transport` at node creation before
                                                          any test body ran, in a merkle-payment
                                                          test that touches no storage path.
                                                          Passed alone on re-run. Recorded
                                                          rather than rounded up; CI is the
                                                          check that counts.
cargo test --test chunk_store_crash_safety                   2 passed, 1 child ignored
cargo test --test storage_scale                              2 passed
cargo test --test poc_commitment_audit_attacks              19 passed
cargo test --test poc_audit_handler_live                    16 passed
cargo test --test poc_bootstrap_stall                        3 passed
cargo clippy --all-targets --all-features -- -D warnings  clean (the CI invocation)
cargo clippy -- -D warnings                               clean (stricter than CI; it is what
                                                          caught a redundant pub(crate) that
                                                          --all-features hides)
cargo fmt --all                                           clean
RUSTDOCFLAGS=--deny=warnings cargo doc --no-deps          clean
python3 scripts/adr-governance.py                         passed, 15 ADRs checked
```

Three properties are newly pinned, none of which had a test before:

- **the cleanup removes exactly what the fleet signal calls finished with**, asserted in both
  directions over every shape a root can present. This is the correspondence that lets the
  previous release's count authorise this one, and it is what fails if anyone gives the cleanup
  its own classifier again;
- **all sixty-six leftovers** a root can hold — the live directory, the unnumbered tombstone and
  sixty-four numbered ones — are removed by one start, on one thread rather than sixty-six;
- **a root that cannot be listed has nothing removed from it**, which is the case that used to
  return silently while the decision record promised a warning;
- **a value over the ceiling is refused before anything is written** — not claimed, no file left
  under its name, and nothing for a restart's scan to index. Addressed to its own bytes so the
  refusal is the size arm and not the content-address arm, and mutation-checked: with the size
  branch deleted the test fails.

Three adversarial review rounds with `codex xhigh` — the release graph and design before any
editing, the exact diff afterwards, and a third against the deleter written to answer the human
review. Logs in `ugly_lmdb_removal/r3/`. The third round found two blockers in that new code: the
mark's type test accepted a symlink or a socket (`S_IFLNK` and `S_IFSOCK` each contain every bit
of `S_IFREG`, so it has to mask with `S_IFMT`), and `O_NOFOLLOW` does not refuse a mount point,
so a subdirectory inside a leftover is now refused rather than entered. Both have tests, both
mutation-checked.

## New dependency

**One, added in review: `rustix` (unix targets only, `fs` feature).** It supplies safe
`openat`/`unlinkat`, so the cleanup deletes relative to a directory handle it opened
`O_NOFOLLOW` instead of re-resolving a path it has already checked — see Mitigation below.
The alternative was hand-written `unsafe` around the same libc calls, in a code path whose job
is deleting things, which seemed the worse trade.

**It adds nothing to the build.** `tempfile`, already a direct dependency, pulls in this exact
crate and version with the `fs` feature on; `cargo tree` shows one `rustix v1.1.4` before and
after, and `Cargo.lock` gains no entry. Flagging it anyway because the template asks for
explicit acknowledgement of new dependencies, and this is a new line in `Cargo.toml`.

`heed` stays in the dependency list: the paid-key list has its own LMDB environment, which this
change does not touch.

## ADR

https://github.com/grumbach/ant-node/blob/storage/lmdb-removal-release-3/docs/adr/ADR-0015-remove-the-lmdb-chunk-store.md

## Mitigation / rollback

The blast radius is bounded at the point of deletion rather than after it. A directory is removed
only when it provably holds no chunks — it carries the previous release's `RETIRED` mark, or it is
empty — and both conditions are re-checked on the deleting thread, against the path it is about to
unlink, immediately before it does. Everything else is kept, and the node runs and serves what it
did migrate.

**Operator-facing, and it belongs in the release notes rather than only in the ADR:** the mark
is believed and never re-verified, because this build cannot read the environment to compare.
An operator who restores a backup of `chunks.mdb` **must remove the `RETIRED` file from it
first**. Left in place, the next start reads it as the previous release's record that the
directory was finished with, and deletes it. No release produces that state; a restore can.

The close-group unheld-chunk penalty stays suspended for one more release, so the emergency lever
for the accusation half is still `ANT_SUSPEND_UNHELD_CHUNK_PENALTY`. The removal half has no lever:
the on-disk format rolls back cleanly, but the operation does not, because a node put back on the
previous binary is dragged forward again within the hour by an upgrade monitor that cannot be
switched off. Adding a disable or a version ceiling is an upgrade-subsystem change and is
deliberately not in the release that also deletes a store; it is named as an open gap in the ADR.

**Fleet gate, not satisfied, and not close.** This must not be published until the fleet is
demonstrably on the file store, because it is the release that stops any node going back. Three
separate reasons the count cannot answer that yet:

- the signal first shipped in `v0.19.0-beta.1`, published 2026-09-09T23:53Z, so there is under a
  day of it against a gate that asks for several consecutive days;
- **the staged migrations have not started**, so a node reporting `files` today is reporting that
  it has nothing to move, not that it has finished moving it — and the tally cannot tell those two
  apart. Current beta lines are not progress towards this gate and should not be read as any;
- the population the gate exists for is not reporting to us: the community nodes, and the NTFS
  hosts where the store's case-folding and directory-entry behaviour differ from ours, are
  unavailable.

When the gate is met, and any devnet or canary run before it, are the release manager's call and
not this PR's.


dirvine (changes requested):
## Review — LMDB chunk store removal (final migration release)

Six-seat independent review against the PR head (`94e247b`) plus the previous release's store for provenance. CI green across macos/ubuntu/windows and ext4/xfs/btrfs; ADR-governance, clippy, fmt, doc, security-audit all pass. Data-loss, concurrency, API/semver, and replication/audit lanes are sound. One security finding to address before merge; release-gating handled separately (below).

### MAJOR — path-based deletion in `delete_mark_last` enables symlink-swap escalation outside the data dir (`src/storage/legacy_artifacts.rs:223–265`)
`delete_mark_last` re-checks `symlink_metadata(dir)` is a real directory at line 229, but then re-opens the same *path* with `read_dir(dir)` at lines 240 and 247. `read_dir` follows a symlink. A local actor with write access to the node's data dir who wins the window between the metadata check and the `read_dir` can swap the marked directory for a symlink to an arbitrary target, redirecting the node's `remove_file` / `remove_dir_all` (lines 252–255) at the **symlink target's** contents — files the attacker cannot reach but the node process can (e.g. the root-run node under the terraform unit's `ProtectSystem=strict`). The ADR's accepted-risk paragraph (":214–222: whoever can win that race already has write access to this node's data directory") only covers deleting things *inside* the data dir; it does not address this escalation-outward variant. This is real privilege escalation, not just data-dir data loss.

The fix is available in this same codebase: `open_regular` already uses `libc::O_NOFOLLOW`, "checked on the handle, not the path" (`src/storage/chunk_store.rs:2479–2487`). The same pattern — open the directory with `O_NOFOLLOW | O_DIRECTORY`, then `unlinkat`/`fdopendir` relative to that handle — closes the window. Please either apply that (unix path; keep the existing portable fallback) or amend the ADR's accepted-risk paragraph to explicitly name and justify the external-target escalation. Not blocking the code otherwise.

### Release gate — do not publish until the fleet gate in ADR-0015 is actually met
Consistent across the PR body, the ADR and the code: the migration signal shipped under a day ago (2026-09-09), staged migrations haven't started, and community/NTFS hosts aren't reporting. ADR-0015 is explicit that no pre-staged-migration count is progress and "this release is not published on the strength of one." Given the no-rollback asymmetry (upgrade monitor drags nodes forward; marked directories are deleted finally), holding publication is the only defensible call. Merging the code is fine; **publishing is not yet safe.** This is an operational gate, not a code defect.

**Fleet-ops MAJOR (roster gap):** the positive half of the gate — "roster member finished" — is not implementable from what the code emits. Per-peer `migration_event="peer_state"` lines name only peers that are *not* `Files` (`src/storage/migration_signal.rs:368–379`); `Files` peers appear only as an aggregate count (`:464–470`), so a roster node that finished is indistinguishable, in the logs, from one that never connected. The negative half (no outstanding peers) is fully implementable. Recommend a low-rate named `peer_state` line for `files` peers (or a `peers_seen` list) before the gate is claimed met, or satisfy the roster check out-of-band.

### Minor / follow-ups (none blocking)
- `config/production.toml` still ships `[upgrade] enabled = false`, but `UpgradeConfig` has no `enabled` field — stale config telling operators they can disable upgrades when they can't. Remove or annotate.
- `pub struct StoreLayout` sits in a `pub(crate)` module under non-test cfg — a future cfg flip would silently re-publish it. Consider `pub(crate)`.
- ADR-0015 "values over 4 MB never existed" is overstated — they could exist via the bridge's ceiling-free `LmdbStorage::put`. Conclusion holds (they are not chunks, safe to delete), but reword to "can exist but are not chunks."
- The RETIRED-in-a-restored-backup data-loss path is accepted in the ADR; make it prominent in release notes / operator docs, not buried in the ADR.
- NTFS: tombstone matching is case-sensitive-by-string but case-folded-by-filesystem; one sentence in the ADR noting this would close the gap.
- Version lineage: Cargo.toml is `0.18.1` while ADR-0015 cites `v0.19.0-beta.1`. Bump to 0.19.x, call the break out in release notes.
- Concurrency seat flagged an ordering point worth the payments seat's eye: in the client-PUT handler, `holds_verified` runs (step 3) before payment verification (step 5), and its `Wrong` branch can write/replace a file on disk before any payment gate. Not a data-integrity race, but an unauthenticated remote reaching the write path.

### Verdict
**No data-loss, concurrency, API, or replication blockers.** The only must-address item is the `delete_mark_last` TOCTOU (MAJOR). Everything else is minor or release-timing. The author's claims check out against the code; the "no wire/protocol change" and "replication routes through ChunkStore" claims verified. Requesting changes on the security finding; otherwise this is approve-quality once the gate elapses.

dirvine (approved):
## Re-review — fix verified, approving

Reviewed the two fix commits (`efd9289`, `cd5e91d`) at head `cd5e91d` against my prior findings. CI fully green (all platforms + storage fs matrix); the 21 `legacy_artifacts` tests pass locally.

**MAJOR (delete_mark_last privilege escalation) — resolved, and resolved well.**
`delete_mark_last` now opens the directory once with `O_NOFOLLOW | O_DIRECTORY` and makes every unlink against that handle (`unlinkat`/`openat` via `rustix`, unix only). A link can never produce a handle, so no unlink can be redirected. The follow-up details are the right ones: the mark's type test is masked with `S_IFMT` (the prior `S_IFREG` test wrongly accepted a symlink/socket — a data-deletion-direction bug, now caught and pinned with per-type tests); a subdirectory inside a leftover is refused rather than entered (`O_NOFOLLOW` doesn't stop a mount-point walk or bind-mount cycle); the mark is removed through the same handle (closing a second re-open window); the handle is re-asked immediately before the irreversible step (a listing is not a snapshot); and an implausibly wide listing is refused rather than read into memory unbounded. The improved symlink test now marks the target so it satisfies every gate, plus a test that a link never yields a handle. The off-Unix path-based fallback and its exposure are now stated honestly in the ADR. The author's own third-round adversarial pass also fixed a data-loss bug their first fix introduced — exactly the right discipline on a destructive path.

**Release gate — unchanged, still correct to wait.** The migration signal shipped under a day ago; staged migrations haven't started; community/NTFS hosts aren't reporting; no rollback once published. Do not publish until the ADR-0015 gate is actually met. Merging the code is fine.

**Standing follow-ups (not blocking this merge):**
- **Roster half of the gate (MAJOR, gate-mechanics):** `Files` peers still aren't named individually — only as an aggregate count — so "roster member finished" can't be confirmed from the logs. Needs a low-rate named `peer_state` line for `files` peers (or a `peers_seen` list), or an out-of-band roster check, before the gate can be claimed met.
- **Version:** Cargo.toml is still `0.18.1` while ADR-0015 cites `v0.19.0-beta.1`; bump to 0.19.x and call the break out in release notes at release time.
- **Payments-seat flag:** in the client-PUT handler `holds_verified` (step 3) runs before payment verification (step 5) and its `Wrong` branch can hit the write path before any payment gate — worth a look from the payments side (not a data-integrity race).
- NIT: `pub struct StoreLayout` remains in a `pub(crate)` module under non-test cfg (now documented in the ADR's API-narrowing caveat).

No remaining code-level blockers. Approving the code; publish timing governed by the fleet gate.

jacderida:
## T3 testnet evidence — [V2-1213](https://linear.app/autonominetwork/issue/V2-1213), DEV-01 run 583 (2026-09-11/12)

**What ran.** This PR's head `cd5e91d`, built from `grumbach:storage/lmdb-removal-release-3` (binary sha256 `bba27f6136924215300cd183bb4d058a625a6877164add77ba33e16f2f6f6a63`, verified on 73/73 node+bootstrap VMs; every `ant-node starting` line carries `commit=cd5e91d`), on the staging-short profile: **990 nodes** (66 VMs × 15 across DO/Vultr/OVH/OVH-3AZ, 30 % NAT), 7 bootstraps, 10 uploaders + 2 downloaders, 6 hours, nodes at `info`. Adversarial cohorts planted with services stopped, then restarted twice (T+1:00 and T+3:30): marked env (A), tombstones + `migration-state.json` (B), unmarked store with data on a public and a NAT VM (C), plain file / symlink to a decoy / four near-miss names / marked store with a subdirectory / marked store bind-mounted read-only (D), sixty-six marked leftovers and symlink/socket marks (H), `ANT_SUSPEND_UNHELD_CHUNK_PENALTY=0` drop-in (E), 15 chunk files overwritten in place (K), 150 nodes restarted with measured stop cost (R), and 30 nodes on a previous-release config file with `db_size_gb` + `[storage.migration]` + `[upgrade] enabled = false` (G). Full procedure, numbers and per-node sha256 are in the two comments on the Linear issue.

**Result: all cleanup, corrupt-chunk, config, data-loss and hot-path gates passed.**

| Gate | Result |
|--|--|
| Cleaned leftovers go (A, B, H×66, D read-only after `umount`) | **PASS** — all 282 directories gone within 0.6 s of start (1 GiB in ≤0.6 s, 66 dirs in ≤0.1 s, in sequence), exactly one `space_returned` each, `df` back within 2 min, marker file untouched |
| Kept leftovers stay (C, plain file, symlink + decoy, near-misses, subdirectory, symlink/socket marks) | **PASS** — all 132 kept files byte-identical after both starts; exactly one `legacy_store_left` per leftover per start with the shape's reason text |
| Mark last (read-only store) | **PASS** — `RETIRED` present and dir non-empty after the first start; gone with one `space_returned` after `umount` + second start |
| Exact names only | **PASS** — near-misses never touched or named; fleet sweep of 66 VMs found nothing created/removed outside the cohorts; 0 cleanup lines on untouched nodes |
| Second start idempotent | **PASS** — cleaned cohorts logged nothing; kept cohorts logged the same single warning and deleted nothing |
| Stranded nodes serve | **PASS** — 59/59 C-held keys fetched at every check, C nodes seen by ≥946 observers throughout |
| Planted nodes read right | **PASS** — C/symlink/false-mark → `legacy`, plain file → `unknown`, all others `files`; the fleet-wide `peer_state` set was exactly the 42 planted nodes, 0 false positives, 0 misses |
| Corrupt chunks | **PASS** — 15/15 detected on first read (`does not match its name` → `Removed corrupt chunk file`), no restart or stall, all re-replicated byte-identical in 12 s–7 min |
| Old config loads | **PASS** — 30/30 started with `--config`, 0 parse errors |
| Penalty policy | **PASS** — suspension line on 1395/1395 starts, override VM logged `Ignoring it` 15/15, 0 `APPLIED` |
| Hot path / shutdown | **PASS** — `systemctl stop` p95 5.15 s, max 8.6 s over 375 restarts; 0 store-lock failures; upload p50 within +2 % of Phase 0 outside the restart hour |
| Stability / joins / transfers | **PASS** — 0 panics, 0 unplanned restarts; join p50 21.7 s; uploads 3729/3729, downloads 777/778 (one transient datamap miss in the restart hour) |
| No data loss | **PASS** — 535-chunk canary set 535/535 at all five checks |
| `UnknownCommitment` | 0 across 4 099 audit failures in Phases 2–4 |

**Two caveats.**

1. **"Clients never counted" (S3) failed by one edge.** On one of 21 402 signal ticks, one observer reported `peers_unreported=1` and a `peer_state … agent=none` line for a `client/0.27.3` connection. The `NotANode` filter and the agent string were correct (the same client was skipped correctly 780 times); the leak was a peer whose user agent had gone from the transport while its connection was still in the tally. **This has since been traced (V2-1260) and addressed by WithAutonomi/ant-node#228 (stop counting a peer that disconnects mid-tally as `unreported`) and WithAutonomi/saorsa-core#163 (record and drop the user agent under the connection entry).**
2. The audit-rate gate was not measurable as written on this build: commitment rotation is a 3600 s constant and the start-time build retires an empty store, so Phase 0 has no commitment-bound audits to baseline against. What could be measured — `UnknownCommitment` = 0 throughout, hourly rotation on every node — was clean.

Deviations from the planned manifest: uploader `eth` funding raised per the gas preflight (4.8 → 6.05 ETH) and `ant` pinned to 0.3.6; nothing else.

PR #220, opened 2026-09-01
feat(node): serve authenticated browser clients over WebRTC Direct

mickvandijke:
## Summary

Adds a browser endpoint alongside native QUIC. Browser v5 sessions use the shared ML-KEM/ML-DSA handshake and encrypted records; binary chunk requests reuse `ant_protocol::ChunkMessage` and the ordinary storage/payment handler. Nodes remain native.

Standard builds enable the listener by default, with configuration and build-feature opt-outs. Endpoints are published through the shared V1/V2 address plane. The devnet HTTP manifest supplies bootstrap metadata; file data travels directly between clients and nodes.

Request admission survives disconnect until work drains. Binary RPCs decode once, avoid redundant PUT serialization, reserve response headroom by request type, preserve structured settlement-version refusals, and redact backend details. Certificate initialization is locked and atomic. Failed startup drains migration and background work through the normal shutdown path. The browser endpoint uses the native filesystem/SQLite storage implementation.

## Linear issue

Closes V2-803 — [Linear issue](https://linear.app/autonominetwork/issue/V2-803/investigate-webrtcwebtransport-support-for-browser-access-to-the)

## Risk tier

Proposed for human review.

## Compatibility

- Wire: Adds browser v5 RPC and WebRTC Direct endpoint publication. Existing native QUIC, V1 publication, chunk wire messages, and payment proofs remain supported.
- Storage: Browser requests use the same native filesystem/SQLite store and migration path as QUIC requests; no browser-specific storage format.
- Api: Adds listener configuration, diagnostics, devnet endpoint artifacts, and the webrtc-direct feature. Browser sessions do not become routing-table members.

## Semver impact

- [ ] breaking
- [x] feature
- [ ] fix

## Test evidence

Validation of the updated EVM/protocol dependency pins during this refresh:

- `cargo check --all-targets --all-features`: passed.
- `cargo test --lib web_rtc --features test-utils`: 37 passed.

Earlier validation at `cf8a84cc` passed 1,127 library tests, four regular WebRTC integration tests, and all-target/all-feature Clippy. The coordinated Chromium seven-node Anvil upload/recovery/download check passed with that revision. This refresh changes dependency pins; it does not claim a new Chromium or full-node-suite run.

## New dependency

Shared transport WebRTC feature, `self_encryption` for devnet fixtures, and coordinated immutable Git patches for protocol/core/PQC/EVM/transport. Published compatible dependency releases remain required for registry packaging.

## ADR

- [ADR-0015-direct-browser-clients-over-webrtc-direct](https://github.com/WithAutonomi/ant-node/blob/web-support/docs/adr/ADR-0015-direct-browser-clients-over-webrtc-direct.md)

## Mitigation / rollback

Disable `webrtc_direct.enabled` or build without the WebRTC feature. Retain native clients and existing stored data.

## Coordinated dependencies

- [saorsa-transport#160](https://github.com/WithAutonomi/saorsa-transport/pull/160)
- [saorsa-core#158](https://github.com/WithAutonomi/saorsa-core/pull/158)
- [ant-protocol#29](https://github.com/WithAutonomi/ant-protocol/pull/29)
- [ant-client#186](https://github.com/WithAutonomi/ant-client/pull/186)

Exact immutable revisions are recorded in [Cargo.toml](https://github.com/WithAutonomi/ant-node/blob/26f5fb79fbd82ff7dbe016ea46cd1afffaf5c442/Cargo.toml) and Cargo.lock. Compatible published releases remain a release gate.

## Current stack revisions

All PRs link to V2-803. Immutable Git revisions used by the validated stack:

| PR | Revision |
| --- | --- |
| [saorsa-pqc#7](https://github.com/saorsa-labs/saorsa-pqc/pull/7) | [`29a2b272`](https://github.com/saorsa-labs/saorsa-pqc/commit/29a2b272c4ee8b72c17102da00033dbfe1ddcd37) |
| [evmlib#17](https://github.com/WithAutonomi/evmlib/pull/17) | [`cf424c04`](https://github.com/WithAutonomi/evmlib/commit/cf424c0447c9f1ca1313b43ce07df65a227e340c) |
| [saorsa-transport#160](https://github.com/WithAutonomi/saorsa-transport/pull/160) | [`6f0b1f62`](https://github.com/WithAutonomi/saorsa-transport/commit/6f0b1f623b487a222187e38d1e7fb25eb2f876cb) |
| [saorsa-core#158](https://github.com/WithAutonomi/saorsa-core/pull/158) | [`0df7853e`](https://github.com/WithAutonomi/saorsa-core/commit/0df7853ea24242694e91c67631986d6bea686199) |
| [ant-protocol#29](https://github.com/WithAutonomi/ant-protocol/pull/29) | [`cdae7d19`](https://github.com/WithAutonomi/ant-protocol/commit/cdae7d19f004bbd384e9a6cc91f4df253b416044) |
| [ant-node#220](https://github.com/WithAutonomi/ant-node/pull/220) | [`26f5fb79`](https://github.com/WithAutonomi/ant-node/commit/26f5fb79fbd82ff7dbe016ea46cd1afffaf5c442) |
| [ant-client#186](https://github.com/WithAutonomi/ant-client/pull/186) | [`f1303e1a`](https://github.com/WithAutonomi/ant-client/commit/f1303e1ad7bab49c8dc12af030ba71ec5fca0237) |

EVM recovery now retains ambiguous journals until finalized failure/replacement evidence is available. The protocol and native client include the matching consumer changes. Existing browser wire and payment formats are retained.

## Security dependency refresh (2026-09-15)

The shared transport now requires rustls >=0.23.45, fixing [RUSTSEC-2026-0285](https://rustsec.org/advisories/RUSTSEC-2026-0285.html). Affected lockfiles and downstream immutable Git pins are updated. This branch passes `cargo check --all-targets --all-features` and `cargo audit` (zero vulnerabilities; existing warning advisories remain).

Stable release dry runs now execute package verification while the upload step remains disabled. Workflow YAML and both conditions were checked locally. **Registry packaging is still blocked pending coordinated dependency releases**: `cargo package --no-verify` reproduces the missing `webrtc` feature in published transport 0.36.3. No crates were published or release versions selected. The [release dependency order and version-requirement steps](https://github.com/WithAutonomi/ant-node/blob/web-support/docs/WEBRTC_DIRECT_TESTNET.md#cratesio-release-dependency-order) document the remaining work.

## ICE admission interoperability fix (2026-09-16)

Admission challenges now carry a random nonzero 64-bit ICE-CONTROLLED tie-breaker, stable for the listener lifetime. This removes the zero value that libjuice treats as a missing role attribute. Downstream Git revisions are aligned.

Validation: `cargo check --all-targets --all-features` passes with the updated pins.

## Manual libjuice ICE validation (2026-09-16)

Built unmodified upstream libjuice 1.7.3 (`5a5ebd88969dc7abdcc2d4aeb160129ee40dc0ea`) and connected a C agent over localhost UDP to the real Rust WebRtcDirectListener at transport `7c77692d`, using v2 ICE credentials. ICE reached completed and the listener admitted it (0.84 seconds). With only the challenge tie-breaker changed back to zero, libjuice reported role-attribute mismatches and timed out while connecting. Restoring the fix passed again (0.63 seconds). Temporary control edits and the local harness were removed from the repository afterward.

This validates ICE/admission with libjuice itself, not DTLS/SCTP DataChannels, the colleague's Node.js wrapper, application authentication, payments/downloads, or the live network.

## Browser performance dependency refresh (2026-09-16)

Pins now include transport `6f0b1f62`, which forwards the admission-completing Binding request to ICE instead of waiting for another client retransmission. The coordinated client caches exact verified signed publications while retaining shared ownership, replacement and witness checks. No additional API, payment or wire changes are introduced by this refresh.

`cargo build --bin ant-devnet` passed; the rebuilt node passed the real Chromium paid-upload/recovery/download test and 64 MiB/600 MB transfer benchmarks. [Full stack measurements and validation](https://github.com/WithAutonomi/ant-browser-sdk/blob/17b7759f49ed9891bbfdec58e60a817271a923cd/docs/audits/2026-09-16-browser-performance-fixes.md).


grumbach:
Independent review of this PR at head `f7127540`, read as one stack together with saorsa-transport #160, saorsa-core #158 and ant-client #186. Everything below was verified against the source.

The server-side work is solid where it counts. The PQ handshake is genuinely unbypassable: `handle_webrtc_channel` establishes the session before any request is read (src/web_rtc.rs:947-951), and non-HELLO requests are refused until HELLO succeeds (src/web_rtc.rs:990-1024), with no plaintext fallback and no pre-handshake reachable handler. Browser stores and quotes go through the same `ant_protocol.try_handle_request` handler as native traffic (src/web_rtc.rs:1711), so there is no weaker parallel payment path, and `payment_quote_from_browser_quote` (src/web_rtc.rs:1763) rebuilds the native `PaymentQuote`, re-checks `quote.hash()` against the artifact and binds the content to the requested address, so the JSON projection cannot smuggle a modified quote past the signature. `validate_webrtc_config` is thorough about limit consistency and each invariant has a matching test. The two non-ignored devnet tests are real: a live listener, a real WebRTC dial, a real PQ session, and a check that the operator's RPC URL and credentials never reach a browser.

I also classified every `src/replication/` and `src/upgrade/` edit as behaviour-preserving, with one item I am less sure about: src/replication/mod.rs:3934 turns a closure returning an async block into an async closure, which looks equivalent but I could not type-check it against the pinned graph.

## The listener is enabled by default, and its failure stops the node

`Cargo.toml:186` has `default = ["logging", "webrtc-direct"]` and `src/config.rs` sets `enabled: cfg!(feature = "webrtc-direct")`, so every stock binary opens a second inbound UDP port, runs ICE/DTLS/SCTP and serves anonymous RPC. saorsa-transport's ADR-015 describes an opt-in listener, which is true of the cargo feature and not of the shipped node, and this repo's own ADR-0009 says production promotion is still gated on browser interop that has not happened.

Four things make that hard to back out of:

1. **A listener failure aborts startup.** In `RunningNode::run` (src/node.rs:549-582), any error from `web_rtc::spawn` calls `p2p_node.shutdown()` and returns `Err`. A UDP bind failure on the derived port, a truncated or unreadable certificate PEM, or on custom EVM networks a failed `eth_chainId` call now stops a healthy storage node from starting. Before this it would simply run.
2. **No CLI kill switch.** All three new flags set `enabled = true` (src/bin/ant-node/cli.rs:254-263). `docs/WEBRTC_DIRECT_TESTNET.md:163` documents `webrtc_direct.enabled = false` in a config file and that does work, but a fleet driven by flags would need a config file introduced to turn this off.
3. **`--no-default-features` does not isolate.** It omits `src/web_rtc.rs`, but `saorsa-webrtc` stays an unconditional dependency (Cargo.toml:38-43), `src/browser.rs:8` imports it unconditionally, and the `[patch.crates-io]` block (Cargo.toml:202-204) redirects `saorsa-core` and `saorsa-transport` to the unreleased PR revisions even in a disabled build. So rebuilding to roll back browser risk still ships the unreleased native core and transport lineage, which is not what `docs/WEBRTC_DIRECT_TESTNET.md:162-164` promises.
4. **The upgrader can leave a node offline.** `stop_on_upgrade` defaults to false (src/config.rs:514). In that mode src/upgrade/apply.rs:628-674 spawns the replacement before graceful shutdown and reports success with no readiness handshake, and the parent then cancels (src/node.rs:656). If the child reaches the same derived UDP port before the old listener releases it, the child hits the fail-closed path above and dies while the parent exits anyway. Rollback only covers binary replacement, not child startup.

Certificate persistence is also not atomic: `load_or_generate_certificate` (src/web_rtc.rs:668-690) writes the PEM to its final path and chmods 0600 in a later await, so a crash in between leaves the private key at the process umask, and a partial write leaves a corrupt PEM which, given (1), then prevents the node from booting. Existing files never have their permissions repaired. Writing a mode-0600 temp file and renaming would fix both halves.

Worth knowing on the dependency side: `Cargo.lock` grows from 653 to 718 packages in every default build, including `turn`, `webrtc-mdns`, `webrtc-media`, `webrtc-srtp`, `rtp`, `rtcp` and `sdp`. ADR-015 says the design does not introduce TURN or STUN servers, yet a TURN implementation and an mDNS responder are now linked in. It is worth confirming `webrtc-mdns` does not emit multicast traffic in this configuration, because an unexpected mDNS responder on production hosts would surprise operators.

## Two CI checks are red

The ADR governance job fails because `docs/adr/ADR-0009-direct-browser-clients-over-webrtc-direct.md` collides with `ADR-0009-audit-proof-shape-and-protocol-families.md` already on main, and `docs/adr/README.md` is not updated so the new ADR is absent from the index. Clippy fails with five denied lints, all in files this PR touched only mechanically: src/replication/config.rs:246, 255, 1663, src/replication/slice.rs:345 and src/replication/protocol.rs:1677.

Relatedly, the MSRV move from 1.75 to 1.91 exists to allow `Duration::from_mins` and `from_hours`, and 16 unrelated files across `replication/` and `upgrade/` were rewritten to use them. That enlarges the diff, couples a browser feature to a toolchain bump, and constrains every build environment. It would read better as its own PR.

## Worth fixing before this ships

**The derived port is not stable by default.** `NodeConfig::default()` sets the native port to 0 (src/config.rs:438), and the browser port is derived from whatever ephemeral port the OS assigns (src/node.rs:539, src/web_rtc.rs:82-120). A node started without an explicit `--port` publishes a different WebRTC address after every restart, invalidating cached endpoints even though it reuses its certificate. ADR-0009:483-488 and `docs/WEBRTC_DIRECT_TESTNET.md:132-137` both claim the complete address stays stable, which holds only for nodes given a fixed native port, which is what the testnet harness does.

**A valid `max_request_bytes` makes the browser pay and then be rejected.** The shared contract fixes the header bound at 64 KiB, but this node lets an operator configure `max_request_bytes` anywhere in `1..=65536` (src/web_rtc.rs:647) and enforces the smaller value at parse time (src/web_rtc.rs:959), before any request processing, and HELLO never advertises the node's real limit. A PUT header always exceeds 10 KiB, because the quote carries a 1,952-byte ML-DSA public key and a 3,309-byte signature hex-encoded, before the optional commitment sidecar. So with the limit set below about 11 KiB, HELLO and `quote_chunk` succeed, the browser pays on-chain, and `put_chunk` is rejected before storage. The 64 KiB default is safe with roughly 27 KiB of margin. Pinning the limit to the shared contract, or advertising it in HELLO, would close it.

**A partial config section inverts the default.** `WebRtcDirectConfig` has no container-level `#[serde(default)]`, and `enabled` carries a plain `#[serde(default)]` (src/config.rs:161-163) resolving to `false`, while every other field has an explicit default function. So a TOML file with no `[webrtc_direct]` section enables the listener, and one with the section but without `enabled` disables it:

```toml
[webrtc_direct]
max_connections = 48
```

`default_webrtc_listener_tracks_compile_time_feature` (src/config.rs:776) only exercises `WebRtcDirectConfig::default()`, never the deserialised path, so it cannot catch this.

**Storage errors reach anonymous browsers.** src/web_rtc.rs:1503 returns `"chunk read failed: {error}"` and :1598, :1684 return `{other:?}`. `src/browser.rs:76-84` goes to real trouble to avoid leaking RPC provider detail; this path does not, and with the file-per-chunk store those strings can carry filesystem paths.

**Per-IP limits do not aggregate IPv6.** `canonical_source_ip` (src/web_rtc.rs:474) only unmaps IPv4-mapped v6, so a routine /64 gives an attacker a practically unlimited number of distinct sources, each with its own connection slot, rate bucket and byte budget.

**`cache_insert_browser_devnet_seed` is behind the wrong feature.** src/payment/verifier.rs:1244 marks content as prepaid and is gated on `webrtc-direct`, which is on by default, so it ships in production binaries. It is `pub(crate)` and only reachable through `Devnet::publish_public_file`, which operates on its own verifier, so I do not believe it is exploitable, but a payment bypass belongs behind `test-utils` or a dedicated devnet feature.

**`resolve_automatic_config` falls back to loopback.** src/web_rtc.rs:99-108 uses `127.0.0.1` when the route probe fails, and the node then publishes that as its browser endpoint. saorsa-core drops loopback on ingest, so the publish is silently wasted rather than an error.

## Test fidelity and validation gaps

The five-node devnet test is valuable native interoperability coverage, but it is not browser coverage and does not test the browser client. `BrowserRpcClient` (tests/webrtc_direct_devnet.rs:426) is a native Rust client built on `saorsa_transport::webrtc_direct::WebRtcDirectClient`, with hand-written framing and PQ logic using the same `saorsa-webrtc` crate and constants as the server. If node framing and this adapter drift together while the deployed WASM client does not, CI stays green and every real browser fails.

Three more specific gaps:

- **Nothing moves a large payload.** The whole suite transfers 31 bytes (tests/webrtc_direct_devnet.rs:206) and 32 bytes (:349); `MAX_CHUNK_SIZE` appears only as a read bound. So the 16 KiB DataChannel chunking loop, multi-fragment reassembly, byte reservations at scale, the size-scaled write deadlines and SCTP backpressure are never exercised, even though `ant-devnet`'s own default public file is 5 MiB (src/bin/ant-devnet/main.rs:201) and the runbook implies automated verification publishes it.
- **The admission-timeout test is a mock.** `first_data_channel_timeout_releases_connection_admission` (src/web_rtc.rs:1874-1915) manually takes an admission guard, hands `pending()` to the timeout helper and lets the guard drop out of local scope. It never drives `handle_connection`, a real association or `accept_data_channel`, so it would still pass if production stopped wrapping the first channel in that helper. That is the invariant the listener's availability depends on.
- **The five-node test bypasses discovery.** It takes its initial endpoints straight from `Devnet::browser_endpoints()` (:203-220) and makes two manually scripted one-hop requests, with responses augmented by the in-process endpoint catalog, so it cannot catch a failure of V2 DHT publication, mixed-fleet propagation, alpha/K convergence, retained routing state, failure cooldowns or production seed bootstrapping.

ADR-0009 is admirably candid that there is no automated real-browser v5 flow and that Chrome, Firefox and Safari interop remain unmet acceptance criteria rather than claimed results. That is the right way to write it up. The tension is that an explicitly unmet acceptance criterion currently ships enabled by default.

For operations, what exists is startup logging, an endpoint file, static resource knobs, fail-fast startup and shutdown ordering. There is no listener health or readiness metric, no DHT reachability probe, no saturation alert, no runtime disable flag, no certificate backup or rotation procedure, no child-readiness handshake in the upgrader, no browser-aware rollback test and no firewall or NAT verification workflow.

## Smaller things

- ADR-0009 contradicts itself on the connection profile: :507-517 says the implemented profile is v2 with no v1 fallback, :742-746 says the implementation currently uses v1. The code uses v2, so :742 is stale.
- `docs/WEBRTC_DIRECT_TESTNET.md:8-10` says Rust 1.88 or newer while the package now requires 1.91, so following the runbook fails before you reach any test. The ADR also names `ant-protocol` 2.3.1 while Cargo uses 2.3.5, and says `saorsa-transport` owns persisted certificates while this repo loads and writes them.
- `--webrtc-direct-advertised-addr` has `requires = "webrtc_direct_bind"` (src/bin/ant-node/cli.rs:44), so overriding only the advertised address means redundantly repeating the bind.
- `saorsa-webrtc` is a git dependency with no version key in `[dependencies]`, which `cargo publish` refuses, and the crate name is already taken on crates.io at 0.1.0 through 0.1.2 by a different crate. That blocks releasing this repo.

## What I could not check

Nothing was built or run: the pinned cross-repo revisions are not in a local cargo cache here. No real browser, no NAT or firewall path, no mixed-version fleet, and no measurement of the resource envelope under load.

dirvine (commented):
Reviewed at `3ce1e9b147aa9074300c861aee7c1c7400710c24`.

**Advisory verdict: REQUEST_CHANGES before this leaves draft.** The core request path looks substantially stronger than the earlier review, but one current code issue and the declared release gates remain.

### Material finding

**Persist the DTLS private key atomically with private permissions** (`src/web_rtc.rs:821-837`).

`load_or_generate_certificate` writes the certificate and private key directly to the final path with `tokio::fs::write`. On Unix that normally creates the file as `0666 & umask` (commonly `0644`), and a crash or concurrent start can leave a truncated PEM. Because the listener is enabled by default and certificate-load failure aborts node startup, this is both key-material handling and boot-availability risk. Please create a same-directory temporary file with mode `0600`, flush/sync as appropriate, atomically rename it, and repair/reject unsafe permissions on an existing file. Add tests for mode and interrupted/corrupt persistence.

### Confirmed draft/release gates

- The current v4/v5 evidence is native Rust against Rust. The ADR explicitly says current Chrome, Firefox and Safari interoperability is still unproven, while `webrtc-direct` is enabled in the default build. This is already acknowledged in the PR's remaining work and should stay blocking until real-browser evidence exists and the ADR has human owners/reviewers/acceptance.
- The Git dependency stack is not publishable as declared. Locally, `cargo package --allow-dirty --locked --no-verify` fails because crates.io `saorsa-transport 0.36.3` lacks the requested `webrtc` feature. This matches the stated requirement to replace draft pins with reviewed releases before merge.
- The PR body gives `cargo test --test webrtc_direct_devnet --features webrtc-direct -- --ignored --nocapture` as the five-node command, but that executes **0 tests** because the five-node test is gated on `test-utils`. The CI/README form using `--features test-utils` is the effective command.

### Verification

- `cargo test --lib web_rtc --features webrtc-direct --no-fail-fast`: **29 passed**
- `cargo test --lib config::tests:: --features webrtc-direct`: **50 passed**
- `cargo test --test webrtc_direct_devnet --features test-utils -- --ignored --test-threads=1 --nocapture`: **1 passed**
- `cargo check --no-default-features --locked`: **passed**
- Current GitHub CI: build/test/clippy/security/devnet jobs green; `linear-link` and `pr-template` red, consistent with the declared draft work.

Review-team split: the focused request-path/security review found no remotely exploitable or availability-blocking flaw in `src/web_rtc.rs`; the architecture/release-readiness review requested changes for the browser-evidence, ADR and dependency gates. Four other review seats timed out and supplied no opinion, so I am not counting them as evidence.

dirvine (commented):
Re-review at new head `4c1509ddfaf1d626594c63d791a3a63dfee3e102`.

**Certificate blocker: resolved.** `src/web_rtc/certificate.rs` now writes the DTLS identity via a same-directory `0600` temp file, fsyncs temp + file + parent dir, atomically renames, repairs unsafe permissions on an existing file, rejects corrupt PEMs instead of silently regenerating, and serialises concurrent initializers with an exclusive `flock` on a retained `.lock` file. All three new tests pass locally (private-perms/repair, concurrent-init, interrupted-write). Verified: `cargo test --lib web_rtc --features webrtc-direct` → **37 passed**.

**Still blocking before this leaves draft:**

1. **rustls RUSTSEC-2026-0285 — Security Audit CI is failing.** The locked dep tree resolves `rustls 0.23.43` ("TLS 1.3 handshake messages incorrectly accepted across encryption level boundaries"; fix ≥0.23.45). This is the same advisory that currently blocks the saorsa-transport dependency stack. The `Security Audit` job fails on it. Since 0.23.45 is resolvable, this looks like a lockfile bump (`cargo update -p rustls`) plus whatever git-pinned crate still forces the older version. This is a genuine blocker, not a warning.

2. **Browser interoperability still unproven.** ADR-0015 line 570: "Current Chrome, Firefox, and Safari interoperability tests remain outstanding"; the ADR's own acceptance criteria require proof against real browsers before production. This is acknowledged in the PR's remaining work.

3. **Dependency stack still not publishable.** `cargo package --allow-dirty --locked --no-verify` fails: crates.io `saorsa-transport 0.36.3` does not have the `webrtc` feature this PR requires. Matches the stated requirement to replace draft git pins with reviewed releases before merge.

**Non-blocking notes.**
- ADR-0015 is still **Proposed** with owners/reviewers `<pending>` — fine for a draft, but the ADR needs human owners/acceptance before merge.
- The PR-body test command is now correct (`cargo test --lib web_rtc --features test-utils`: 37 passed) — the earlier `--features webrtc-direct` devnet command discrepancy is fixed.
- `linear-link` is now green; `pr-template` red is consistent with draft state.
- **Test (windows-latest)** fails on `fetch_responsibility_recheck::stale_fetch_candidate_is_declined_at_download_time` ("control candidate must enqueue", tests/e2e/fetch_responsibility_recheck.rs:156). That test is unrelated to the WebRTC path and the surrounding log shows Windows-specific noise (reachability backoff, `Access is denied` on commitment persistence), so this looks like a pre-existing/flaky Windows e2e test rather than a regression from this PR — worth a re-run to confirm. Ubuntu and macOS test jobs pass.

grumbach (commented):
## Review at `41acee71`

**Verdict: blocked as shipped, because the listener is on by default for an auto-upgrading fleet while nodes behind NAT or a firewall publish addresses browsers cannot reach. The code itself could merge if the listener defaulted off, or stayed on but failed open for browser-only startup errors.** Separately, there are two small code fixes.

### Must fix

1. **Clippy is red because of this PR's MSRV bump.** Raising `rust-version` from 1.75 to 1.91 turns on `clippy::duration_suboptimal_units` for a line that came from main (`src/storage/migration_signal.rs:51`). Fix: `Duration::from_mins(15)`.
2. **Shutdown can hang on a stuck RPC.** Node `cleanup` awaits the listener task (`src/node.rs:933-942`), which ends with `resources.drain_requests().await` and no deadline (`src/web_rtc.rs:967`). A browser PUT inside payment verification waits on `completedPayments(...)` with no timeout (`src/payment/verifier.rs:1822-1825`; evmlib's HTTP provider sets none). Native request handlers get a bounded 5 s drain (`PROTOCOL_DRAIN_GRACE`, `src/node.rs:54`); the browser drain has no bound, so this is new. The same missing timeout also holds browser request permits while an RPC endpoint hangs. Fix: bound `drain_requests` (for example 30 s) and put a deadline on the settlement lookup.

### The release-posture decision

The listener is enabled by default (`Cargo.toml:214`, `src/config.rs:223`) and any listener startup error stops the whole node (`src/node.rs:705-727`). ADR-0015 is still Proposed with no owners and still lists Chrome, Firefox and Safari as outstanding evidence. The open gates:

- **NAT and firewalls.** I did a focused pass on this; summary below.
- **Resource cost:** the node resource A/B has not run yet.
- **Rollback:** in staging run 590, nodes started with `--webrtc-port` died when they auto-upgraded to a stable build that does not know that flag. Any deployment that adds WebRTC flags needs them removed before rolling back.
- **Port stability:** with the default `0.0.0.0:0` bind, the WebRTC port is chosen by the OS on every start (nothing derives it from the QUIC port any more), so a node with a fixed QUIC port may get a different WebRTC port after each restart. The node republishes the new endpoint within a second, but static bootstrap lists and firewall rules cannot follow it. Deployment tooling needs to pin `--webrtc-port` per service and open it.

### NAT: can a browser reach a node behind NAT?

Only through a manual port forward that the operator configures. The browser gets no signaling: it dials the one literal address in the node's signed record, and the node is ICE-lite, so that address must accept unsolicited UDP. Nothing in the stack creates a NAT mapping for the WebRTC socket:
- UPnP maps only the QUIC port (saorsa-transport `nat_traversal_api.rs:2202-2203`), and there is no NAT-PMP or PCP support.
- The WebRTC socket never sends first.
- The MASQUE relays deliver into a secondary QUIC endpoint, and WebRTC publication skips relay addresses (`src/web_rtc.rs:866-873`).

With a wildcard bind, `advertised_addr` pairs the first same-family, non-relay native IP (behind NAT, the observed external IP) with the WebRTC socket's **local** port (`src/web_rtc.rs:855-874`), so behind NAT it publishes a port the NAT never mapped. It is always published as Unverified, and each unreachable node costs a browser a 10 s timeout and then a 30-minute cooldown.

| Node network | Works today? |
|---|---|
| Public IP, UDP port open | Yes |
| Firewall or security group opening only the QUIC port | No: pin `--webrtc-port` and open it |
| Manual forward, same external port | Yes, with a pinned port |
| Manual forward, different external port | Yes, with `--webrtc-direct-advertised-addr` (measured in Chromium, Firefox and WebKit) |
| UPnP, NAT-PMP or PCP router | No |
| Full cone, restricted, port-restricted, symmetric NAT | No |
| CGNAT | No |
| IPv6 behind a home-router firewall | No by default (IPv4-only bind by default) |

Browsers behind NAT reach public nodes fine; only UDP-blocked networks fail (no TURN).

Measured locally, with a UDP translator in front of a node and Chromium, Firefox and WebKit:
- A translated forward with the override set: works, HELLO in 51-194 ms.
- The same forward without the override: ICE, DTLS and the PQ handshake succeed, then the browser rejects the node ("node advertised a different WebRTC Direct endpoint").
- A port with no mapping: fails after 10.0 s.
- An advertised IPv6 endpoint: works.

Smallest options, cheapest first:
1. Configuration only: pinned port, firewall rule, override when forwarding.
2. Only publish a WebRTC endpoint when a port is configured, so home nodes stop costing browsers 10 s timeouts.
3. UPnP mapping for the WebRTC socket, reusing `UpnpMappingService`.
4. A relay-backed WebRTC path, or the single-port RFC 9443 design.

ADR-0015's context and decision drivers say NATed nodes "use direct ICE where possible or an end-to-end relay path" (lines 48-53 and 82-83), while its own "NAT and relays" section (lines 735-743) correctly says there is no WebRTC relay transport. Worth making the two agree.

### Other notes

- Test (ubuntu) red: `fresh_offer_capacity::normal_upload_never_reaches_fresh_offer_capacity`, a test from main. It passed 3 of 3 runs locally with zero refusals, so it looks load-sensitive; rerun.
- ADR-0015's number still collides with open #218's ADR-0015. Its devnet section still says "LMDB storage".
- The structured `QuoteChunk` browser RPC builds the unversioned quote request, so it skips the settlement-version refusal (`src/web_rtc.rs:1882-1885`). The shared upload engine uses the versioned binary path, so only the low-level WASM `quoteChunk` binding reaches it. Removing that legacy RPC, or making it versioned, closes it.
- Older small items still open: `max_connections` above 256 passes node validation but fails startup, the endpoint file is not written atomically, byte budgets can be set below one PUT, and the node's unsupported-version reply is unreachable because the shared parser rejects the version first.

Verified good: browser quotes and PUTs go through the same storage handler and payment verifier as QUIC (`try_handle_message` shares `handle_message_with_context`); I found no browser-only way to store unpaid data. The DTLS key is now written atomically, and on Unix it is forced to `0600`, including repair on load. The release workflow change only makes dry runs execute `cargo publish --dry-run`.

### Since my 2026-09-08 comment

Fixed:
- The `--disable-webrtc-direct` switch (`src/bin/ant-node/cli.rs:31-33`).
- Atomic certificate writes.
- mDNS is disabled on both sides.
- The ADR number, the ADR index and the Clippy lints from then.
- The partial-config default inversion.
- `max_request_bytes`: the header bound is now a fixed, shared 64 KiB.
- Storage errors are sanitised before they reach browsers (`src/web_rtc/errors.rs`).
- IPv6 per-IP limits are applied per /64.
- The prepaid devnet helper now needs `test-utils`.
- The loopback fallback.
- The ADR's v1/v2 contradiction and its version numbers.
- `--webrtc-direct-advertised-addr` no longer requires a bind.

On testing, the admission-timeout test now drives a real ICE/DTLS association, and a maximum-size PUT now crosses WebRTC.

Still open:
- Default-on with fail-closed startup (above).
- The port is still OS-chosen by default.
- `--no-default-features` still builds the unreleased core and transport revisions.
- The standalone upgrader still spawns the new process before the old one shuts down, with no readiness check (`src/upgrade/apply.rs:631-674`). That matters once `--webrtc-port` is pinned.
- The devnet tests still use a native WebRTC client rather than a browser. Real-browser coverage is now in ant-client's CI, Chromium only.
- One acceptance criterion in ADR-0015 (lines 1046-1048) still says protocol v4; the code is v5.

### Tested locally

- `cargo test --locked --lib web_rtc --features test-utils`: 37 passed.
- `cargo test --locked --test webrtc_direct_devnet --features test-utils`: 4 passed, plus the ignored five-node paid devnet: 1 passed.
- The NAT probes above, against this node, in three browser engines.

jacderida:
## Testnet evidence — DEV-03 run 591 (2026-09-17), `web-support` rebased on `main`

Four-hour staging-scale run of the `web-support` branches — **990 nodes + 7 bootstraps** (`nat_percent: 0`, one node VM clock-skewed +3 h), a **native client tier** (7 uploaders, 1 downloader) and a **WASM/browser-path client tier** (3 uploaders, 1 downloader; Node.js + `node-datachannel` driving `ant-core` compiled to WASM over WebRTC-Direct, bootstrapping from the 7 bootstrap multiaddrs and nothing else). Builds: ant-node `41acee71` (crate 0.19.0), ant-client `9b85491b` (crate 0.3.7; native `ant` and the WASM bundle from the same head); evmlib `ccd65f18`, saorsa-transport `3ed66a9c`. Full write-up with all 21 evidence items: [V2-1270](https://linear.app/autonominetwork/issue/V2-1270).

This is the first run of this series where the node fleet stayed the build it was deployed as (the previous run's 0.18.1-identified nodes auto-upgraded mid-run to a stable without the WebRTC listener): 997/997 services logged `No upgrade available` at every check, 0 `New version detected`, 0 service restarts, 0 replaced binaries, 7/7 bootstraps live at the end.

### The result that matters: sustained, flat throughput over the whole run, on both transports

Successful / attempted transfers per 30-minute slot from T0 (measurement window = slots 3–8):

| 30-min slot from T0 (UTC) | native uploads | WASM uploads | native downloads | WASM downloads |
|---|---|---|---|---|
| T0 → +30m (19:16) | 356/356 (29.8 GiB) | 12/13 (1.4 GiB) | 42/42 (9.8 GiB) | 13/13 (0.5 GiB) |
| +30m → +60m | 371/371 (30.9 GiB) | 16/16 (2.6 GiB) | 48/48 (8.9 GiB) | 15/15 (0.5 GiB) |
| +60m → +90m *window* | 385/385 (32.3 GiB) | 16/16 (2.0 GiB) | 42/42 (12.8 GiB) | 7/7 (0.3 GiB) |
| +90m → +120m *window* | 392/392 (33.4 GiB) | 16/16 (2.6 GiB) | 50/50 (10.3 GiB) | 11/11 (1.1 GiB) |
| +120m → +150m *window* | 391/391 (32.4 GiB) | 16/16 (2.6 GiB) | 48/48 (10.7 GiB) | 14/14 (0.5 GiB) |
| +150m → +180m *window* | 388/388 (32.3 GiB) | 18/18 (2.6 GiB) | 49/49 (11.2 GiB) | 11/11 (0.6 GiB) |
| +180m → +210m *window* | 388/389 (32.7 GiB) | 16/16 (1.7 GiB) | 53/53 (9.3 GiB) | 11/11 (0.6 GiB) |
| +210m → +240m *window* | 392/393 (33.8 GiB) | 17/17 (2.6 GiB) | 50/50 (9.9 GiB) | 15/15 (0.6 GiB) |

No decay in rate or success on either transport from the first half-hour to the last. Measurement-window totals: **native uploads 2336/2338 (99.91%)**, **native downloads 292/292**, **WASM uploads 99/99**, **WASM downloads 69/69**. Whole run (incl. warm-up): native 3085/3087 up / 386/386 down; WASM 127/128 up / 98/98 down. WASM covered 20 MB, 300 MB (single payment) and 900 MB (merkle) uploads, and downloads up to 500 MB, all paid through the JS callback (`approve` + `payForQuotes`) with the payment network taken from the node's HELLO.

Durations (ok ops, window): native upload p50 24.4 s / p95 126.5 s, WASM upload p50 146.1 s / p95 1021.8 s; native download p50 17.2 s / p95 72.2 s, WASM download p50 110.8 s / p95 236.3 s. Size-matched 20 MB: upload 16.8 s native vs 137.5 s WASM, download 15.5 s vs 96.6 s — the browser path is slower, as expected, but consistently so.

### Failures, all three of them (out of 3,699 transfers)

- WASM upload #1 at T0 on the 20 MB uploader: `Witnessed close group inconclusive before payment: got 0/1 quorum-recognised peers` — first-op-after-start transient, pre-payment, nothing spent; #2 succeeded three minutes later and nothing recurred in 127 further uploads.
- Native sgp/50 MB #336 (22:19Z): `witnessed close group initial lookup found 6 peers, need 7` — DHT transient, pre-payment.
- Native sjc/100 MB #418 (23:06Z): `payment error: Could not get current gas price: HTTP error 429` from the public Arbitrum Sepolia RPC, pre-payment, nothing spent, next upload fine.

That last one (plus one identical `-32000 context deadline exceeded` on the same call five minutes after the window) technically fails the run's zero-tolerance "no RPC-transport payment failure" criterion, so the issue is scored 10/11. **It is not a significant finding**: 2 in 3,087 native uploads (0.065%), never post-payment, never money lost, and the cause is narrow — `get_gas_price` is the one read in the payment prepare path that evmlib `ccd65f18` did not put behind `retry()` (`estimate_gas`, `get_transaction_count`, `get_chain_id`, `get_transaction_receipt`, `get_block_by_number` are). Zero `evmlib::retry` warnings appeared anywhere in the run, i.e. the public RPC was otherwise quiet. A one-line retry wrap closes it; it says nothing about the transport, the node, or the WASM path.

Also clean: 0 clock-skew signatures on either tier against the +3 h node (`unsubmitted payment quotes expired` / `paid proof expired`), 0 `awaiting chain finality`, 0 partial uploads, 0 `not within its local closest peers`, 0 `invalid public file size` (the WASM downloader skipped 111 over-cap entries by design), 0 OOM (WASM downloader peak 2.45 GB on the 500 MB pick), and a fleet-wide sweep of 1,009 `ant-*` units on 85 hosts found no panic, restart or replaced binary.

Node tier under this load: CPU mean 23.5% per host, per-service RSS p50 262 MB / p95 283 MB; steady-state 35.4 GB egress per GB uploaded, 8.7 GB disk per GB stored — in line with the previous (contaminated) run and not split by transport, since one fleet served both.

jacderida (approved):

PR #224, opened 2026-09-09
feat(storage): say on the wire whether a node still holds an old chunk store

grumbach:
Stacked on #223. The release that deletes the old chunk store may only be published once the
fleet has moved, and nothing today can establish that. This is the measurement that gate needs,
and it has to ship and soak **before** that release, not with it.

**The state rides the user agent.** `saorsa-core` already sends one with every signed message
and keeps each peer's, so this is a different value in a string that was already on the wire:
no new message, no new field, no protocol version. The `node/` prefix stays, because
`is_dht_participant` gates DHT membership on it and is its only consumer.

**Three states, never folded into two.** A directory that could not be read is not one that is
not there, and reading "cannot tell" as "finished" is exactly how a gate comes back clean over
a fleet that is not.

**Only a node's own line can close the gate.** The agent is fixed when the transport is built,
so a node that finishes migrating keeps telling peers it has an old store until it restarts.
Its own fifteen-minute line re-reads the disk and can reach zero; peer observation is how nodes
we have no logs from are found at all, and can never reach zero. Per-peer lines are emitted at
`info`, because nodes run at `info` and a line nobody emits cannot gate a release, and only
for peers that have **not** finished, so the volume is bounded and falls away as the answer
arrives.

**The reporter holds a weak handle.** A reporter must not be the reason the thing it observes
stays alive: a strong one would keep a dropped node's transport, and its bound port, for as
long as the task ran.

Also persists the staged-rollout window under the node root. It was in memory only, so a node
restarting inside its own delay restarted its own timer and could defer an upgrade
indefinitely. Every failure to read or write it resolves to "upgrade now", never "keep
waiting". And the terraform worker unit passes `--enable-logging`, without which a node on
that path emits nothing at all and none of this is readable.

## Linear issue

Closes V2-1200

## Risk tier

## Compatibility

- Wire: the user agent gains a `migration/<state>` token after the existing `node/<version>`
  prefix. The prefix is preserved. Peers on older builds do not read the token; this build reads
  a peer without one as `unreported`, counted separately and never as finished.
- Storage: one new file, `{root}/upgrade-rollout.json`. An older build ignores it.
- API: adds `storage::migration_signal` and `upgrade::RolloutState`.

## Semver impact

- [ ] breaking
- [x] feature
- [ ] fix

## Test evidence

Local, on this branch: `cargo test --lib` **1110 passed, 0 failed**.
`cargo clippy --all-targets --all-features -- -D warnings` clean (the exact CI command), plus
`cargo fmt` and `RUSTFLAGS="-D warnings" cargo build --release --no-default-features`.

Fourteen tests in the two new modules, including: an unreadable directory is never read as
finished; a client is not a node that failed to report; silence is counted as silence; the user
agent keeps the prefix that gates DHT membership; the rollout window survives a restart; an
unreadable or corrupt stamp means upgrade rather than a fresh window; and the reporter lets go
of a node that was dropped.

**Not covered here:** a dev testnet run showing the per-peer lines aggregating into a usable
fleet count.

## New dependency

none

## ADR

https://github.com/grumbach/ant-node/blob/aa4de41f296a2fe79488817e6db5b5c8b15d7017/docs/adr/ADR-0014-file-based-chunk-store-and-lmdb-retirement.md

## Mitigation / rollback

Revert the commit. The user agent loses its token and peers read the node as `unreported`,
which is where every pre-signal build already sits. The rollout stamp becomes an ignored file.

### DIFF
--- a/config/production.toml
+++ b/config/production.toml
@@ -49,57 +49,18 @@ verify_on_read = true
 # Maximum size in GiB of the legacy LMDB store, while one still exists
 # (0 = derive it from available disk). Retired along with LMDB itself.
 db_size_gb = 0
-
-# --- Moving off the legacy LMDB chunk store ---
+# Storage migration off the old chunk store
 #
 # Chunks are now one file each, under {root_dir}/chunks/. A node that still has a
 # chunks.mdb copies it into files in the background, then deletes it whole, which is the
 # only moment LMDB's disk comes back.
 #
-# The two release-level switches (whether to delete the old store, and whether audits
-# still penalise) belong to the build, not to this file, so they are deliberately absent.
-[storage.migration]
-# Run the copier. Turning this off leaves both stores in place forever and never
-# returns the old store's disk.
-enabled = true
+# There is deliberately no [storage.migration] section. Every one of its settings belongs
+# to the build rather than to this file: the release that later deletes the old store has
+# to be able to assume every node ran the same schedule, and it cannot assume that while
+# the schedule belongs to whoever last edited a config. A [storage.migration] section left
+# over from an earlier release still parses, it is simply ignored.
 
-# Also write new chunks to the legacy store while it exists, so a fleet rollback to an
-# older build cannot lose a chunk uploaded during the migration.
-dual_write_legacy = true
-
-# Allow a node that cannot fit its chunks to give up the ones it is furthest from.
-#
-# Whatever this is set to, a chunk is only ever given up when the node is near the back of
-# its group for it, its close group has received the node's reduced commitment, AND all but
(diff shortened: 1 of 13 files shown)


grumbach:
Folded into #223 rather than stacked. The two changes are one release, and splitting them made the chain longer without making either smaller.

The peer-observation half is also gone. It could show that *some* node was unfinished, but never that none was: a node only sees the peers it is connected to, so completing that picture needs an authoritative roster of every node that ought to exist, and there is no such inventory. What each node says about itself can reach zero, so that is what the gate reads. That removed about 260 lines, and the persisted staged-rollout window came out too, as an upgrade-subsystem change rather than a migration one.

PR #225, opened 2026-09-14
feat: stamp node_version and node_commit on every json log line

jacderida:
## Linear issue
Closes V2-1153

## Risk tier

Proposed T1: the only behavioural change is two extra fields on each JSON log line. No wire, storage, payment, routing or upgrade-mechanism change.

## Compatibility
- Wire: none
- Storage: none
- API: none. JSON log lines gain `node_version` and `node_commit` fields; the text format and the startup line's existing `version` / `commit` fields (parsed by the ant-client beta forwarder) are unchanged.

## Semver impact
- [ ] breaking
- [ ] feature
- [x] fix

## Summary

Every document in production Elasticsearch carries `tag.node_version = "unknown"`, so the build a node is running cannot be read from telemetry. That made the 0.18.1 staged rollout unreadable: fleet-wide `relay recv stream closed` errors rose 6x across the window, which looked like a regression in the build that fixed them; only splitting by the build each service was actually running (inferred from process start time) showed the new build was ~4x better (V2-986).

The tag is a static Telegraf `[global_tags]` value set at provisioning time in `saorsa-testnet-registry` and never passed by any caller. Even if it were passed, a provisioning-time string can never be right across an auto-upgrade, so the process itself has to report its build.

- `ant_node::logging::WithBuildInfo` wraps the JSON event formatter: it renders the inner event into a thread-local buffer and replaces the closing brace with a compile-time `,"node_version":"<CARGO_PKG_VERSION>","node_commit":"<ANT_GIT_COMMIT>"}` tail. One memcpy per line on top of the serialisation already done; re-entrancy falls back to a fresh `String` rather than panicking; non-object output passes through untouched.
- Why a wrapper: tracing-subscriber's JSON formatter has no hook for constant fields, and span fields do not survive `tokio::spawn`, so a root span would miss most lines.
- Both JSON sinks (stdout and daily-rolling file) use it via `json_event_format()` in `main.rs`.
- Auto-upgrade: the upgrade path either `exec`s the new binary or exits for the service manager to restart it, so the new process stamps its own constants — correct on both sides of the restart by construction.

Companion change: jacderida/saorsa-testnet-registry adds `node_version`, `node_commit` to Telegraf `tag_keys` so the fields land as `tag.node_version.keyword` / `tag.node_commit.keyword` and beat the static global tag. Until that config is re-provisioned on ant-prod-01, the values still land as top-level `node_version` / `node_commit` fields, so they are queryable either way. The `bootstrap-upgrade-check` / `track-upgrades-setup` skill notes saying `tag.node_version` is static remain true until both have rolled out.

## Test evidence
- 3 new unit tests in `src/logging.rs`: every line is valid JSON and keeps `message`/`level`/`target`/custom fields; the stamp lands after a nested `"span":{...}` object; `node_version` equals `CARGO_PKG_VERSION` and parses as semver.
- `cargo test`: 1088 lib + 16 integration passed, 0 failed.
- Smoke run in development mode: stdout sink 295/295 lines and file sink 81/81 lines carried `node_version=0.18.1 node_commit=31fcbae`, 0 invalid JSON (`jq`-checked); text format byte-identical.
- `cargo clippy --all-features -- -D clippy::panic -D clippy::unwrap_used -D clippy::expect_used` clean; CI's `cargo clippy --all-targets --all-features -- -D warnings` clean; `cargo build --no-default-features` compiles.

Sample line:
```json
{"timestamp":"2026-09-11T23:58:07.002929Z","level":"INFO","message":"Bootstrap peers provided via CLI","count":1,"target":"ant_node","node_version":"0.18.1","node_commit":"31fcbae"}
```

## New dependency
none

## ADR
n/a

## Mitigation / rollback
Revert this commit; the two fields disappear and Telegraf's static global tag takes over again. No stored data or peers are affected.

https://claude.ai/code/session_01MYvuVpFYXXNrY9YFVBswvp


dirvine (changes requested):
**REQUEST_CHANGES — Documentation CI job is red due to a doc lint error introduced by this PR.**

`cargo doc` fails under `-D warnings`:

```
error: public documentation for `WithBuildInfo` links to private item `BUILD_INFO_TAIL`
  --> src/logging.rs:142:7
142 | /// [`BUILD_INFO_TAIL`], and the result is copied to the real writer. Cost is
     |       ^^^^^^^^^^^^^^^^ this item is private
note: `-D rustdoc::private-intra-doc-links` implied by `-D warnings`
```

The new public `WithBuildInfo` struct's doc comment intra-doc-links the private `BUILD_INFO_TAIL` const (src/logging.rs:142). `stamp_build_info` (line 193) is fine because it's private. Your local clippy runs don't catch this — the `cargo doc` build is a separate CI job.

**Fix (one line):** change the link to plain backticks at line 142:
```rust
/// `BUILD_INFO_TAIL`, and the result is copied to the real writer. Cost is
```

The logic itself (WithBuildInfo wrapper, re-entrancy fallback, empty-object/nested-span splice handling, both sinks via json_event_format, text format unchanged) all looks correct and the tests cover the edge cases well. Just need the doc link fixed so CI is green — then this is good to go.

jacderida:
Fixed in 1c892f9 — the `WithBuildInfo` doc comment now uses plain backticks for `BUILD_INFO_TAIL` instead of an intra-doc link. Verified locally with `RUSTDOCFLAGS="-D warnings" cargo doc --all-features --no-deps`, which now passes.

dirvine (approved):
**APPROVE** — re-review complete.

The doc-lint fix is confirmed in place: line 142 now uses plain backticks ( `BUILD_INFO_TAIL` ) instead of the intra-doc link to the private const, and the **Documentation** CI job is green.

Remaining state:
- Format, Clippy, Documentation, Test (no logging), linear-link, pr-template, self-test — all SUCCESS.
- **Security Audit is red**, but it's a Cargo.lock dependency advisory (RUSTSEC-2026-0285), not something this PR introduced — the PR only touches `src/bin/ant-node/main.rs` and `src/logging.rs` with no dependency changes. Pre-existing and independent of this change; not a blocker here.

The logging change itself is sound (verified in the prior pass): `WithBuildInfo` stamps node_version/node_commit on every JSON line via both sinks, with correct handling of re-entrancy, empty objects, and nested span objects, and text format untouched. Good to go.

PR #228, opened 2026-09-15
fix(storage): stop counting a peer that disconnects mid-tally as unreported

grumbach:
## Linear issue

Closes V2-1260

## Risk tier

Changes the migration signal's output: the per-node tally no longer counts a peer that disconnected mid-tally, and no `peer_state` line is written for it. No wire, format or penalty change.

## Compatibility

- Wire: none.
- Storage: none.
- API: none.

## Semver impact

- [ ] breaking
- [ ] feature
- [x] fix

## Test evidence

- `cargo test --lib --features test-utils migration_signal`: 15 passed, 0 failed.
- `cargo clippy --all-targets --all-features -- -D warnings` (the CI invocation), `RUSTFLAGS=-D warnings cargo check --no-default-features --lib`, `cargo fmt --all -- --check`, `RUSTDOCFLAGS=--deny=warnings cargo doc --no-deps`: clean.
- What it fixes, from the 990-node testnet for #218: one tick in 21,402 had `peers_unreported=1`, a `client/0.27.3` connection tallied with `agent=none`. The tally takes the peer list, then reads each agent, and saorsa-core drops the agent when the peer's last channel closes. No unit test: the race needs a peer to disconnect inside the loop, which the store harness cannot stage.
- Companion: WithAutonomi/saorsa-core#163 records and drops the agent under the peer's connection entry, so a connected peer always has one. This PR is correct without it: a peer still connected with no agent stays counted as unreported. The `saorsa-core` pin in `Cargo.lock` moves when the rc picks that PR up.

## New dependency

none

## ADR

https://github.com/WithAutonomi/ant-node/blob/rc-2026.9.2/docs/adr/ADR-0014-file-based-chunk-store-and-lmdb-retirement.md

## Mitigation / rollback

Revert the commit. It only changes the migration signal's log output.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -809,7 +809,7 @@ dependencies = [
 
 [[package]]
 name = "ant-node"
-version = "0.19.0-rc.1"
+version = "0.19.0-rc.2"
 dependencies = [
  "alloy",
  "ant-protocol",
@@ -862,8 +862,8 @@ dependencies = [
 
 [[package]]
 name = "ant-protocol"
-version = "2.4.0-rc.1"
-source = "git+https://github.com/WithAutonomi/ant-protocol?branch=rc-2026.9.2#b52d5fa4cec1932c5c9bd34cd86aca823eaf5dde"
+version = "2.4.0-rc.2"
+source = "git+https://github.com/WithAutonomi/ant-protocol?branch=rc-2026.9.2#0713e91174145c8abf15692210622e4d1194bfd1"
 dependencies = [
  "blake3",
  "bytes",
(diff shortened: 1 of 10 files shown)


dirvine (approved):
APPROVE — reviewed at exact head `674aa0d2eb9dcf794497a69ac6affc15accf47ff` together with saorsa-core #163.

The tally fix is conservative in every relevant interleaving:
- `agent=None` followed by disconnected means the peer left after the `connected_peers` snapshot, so skipping it removes the observed false `Unreported` tick.
- If the peer reconnects between the two reads, it remains `Unreported`; that is a possible false outstanding count, not false clearance.
- If an agent is read before a disconnect, the snapshot's last announced state is counted, matching the existing snapshot semantics.
- A still-connected peer with no agent remains `Unreported`.

The change therefore cannot manufacture a `Files` result or falsely clear the migration gate. It is safe on its own and gains the stronger missing-agent meaning from saorsa-core #163.

Verification:
- focused migration-signal suite: 15 passed
- clippy all targets/features: passed
- no-default-features library check: passed
- formatting and rustdoc warnings check: passed
- GitHub build/test/lint/filesystem/platform matrix is green

Non-blocking gap: no deterministic mid-tally disconnect/reconnect test was added. The branch logic is small and the paired transport invariant was checked directly, so I do not consider this a merge blocker.

The failing Security Audit is inherited, not introduced: `Cargo.lock` is byte-identical to the base and the newly published RUSTSEC-2026-0285 affects that existing rustls version. It should be handled by the release train in a separate dependency bump to rustls >=0.23.45.

Operationally, this is consistent with — and closes — the reported false-positive path in the #218 testnet. I have not independently reclassified the whole run from its Linear evidence here. The peer tally remains observational evidence rather than proof of whole-fleet completion, so the rest of the ADR-0015 gate still applies.

